PR #329, opened 2026-04-10
feat: include tool_name in MCP audit input when tool_input is empty

### LINKED ISSUE #299: audit: record MCP tool call parameters in hook
## Background

MCP tool calls are currently recorded as `command_executed` events via the PostToolUse hook, but only the tool name is stored in the `body` field. Parameters (tool_input) and return values are not captured.

Example of current recording:
```
kind: command_executed
body: "mcp__plugin_traceary_traceary__search"
```

In contrast, Bash command audits include the full command string.

## Proposal

Extract `tool_input` from the hook payload and pass it to `traceary audit --input` so that MCP tool call parameters are recorded alongside the tool name.

## Considerations

- Return values may be too large to store; parameters alone may be sufficient
- Sensitive data filtering (API keys, tokens) should be considered
- The hook script (`traceary-audit.sh`) already has MCP tool_name fallback logic that can be extended

duck8823:
## Summary
- When MCP tool calls have empty \`tool_input\`, build fallback input with \`tool_name\`
- Updated hook script in all 3 integration packages

Closes #299

## Test plan
- [x] \`go test ./...\` passes (including hook script tests)
- [x] \`python3 scripts/verify_integrations.py\` passes

### DIFF
--- a/integrations/claude-plugin/scripts/traceary-audit.sh
+++ b/integrations/claude-plugin/scripts/traceary-audit.sh
@@ -36,6 +36,13 @@ fi
 
 REPO_VALUE="$(traceary_resolve_effective_repo "$CLIENT")"
 AUDIT_INPUT="$(traceary_json_get 'tool_input' '{}')"
+# For MCP tools, ensure tool_input includes at least the tool name when empty
+if [[ "$AUDIT_INPUT" == "{}" || -z "$AUDIT_INPUT" ]]; then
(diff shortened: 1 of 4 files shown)


duck8823:
## 🤖 AI コードレビュー結果

### レビュアー
- Codex verifier: ✅
- Claude final: ✅

### 結論
**APPROVE** — 問題なし

PR #330, opened 2026-04-10
fix: warn when ending an already-ended session

### LINKED ISSUE #308: consider: double session end silently succeeds and creates duplicate events
## Observation

Ending an already-ended session creates a second \`session_ended\` event without error:

\`\`\`sh
sid=\$(traceary session start --client test --agent test)
traceary session end --session-id "\$sid"   # Recorded: abc123
traceary session end --session-id "\$sid"   # Recorded: def456 (no error!)
\`\`\`

This creates two \`session_ended\` events for the same session, which may cause incorrect duration calculations or confuse session-list queries.

## Question

Is this intentional? If not:
- Option A: Return an error if the session is already ended
- Option B: Return the existing end event ID (idempotent)
- Option C: Add a warning to stderr but still succeed

duck8823:
## Summary
- Check \`ended_at\` before UPDATE in \`SaveSession()\`
- Log \`slog.Warn\` if session already ended
- Do not fail (hooks need stability)

Closes #308

## Test plan
- [x] \`go test ./...\` passes
- [x] \`go tool golangci-lint run\` passes

### DIFF
--- a/infrastructure/sqlite/session_store.go
+++ b/infrastructure/sqlite/session_store.go
@@ -2,12 +2,13 @@ package sqlite
 
 import (
 	"context"
+	"database/sql"
 	"log/slog"
 
 	"golang.org/x/xerrors"
 
-	"github.com/duck8823/traceary/domain/port"
 	"github.com/duck8823/traceary/domain/model"
+	"github.com/duck8823/traceary/domain/port"
 )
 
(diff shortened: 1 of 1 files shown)


duck8823:
## 🤖 AI コードレビュー結果

### レビュアー
- Codex verifier: ✅
- Claude final: ✅

### 結論
**APPROVE** — 問題なし

PR #487, opened 2026-04-13
fix: dedupe extracted memories after redaction changes

### LINKED ISSUE #474: v0.5.1-3: dedupe extracted memories across redaction-rule changes
Part of #480.

## Summary

Normalize extraction dedupe against the currently sanitized durable-memory fact, so extraction does not create duplicate candidates after redaction rules change.

## Motivation

`memory extract` dedupes newly proposed candidates against:

- the newly sanitized candidate fact for the current run
- the stored `summary.Fact()` value returned from the existing memory list

That means a historical candidate accepted or proposed under older redaction rules can still collide semantically after a later redaction-pattern change. Example:

- old stored fact: `token is abc123`
- new sanitization rules now produce: `token is [REDACTED]`
- extraction run sanitizes the new candidate before keying it
- the historical stored fact is keyed as raw text, so the duplicate is not detected

This is not destructive, but it can create noisy duplicate candidates.

## Acceptance criteria

- extraction dedupe compares against the currently sanitized form of existing stored facts
- regression coverage proves duplicate candidates are not created when redaction patterns change between runs
- behavior remains candidate-only; no migration or automatic acceptance is required

duck8823:
## Summary

- sanitize existing durable-memory facts with the current redaction pipeline before extraction dedupe
- keep candidate extraction behavior unchanged while removing noisy duplicates after redaction-rule changes
- add regression coverage for historical raw facts that now collapse to the same sanitized candidate

Closes #474

### DIFF
--- a/application/usecase/memory_extraction_usecase_impl.go
+++ b/application/usecase/memory_extraction_usecase_impl.go
@@ -77,13 +77,13 @@ func (u *memoryExtractionUsecase) Extract(ctx context.Context, criteria apptypes
 	}
 
 	scope := extractedMemoryScope(session)
-	existingKeys, err := u.loadExistingCandidateKeys(ctx, scope)
-	if err != nil {
-		return nil, err
-	}
 	extraRedactors, err := compileExtraRedactPatterns(u.extraRedactPatterns)
 	if err != nil {
 		return nil, xerrors.Errorf("failed to compile extraction redaction patterns: %w", err)
+	}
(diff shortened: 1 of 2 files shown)


duck8823:
Claude CLI: ✅ NO_BLOCKING_ISSUES

Validation:
- `go test ./...`
- `go build ./...`
- `go tool golangci-lint run`

Gemini scout: retrying (transient model-capacity issue on the latest run).

@codex review

duck8823:
Gemini scout: ✅ NO_BLOCKING_ISSUES

PR #492, opened 2026-04-13
fix: scan tail windows under a stable read snapshot

### LINKED ISSUE #489: v0.5.1-7: fix tail pagination data loss under concurrent writes
## Problem

`presentation/cli/tail.go:302-344` paginates the follow-mode poll window (`created_at >= cursor.timestamp` / `< snapshotTo`) with OFFSET. Because the event timestamps are assigned before persistence, a concurrent event committed between page reads can shift later pages, so `offset += len(page)` silently skips rows. `cursor.Advance()` then advances the watermark to the newest fetched timestamp, and the skipped events are never revisited.

Under concurrent logging, `traceary tail` can silently drop events.

Reported by Codex verifier (gpt-5.4 xhigh) during v0.5.1 pre-release review.

## Additional concern

The paging loop itself is a read-side orchestration responsibility that currently lives in the presentation layer. The fix is a natural place to relocate it.

## Acceptance

- Tail follow mode does not drop events when another client writes to the same range between page reads. Add a regression test that simulates concurrent inserts mid-scan.
- Paging orchestration (batch size, cursor windowing) moves into `application/queryservice` (or a dedicated usecase). `presentation/cli/tail.go` calls a single method.
- Clarify and document whether `EventListCriteria.From` is inclusive or exclusive.
- `go test ./...` / `golangci-lint` green.

Parent: #480

duck8823:
## Summary

- Fix Codex-reported data-loss bug: `traceary tail` follow mode was paginating its poll window with OFFSET across independent queries, allowing a concurrent writer to shift later pages and cause silently dropped events.
- New `EventQueryService.ListWindow` runs the paged scan inside a single SQLite read transaction so snapshot isolation keeps the page sequence stable.
- Exposes the method on `EventUsecase` and collapses `presentation/cli/tail.go`'s former `listTailEventsSince` loop into one `ListWindow` call, also moving the paging orchestration out of the CLI layer.

## Test plan

- [x] \`go test ./...\`
- [x] \`go tool golangci-lint run\`
- [x] New \`TestDatasource_ListWindow_ReturnsAllEventsAcrossBatches\` verifies a 250-event scan returns every row in order with no duplicates across three internal batches.
- [x] New \`TestDatasource_ListWindow_RespectsToUpperBoundExclusive\` guards the \`To\` boundary contract.
- [x] Updated \`TestPollTailEvents_*\` confirm the CLI layer issues one \`ListWindow\` call per poll.

Closes #489

### DIFF
--- a/application/queryservice/event_query.go
+++ b/application/queryservice/event_query.go
@@ -13,6 +13,11 @@ import (
 type EventQueryService interface {
 	// ListRecent returns events in descending time order.
 	ListRecent(ctx context.Context, limit, offset int, kind types.EventKind, client types.Client, agent types.Agent, sessionID types.SessionID, workspace types.Workspace, failuresOnly bool, from, to time.Time) ([]*model.Event, error)
+	// ListWindow returns every event matching the criteria whose created_at
+	// falls in [From, To) under a single read snapshot so concurrent writers
+	// cannot cause the scan to drop events. Callers supply the batch size via
+	// criteria.Limit(); offset is ignored.
+	ListWindow(ctx context.Context, criteria apptypes.EventListCriteria) ([]*model.Event, error)
 	// Search performs full-text search across events.
 	Search(ctx context.Context, query string, workspace types.Workspace, sessionID types.SessionID, client types.Client, agent types.Agent, kind types.EventKind, from, to time.Time, limit, offset int, failuresOnly bool) ([]*model.Event, error)
 	// GetContext returns recent events for context retrieval.
--- a/application/usecase/event_usecase.go
+++ b/application/usecase/event_usecase.go
@@ -22,6 +22,12 @@ type EventUsecase interface {
 	// List returns events in descending time order.
 	List(ctx context.Context, criteria apptypes.EventListCriteria) ([]*model.Event, error)
 
+	// ListWindow returns every event matching the criteria whose created_at
+	// falls in [From, To) under a single read snapshot, so concurrent writers
+	// cannot cause the paged scan to drop events. criteria.Limit() controls
+	// the per-page size for the internal scan; criteria.Offset() is ignored.
+	ListWindow(ctx context.Context, criteria apptypes.EventListCriteria) ([]*model.Event, error)
+
 	// Show returns the details for a single event.
 	Show(ctx context.Context, eventID types.EventID) (apptypes.EventDetails, error)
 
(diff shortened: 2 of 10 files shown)


duck8823:
## Multi-AI Review Summary

| Reviewer | Result |
|---|---|
| **Codex verifier (gpt-5.4 xhigh)** | ✅ no issues — confirms the OFFSET data-loss bug is closed; verified SQLite rollback-journal snapshot semantics via official docs (lang_transaction, lockingv3) |
| **Claude architect** | HIGH×2: offset footgun, ListWindow placement on EventUsecase |
| **Claude reviewer** | HIGH×1: test didn't assert multi-batch execution |

### Resolved

- **Offset footgun** — `EventUsecase.ListWindow` and `EventDatasource.ListWindow` now return an explicit error when `Offset() != 0`. New `TestDatasource_ListWindow_RejectsNonZeroOffset` covers it.
- **Multi-batch assertion** — added `SetListWindowBatchHookForTest`; `TestDatasource_ListWindow_ReturnsAllEventsAcrossBatches` now asserts the exact sequence `[100, 100, 50]`, so a regression that returned everything in a single query would be caught.

### Pushed back (with reasoning)

Claude architect flagged placing `ListWindow` on `EventUsecase` as leaking persistence-layer concerns. I'm keeping it there for the following reason: `ListWindow` is a query **contract** ("return every event in [From, To) consistently"), not an implementation detail. `EventUsecase` already aggregates analogous query operations (`List`, `Search`, `Context`, `Timeline`) as the single presentation-facing port, and `presentation/cli/tail.go` only holds `usecase.EventUsecase`. Moving a single operation to a separate `TailQueryService` injected directly into `RootCLI` would break that symmetry for no correctness gain. The snapshot-isolation guarantee lives in the interface doc comment as a contract; infra is free to satisfy it however it likes.

duck8823:
🤖 AI コードレビュー結果

- Codex verifier (gpt-5.4 xhigh): no issues. SQLite rollback-journal snapshot isolation confirms the OFFSET data-loss bug is closed.
- Claude architect: HIGH×2 → Offset footgun (guard added), ListWindow placement (pushed back with reasoning).
- Claude reviewer: HIGH×1 → multi-batch assertion (fixed via SetListWindowBatchHookForTest).

All blockers resolved. Mergeable.

PR #493, opened 2026-04-13
test: cover custom redact pattern in memory dedupe test

### LINKED ISSUE #490: v0.5.1-8: strengthen dedupe test to cover redaction-pattern normalization
## Problem

`application/usecase/memory_extraction_usecase_impl_test.go:427` — `TestMemoryExtractionUsecase_Extract_DeduplicatesExistingFactsAfterSanitization` runs with `extraRedactPatterns=nil`, so it does not actually exercise the core guarantee of fix 421b2181e: that two facts whose raw values differ collapse to the same key *after* redaction-pattern normalization.

## Acceptance

- Add a sub-case with `extraRedactPatterns: []string{\`password=\\S+\`}` that stores an existing fact `password=secret-one` and verifies a new candidate `password=secret-two` is blocked as duplicate post-normalization.
- `go test ./...` green.

Parent: #480

duck8823:
## Summary

Strengthen \`TestMemoryExtractionUsecase_Extract_DeduplicatesExistingFactsAfterSanitization\` so it actually exercises the core guarantee of the v0.5.1 dedupe fix (#474 / 421b2181e): that two facts whose raw values differ collapse to the same key *after* a caller-supplied redaction pattern normalizes them.

Previously the test passed \`extraRedactPatterns=nil\`, so it only verified the dedupe path driven by built-in redactors.

## Test plan

- [x] Table-driven restructure with two sub-cases (built-in + custom).
- [x] Custom sub-case uses \`internalCode=\\S+\` — outside the built-in redactor set, so the custom pattern is the only thing that can make the two facts collide after sanitization.
- [x] \`go test ./...\`
- [x] \`go tool golangci-lint run\`

Closes #490

### DIFF
--- a/application/usecase/memory_extraction_usecase_impl_test.go
+++ b/application/usecase/memory_extraction_usecase_impl_test.go
@@ -460,12 +460,19 @@ func TestMemoryExtractionUsecase_Extract_DeduplicatesExistingFactsAfterSanitizat
 			// This sub-case exercises the core guarantee of the dedupe fix:
 			// two facts whose raw values differ must collapse to the same key
 			// *after* a caller-supplied redaction pattern normalizes them.
-			// `internalCode=...` is outside the built-in redactor set, so the
-			// custom pattern is the only thing that can make these equal.
+			// Both facts start with "Please" so inferMemoryTypeFromText
+			// actually produces a MemoryTypePreference candidate — an earlier
+			// iteration used "Remember ..." which matches no heuristic
+			// prefix, making the extractor emit nothing and the test pass
+			// for the wrong reason (caught by the Codex verifier). The
+			// `internalCode=` token is outside the built-in redactor
+			// alternation (password|secret|token|...), so only the
+			// caller-supplied pattern below can normalize the two values to
+			// the same key.
 			name:                "custom extra redact pattern",
 			extraRedactPatterns: []string{`internalCode=\S+`},
-			existingFact:        "Remember the deploy secret internalCode=alpha-one for next release.",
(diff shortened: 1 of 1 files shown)


duck8823:
## Multi-AI Review Summary

| Reviewer | Result |
|---|---|
| **Codex verifier (gpt-5.4 xhigh)** — first pass | HIGH: `Remember ...` prompt didn't match any `inferMemoryTypeFromText` heuristic, so the extractor produced zero candidates and the test passed for the wrong reason. |
| **Codex verifier** — after fix | ✅ no issues |
| **Claude reviewer** | ✅ 問題なし (verified the custom pattern is strictly outside the built-in redactor set and that breaking the dedupe fix makes the test FAIL) |

### Resolved

Rewrote the `custom extra redact pattern` sub-case to start both facts with `Please ...` so `inferMemoryTypeFromText` actually emits a `MemoryTypePreference` candidate. Verified with a local mutation — setting `extraRedactPatterns=nil` on the custom sub-case makes it FAIL with `len(Extract()) = 1, want 0`, proving the assertion is now sensitive to the custom-pattern dedupe path it was always meant to guard.

duck8823:
🤖 AI コードレビュー結果

- Codex verifier (gpt-5.4 xhigh) 1回目: HIGH 'Remember' prompt didn't trigger inferMemoryTypeFromText, test passed for wrong reason.
- Fix: rewrote to 'Please ...' prefix. Mutation verified (extraRedactPatterns=nil → sub-case FAILs).
- Codex verifier 2回目: no issues.
- Claude reviewer: 問題なし.

Mergeable.

PR #494, opened 2026-04-13
test: guard tail boundary dedupe at From==cursor.timestamp

### LINKED ISSUE #491: v0.5.1-9: add boundary test for tail cursor at From==cursor.timestamp
## Problem

`presentation/cli/tail.go:271` — tail follow mode relies on `cursor.isNew()` (seenIDs) to drop events whose `created_at` equals the cursor watermark. The existing tests do not explicitly cover the case where the first polled event's timestamp equals the From boundary of the next query. If the query boundary semantics ever drift (inclusive vs exclusive), the duplicate suppression depends entirely on seenIDs, which resets when the watermark advances.

## Acceptance

- Add a `tail_test.go` case where the initial fetch returns a single event whose timestamp is re-used as the next poll's From bound, and assert the event is not emitted twice.
- Document the inclusivity contract of `EventListCriteria.From` at the type or builder level (may overlap with #sibling if already addressed there — coordinate to avoid duplication).
- `go test ./...` green.

Parent: #480

duck8823:
## Summary

Add \`TestRunTail_DoesNotReEmitBoundaryEventWhenFromEqualsInitialTimestamp\` — an end-to-end \`runTail\` regression for #491. The scenario: initial fetch returns a single event at time T, follow poll uses \`From=T\`. Because \`EventListCriteria.From\` is inclusive, the query service re-returns the same event, and only the cursor's \`seenIDs\` dedupe keeps it from being emitted twice per tick.

Also lifts the From/To inclusivity contract into \`EventListCriteria\`'s struct-level doc comment.

## Stacking

Based on \`fix/tail-pagination-snapshot-consistency\` (PR #492) because the new test exercises \`runTail\` + \`pollTailEvents\`, which that PR introduces. Merge #492 first.

## Test plan

- [x] \`go test ./...\` — new test PASS
- [x] \`go tool golangci-lint run\` — 0 issues

Closes #491

### DIFF
--- a/application/types/event_list_criteria.go
+++ b/application/types/event_list_criteria.go
@@ -8,6 +8,13 @@ import (
 
 // EventListCriteria holds filter parameters for event listing.
 // Zero-value fields are ignored (no filter applied).
+//
+// Time range semantics (documented here so callers can reason about cursor
+// advancement without reading the SQL): From is inclusive and To is
+// exclusive. i.e. events with created_at == From are returned, events with
+// created_at == To are not. Tail-style cursors that poll repeatedly with
+// From set to the last seen timestamp must therefore dedupe returned
+// events against an already-seen set at the boundary.
 type EventListCriteria struct {
 	limit        int
 	offset       int
--- a/infrastructure/sqlite/event_store_test.go
+++ b/infrastructure/sqlite/event_store_test.go
@@ -616,3 +616,79 @@ CREATE TABLE command_audits (
 		t.Fatalf("got[0].EventID() = %q, want event-inside", got[0].EventID().String())
 	}
 }
+
+func TestDatasource_ListWindow_RespectsFromLowerBoundInclusive(t *testing.T) {
+	t.Parallel()
+
+	migrations := fstest.MapFS{
+		"000001_init.sql": {
+			Data: []byte(`
+CREATE TABLE events (
+    id TEXT PRIMARY KEY,
+    kind TEXT NOT NULL,
+    agent TEXT NOT NULL,
+    session_id TEXT NOT NULL,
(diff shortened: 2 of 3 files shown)


duck8823:
## Multi-AI Review Summary

| Reviewer | Result |
|---|---|
| **Codex verifier (gpt-5.4 xhigh)** — first pass | HIGH: stubbing `ListWindow` meant a SQL-layer flip from inclusive-From to exclusive would not trip the new runTail test |
| **Codex verifier** — after fix | ✅ no issues |
| **Claude reviewer** | ✅ 問題なし (ran a 20× mutation with the `seenIDs` filter disabled — every run failed the test, so no false-pass risk) |

### Resolved

Added `TestDatasource_ListWindow_RespectsFromLowerBoundInclusive` in `infrastructure/sqlite/event_store_test.go`, mirroring the pre-existing `…RespectsToUpperBoundExclusive` test: three events (before / at / after the boundary) pin the inclusivity contract at the SQL layer. A regression that flips `From` from `>=` to `>` now fails at the infra layer, and the end-to-end `runTail` test keeps guarding the cursor's `seenIDs` boundary dedupe at the presentation layer — together they fully constrain the contract.

duck8823:
🤖 AI コードレビュー結果

- Codex verifier (gpt-5.4 xhigh) 1回目: HIGH stubbed ListWindow doesn't pin SQL-layer From inclusivity contract.
- Fix: added TestDatasource_ListWindow_RespectsFromLowerBoundInclusive (before/at/after boundary points).
- Codex verifier 2回目: no issues.
- Claude reviewer: 問題なし (20× mutation with seenIDs disabled all failed, no false-pass).

Mergeable.

PR #688, opened 2026-04-24
fix(search): skip thinking-block text in transcript JSON envelopes

### LINKED ISSUE #682: v0.8.2: search_events matches thinking-block text inside transcript JSON envelopes
## Background
Since #662 (PR #680), transcript event bodies are stored as JSON envelopes with \`type=thinking\` / \`type=text\` blocks. \`ExtractPlainBody\` intentionally drops thinking blocks from read projections so memory extraction / search UIs do not surface internal reasoning.

However \`infrastructure/sqlite/sql/search_events.sql\` still runs \`e.body LIKE ?\` against the raw stored body. That means a search query that only matches a word inside a \`thinking\` block still returns the row — and the CLI/MCP will then render the same event with the thinking text stripped out. The user sees a matched event they cannot locate in the displayed body.

## Scope
- Make search skip matches that only hit \`thinking\` blocks (either post-filter in Go, or by maintaining a \`body_plain\` projection column, or via sqlite-json1 path extraction in the SQL).
- Add regression tests.

## Acceptance
- \`traceary search needle\` on a body \`{"blocks":[{"type":"thinking","text":"needle"}]}\` returns no match.
- Pre-#662 legacy plain-text rows remain searchable exactly as before.

Found by: Codex architect during v0.8.1 quality phase.

duck8823:
## Summary
Closes #682.

Since #662/#680, transcript bodies are stored as \`{"blocks":[{"type":"thinking","text":"..."},{"type":"text","text":"..."}]}\` JSON envelopes. The search SQL still ran \`e.body LIKE ?\` against the raw stored body, so a query that only matched text inside a \`thinking\` block returned the row — but CLI/MCP rendered it via \`ExtractPlainBody\` with thinking stripped, so the user saw a "match" they could not locate in the displayed body.

Fix: in \`search_events.sql\`, match against a CASE expression that extracts only text-block content when the body is a canonical envelope. Legacy plain-text rows and non-envelope JSON bodies keep matching verbatim.

## Test plan
- [x] \`go test ./...\` — 1022 passed / 16 packages (2 new test cases: thinking-only filtered, non-envelope JSON preserved)
- [x] \`golangci-lint run\` — 0 issues

### DIFF
--- a/infrastructure/sqlite/search_events_test.go
+++ b/infrastructure/sqlite/search_events_test.go
@@ -211,6 +211,58 @@ func TestDatasource_Search_SkipsThinkingOnlyMatches(t *testing.T) {
 	}
 }
 
+func TestDatasource_Search_PreservesNonCanonicalBlocksBody(t *testing.T) {
+	t.Parallel()
+
+	// Non-canonical envelope: has a "blocks" key but elements do not
+	// match the {type:string, text:string} shape. ExtractPlainBody
+	// returns the raw body for these, so search must keep the raw body
+	// searchable too — otherwise the search surface and the display
+	// surface disagree.
+	migrations := searchEventsMigrations()
+	dbPath := filepath.Join(t.TempDir(), "traceary", "traceary.db")
+	sut, storeManager := newEventDatasource(t, dbPath, migrations)
+	if err := storeManager.Initialize(context.Background()); err != nil {
+		t.Fatalf("Initialize() error = %v", err)
+	}
+
+	ws := "github.com/duck8823/traceary"
+	cases := []struct {
+		name string
+		id   string
+		body string
+	}{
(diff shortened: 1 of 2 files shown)


duck8823:
## 🤖 AI コードレビュー結果

### レビュアー
- Codex verifier: ✅ (2 rounds, final approve)
- Claude final: ✅

### 解消済み MUST
- envelope 判定が広すぎて non-canonical body (element shape 不整合) で SQL と ExtractPlainBody が不一致 → typeof() guard 追加で `decodeCanonicalEnvelope` と整合

### 検証
- `go test ./...` ✅ 1023 passed
- `golangci-lint run` ✅ 0 issues
- 回帰テスト: thinking-only / mixed-block / non-envelope JSON / non-canonical shape (4 cases) すべて想定通り

Ready to merge.

Issue #679, opened 2026-04-24
v0.8.1: expose source_hook on read side — CLI --source-hook filter, JSON field, retire body-prefix markers


duck8823:
Follow-up to #672 (write side merged via PR #678).

## Background

#672 added the `events.source_hook` column and plumbs hook runtime handlers to stamp it. Write side is done. Read side was deferred to keep the PR reviewable.

## Scope

1. **List / search filters** — `traceary list --source-hook stop`, `--source-hook pre_compact` etc. Apply to both CLI `list` and MCP `list_events`.
2. **JSON output** — include `source_hook` in `traceary list --json`, `traceary show`, MCP event serialisation so external consumers can distinguish hook sources.
3. **Replay HTML** — event rows render the hook name alongside kind / agent so operators can see at a glance whether a `session_ended` came from Stop, SessionEnd or SubagentStop.
4. **Retire body-prefix markers** once source_hook is surfaced:
   - `[phase:subagent]` on session_ended (introduced with #636)
   - `[phase:pre-compact]` on compact_summary (introduced with #636)

   Readers (`loadCompactSummary` etc.) currently parse those prefixes. After the switch, they filter on `source_hook` instead. The old rows still carry the prefix so migration keeps compatibility, but the write side stops emitting new prefixed bodies.

## Acceptance

- `traceary list --source-hook subagent_stop` returns only rows stamped subagent_stop (or matches legacy `[phase:subagent]` body-prefix rows during migration window).
- JSON output includes `source_hook` field; empty/NULL serialised as omitted.
- Replay HTML shows the hook name where relevant.
- After #662 (transcript JSON blocks) ships, body-prefix markers can be removed from the domain constants + write paths.

Parent: #562 (v0.8 umbrella)

Issue #672, opened 2026-04-22
v0.8.1: record the source hook event on each event (Stop vs SessionEnd vs SubagentStop, PreCompact vs PostCompact, Stop vs AfterAgent)


duck8823:
Tracked from v0.8 dogfooding (2026-04-22 Claude session).

## Finding

Every event written by a hook lands in the `events` table with:

```
id, kind, agent, session_id, body, created_at, client, workspace
```

There is **no field indicating which hook fired**. The kind field captures the logical category (note / command_executed / prompt / session_started / session_ended / compact_summary / transcript), but multiple hook events collapse into the same kind:

| Event kind | Possible source hooks |
|---|---|
| `session_ended` | Claude `SessionEnd`, Codex `Stop` (session-end side), Claude `SubagentStop` |
| `compact_summary` | Claude `PostCompact`, Claude `PreCompact` |
| `transcript` | Claude `Stop`, Codex `Stop` (transcript side), Gemini `AfterAgent` |
| `command_executed` | `PostToolUse`, `PostToolUseFailure`, Gemini `AfterTool` |

Current partial workarounds:

- Body prefix `[phase:subagent]` distinguishes SubagentStop from SessionEnd
- Body prefix `[phase:pre-compact]` distinguishes PreCompact from PostCompact
- `agent` field tells you the client (claude / codex / gemini) but not the specific hook

Many combinations still collapse. For example Claude `Stop` vs Gemini `AfterAgent` both produce a `transcript` event with `client=hook` — you have to infer from `agent` which client-side hook ran.

## Why it matters

- Debug: operators reviewing an event row cannot tell which hook originally fired it. Affects "did SubagentStop fire for this subagent?" queries.
- Analytics: no SQL query for "count how many PreCompact fired vs PostCompact" without body-prefix string matching.
- Provenance: future downstream consumers (replay, memory extraction) may want to filter by hook kind. Body-prefix substring matching does not scale.

## Proposed fix

Add a column `source_hook TEXT` (nullable, enumerated set like `stop`, `session_start`, `session_end`, `subagent_stop`, `pre_compact`, `post_compact`, `after_agent`, `after_tool`, `post_tool_use`, `post_tool_use_failure`, `user_prompt_submit`).

- Populated by the hook runtime at write time (e.g., `runHookTranscript` tags `stop` for Claude/Codex, `after_agent` for Gemini).
[...]

PR #690, opened 2026-04-24
fix(sqlite): use compound index for --source-hook filter

### LINKED ISSUE #683: v0.8.2: source_hook filter query plan does not use idx_events_source_hook
## Background
#672 added the partial index \`idx_events_source_hook ON events(source_hook) WHERE source_hook IS NOT NULL\` so \`traceary list --source-hook <name>\` could answer cheaply without a full scan.

#679 wired the filter by adding an OR chain to \`select_recent_events.sql\`:
\`\`\`
AND ( ? = '' OR e.source_hook = ? OR (? = 'subagent_stop' AND e.kind = 'session_ended' AND e.body LIKE '[phase:subagent]%') OR (? = 'pre_compact' AND e.kind = 'compact_summary' AND e.body LIKE '[phase:pre-compact]%') )
\`\`\`

SQLite's planner picks \`idx_events_created_at\` (for ORDER BY) instead of the source_hook index when the filter is part of an OR chain — confirmed via \`EXPLAIN QUERY PLAN\`. As event history grows this defeats the index's purpose.

## Scope
- Split the query: keep a fast path when \`source_hook != ''\` AND no legacy fallback needed (\`e.source_hook = ?\` as a single predicate so the planner can use the partial index).
- Either UNION ALL a legacy-match branch, or add a compound index \`(source_hook, created_at DESC, id DESC)\`.
- Measure with EXPLAIN QUERY PLAN + a synthetic 100k-row benchmark.

## Acceptance
- \`EXPLAIN QUERY PLAN\` for \`traceary list --source-hook pre_compact\` uses the source_hook index (or a new compound index) instead of scanning via \`idx_events_created_at\`.

Found by: Codex architect during v0.8.1 quality phase.

duck8823:
## Summary
Closes #683.

#679 wired \`traceary list --source-hook <name>\` via an OR chain in \`select_recent_events.sql\`, which prevented the planner from using \`idx_events_source_hook\` — it fell back to \`idx_events_created_at\` for the ORDER BY and filtered in memory.

Fix (3 moving parts):
- **Migration 000012** — replace \`idx_events_source_hook\` with a compound partial index \`(source_hook, created_at DESC, id DESC) WHERE source_hook IS NOT NULL\` so the primary branch is COVERED.
- **SQL split** — \`select_recent_events.sql\` keeps the no-filter path; new \`select_recent_events_by_source_hook.sql\` has \`e.source_hook = ?\` as a top-level conjunct plus a \`UNION ALL\` branch that matches pre-#672 legacy body-prefix rows (\`[phase:subagent]\`, \`[phase:pre-compact]\`).
- **Dispatch** — Go datasource (\`queryRecentEvents\` / \`queryRecentEventsTx\`) picks between the two queries based on whether \`sourceHook\` is empty. \`LIMIT\`/\`OFFSET\` applied to the combined set so pagination is stable.

## EXPLAIN QUERY PLAN (after migration)
\`\`\`
SELECT e.id FROM events e ... WHERE e.source_hook = ? ORDER BY e.created_at DESC, e.id DESC LIMIT 20
→ SEARCH e USING COVERING INDEX idx_events_source_hook_time (source_hook=?)
\`\`\`

## Test plan
- [x] \`go test ./...\` — 1034 passed / 16 packages (new test: \`TestDatasource_ListRecent_SourceHookFilterIncludesLegacyPrefixRows\` — new rows + legacy prefix rows both returned for \`pre_compact\`; \`stop\` filter excludes legacy compact rows)
- [x] \`golangci-lint run\` — 0 issues
- [x] EXPLAIN QUERY PLAN verified on a real migrated DB

### DIFF
--- a/infrastructure/sqlite/event_datasource.go
+++ b/infrastructure/sqlite/event_datasource.go
@@ -26,6 +26,9 @@ var insertCommandAuditQuery string
 //go:embed sql/select_recent_events.sql
 var selectRecentEventsQuery string
 
+//go:embed sql/select_recent_events_by_source_hook.sql
+var selectRecentEventsBySourceHookQuery string
+
 //go:embed sql/search_events.sql
 var searchEventsQuery string
 
@@ -201,20 +204,12 @@ func (d *EventDatasource) ListRecent(
 		toValue = formatTimestamp(to)
 	}
 
-	rows, err := db.QueryContext(
-		ctx,
-		selectRecentEventsQuery,
-		kind.String(), kind.String(),
-		client.String(), client.String(),
-		agent.String(), agent.String(),
-		sessionID.String(), sessionID.String(),
-		workspace.String(), workspace.String(),
-		boolToInt(failuresOnly),
-		fromValue, fromValue,
-		toValue, toValue,
-		sourceHook, sourceHook, sourceHook, sourceHook,
-		limit,
-		offset,
+	rows, err := queryRecentEvents(
+		ctx, db, sourceHook,
+		kind, client, agent, sessionID, workspace,
+		failuresOnly,
+		fromValue, toValue,
+		limit, offset,
 	)
 	if err != nil {
 		return nil, xerrors.Errorf("failed to query recent events: %w", err)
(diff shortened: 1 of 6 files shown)


duck8823:
## 🤖 AI コードレビュー結果

### レビュアー
- Codex verifier: ✅ (2 rounds, final approve)
- Claude final: ✅

### 解消済み MUST
- UNION ALL が全 sourceHook で走り non-legacy name (stop 等) が 17x 性能退行 → 3-way dispatch に分割 (primary-only / UNION ALL-with-legacy / no filter)

### 検証
- \`go test ./...\` ✅ 1034 passed
- \`golangci-lint run\` ✅ 0 issues
- EXPLAIN QUERY PLAN: COVERING INDEX idx_events_source_hook_time
- Codex bench: stop filter 20x は primary-only query で元の性能に復帰

Ready to merge.

PR #1281, opened 2026-07-13
v0.22.0-8: preserve hook events across host timeout kills

### LINKED ISSUE #1269: v0.22.0-8: prevent hook event loss under host timeout kills
## Background

Hook events can be lost when the host kills a hook process at its timeout, and two design choices make this likely on large databases:

1. **SQLite busy_timeout equals the smallest hook budget.** `infrastructure/sqlite/database.go:189` sets `busy_timeout` to 5000ms, and `infrastructure/filesystem/gemini_hooks_handler.go:10` gives every Gemini hook a 5000ms budget (same value shipped in `integrations/gemini-extension/hooks/hooks.json` and `examples/hooks/gemini.settings.json`). A single lock contention busy-wait can consume the entire hook budget, after which the host kills the process and the event is dropped. This was reproduced under burst concurrency on a 2.96 GB database (gemini MCP server + concurrent writers + parallel test queries) where a one-off session-end hook hit the 5s timeout.
2. **No SIGTERM handling in the hook path.** `main.go:237` only installs broken-pipe signal handling. When the host kills a hook at timeout, any in-flight write is abandoned; there is no journal, deferred write, or resend mechanism, so work that was nearly complete is discarded.

Failures are also fully silent: `runHookBestEffort` (`presentation/cli/hook_runtime.go:1759-1767`) swallows all errors into exit 0 (stderr only with `TRACEARY_HOOK_DEBUG`), and the wrapper scripts add another `>/dev/null 2>&1 || return 0` layer. Loss is only detectable after the fact, and only for Claude SessionEnd via the cancellation diagnostics (#1254 covers triaging those markers; this issue is about prevention).

## Scope

- Decouple the SQLite busy_timeout from the hook budget: either lower `busy_timeout` well below the smallest hook timeout so contention fails fast and can be retried/journaled, or raise the packaged Gemini hook timeout above the busy_timeout (e.g. 10s, matching the Antigravity handler rationale in `antigravity_hooks_handler.go:19`).
- Add SIGTERM handling to the hook path so an in-flight event is preserved (e.g. spool to an append-only journal file) instead of dropped; replay the journal on the next hook invocation or via a doctor fix.
- Consider extending timeout-kill detection beyond Claude SessionEnd (prompt/audit/transcript hooks, other clients) once prevention lands.

## Acceptance

- A hook killed at its host timeout no longer silently loses the event: the event is either committed, journaled and replayed, or surfaced as an actionable diagnostic.
- Packaged hook timeouts and busy_timeout are no longer equal; the relationship (which must exceed which, and why) is documented.
- Regression coverage exists for the contention scenario (lock held by a concurrent writer while a hook fires).

duck8823:
## Motivation

A host timeout can terminate a hook while SQLite is still waiting on a writer lock. Because the previous SQLite busy timeout equaled Gemini's host hook timeout and hook errors were fail-soft, the event could disappear without durable evidence.

## Changes

- atomically spool every hook payload before executing its runtime handler
- remove the mode-0600 spool record only after the handler succeeds
- cancel hook command contexts on SIGTERM while leaving the already-published record intact
- report retained and unreadable records through the new `hook-spool` doctor check
- reduce SQLite `busy_timeout` from 5 seconds to 1 second
- raise generated and packaged Gemini hook timeouts from 5 seconds to 10 seconds
- document the required `SQLite busy_timeout < host hook timeout` relationship

## Safety

- retained records are never deleted based on age
- doctor does not print payload bodies
- spool persistence uses temp-file + atomic rename under the existing local hook state root
- if the spool directory itself is unavailable, existing fail-soft hook execution remains unchanged

## Validation

- `go test ./...` (2,652 tests passed across 21 packages)
- `go tool golangci-lint run --timeout=5m`
- `git diff --check`
- contention test holds a concurrent SQLite writer and confirms the hook-like write returns before the 5-second host budget
- child-process test sends SIGTERM after the spool barrier and confirms the payload record survives
- dogfood failure injection retained one prompt record and `doctor` reported `hook-spool=warn`
- bilingual docs, changelog coverage, release manifests, and packaged integration verification passed

Closes #1269

### DIFF
--- a/docs/hooks/README.ja.md
+++ b/docs/hooks/README.ja.md
@@ -198,6 +198,12 @@ traceary hooks install --client codex --upgrade
 
 その場合はファイルを自分で確認し、本当に置き換えてよいときだけ `--force` を使ってください。
 
+### timeout kill 時の保全
+
+すべての非公開 `traceary hook ...` entrypoint は、SQLite に触れる前に host payload を `~/.config/traceary/hooks/spool/` 配下の mode `0600` record へ保存します。record は atomic に公開され、hook operation が成功した後だけ削除されます。SIGTERM は command context を cancel します。host が競合中または低速な hook を kill しても、先に公開した record が残るため event は黙って消えません。`traceary doctor` は残存 record と読めない record を `hook-spool` check で報告します。残存 record は payload を確認し、対応する command を再実行してから削除してください。経過時間だけでは自動削除しません。
+
+SQLite が writer 競合を待つのは最大1秒です。packaged host hook の budget は必ずこの待機時間を上回る必要があります。Gemini の生成・同梱 hook timeout は10秒で、DB attempt が失敗して spool record を保持したまま host deadline より前に制御を戻せます。この関係は意図的です: `SQLite busy_timeout < host hook timeout`。
+
 ## トラブルシュート
 
 hooks やローカル SQLite ストアの挙動がおかしいときは `traceary doctor --client <claude|codex|gemini>` を実行してください。
--- a/docs/hooks/README.md
+++ b/docs/hooks/README.md
@@ -198,6 +198,12 @@ After `hooks install`, Traceary prints the matching `doctor` command so you can 
 
 In those cases, inspect the file yourself and only rerun with `--force` when replacing the file is truly acceptable.
 
+### Timeout-kill preservation
+
+Every hidden `traceary hook ...` entrypoint writes the exact host payload to a mode-`0600` record under `~/.config/traceary/hooks/spool/` before it touches SQLite. The record is atomically published and removed only after the hook operation returns successfully. SIGTERM cancels the command context; if the host kills a contended or slow hook, the already-published record remains available instead of the event disappearing silently. `traceary doctor` reports retained or unreadable records through the `hook-spool` check. Inspect the payload and retry the matching command before removing a retained record; Traceary does not delete it based on age alone.
+
+SQLite waits at most 1 second for a busy writer. Every packaged host hook budget must exceed that wait. Gemini's generated and packaged hook timeout is 10 seconds, leaving time for the DB attempt to fail and the spool record to remain durable before the host deadline. This relationship is intentional: `SQLite busy_timeout < host hook timeout`.
+
 ## Troubleshooting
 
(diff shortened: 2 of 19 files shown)


duck8823:
## Multi-AI Review Results

### Reviewers
- Antigravity / Gemini 3.5 Flash: REQUEST_CHANGES → APPROVE after re-review of `381efce0`
- Claude Fable 5: unavailable (`quota_or_capacity`: usage credits exhausted)
- Grok: unavailable (`policy_or_permission_denied`: workspace session filesystem permission denied)
- Codex external verifier: unavailable (`command_failed`: local app-server sandbox initialization denied)
- Current orchestrator + deterministic verifier: APPROVE

### Findings triage
- MUST `main.go`: SIGTERM context detection only checked `args[1]` — adopted. `isHookCommandArgs` now recognizes `hook` after preceding arguments and has direct/global-flag/non-hook coverage. Antigravity re-review confirms the MUST is resolved.
- SHOULD `hook_spool_test.go`: test bypassed production argument detection — adopted through `TestIsHookCommandArgs`; the existing child-process SIGTERM test still proves the spool survives actual termination.
- NIT Antigravity transcript failure must retain the spool — already implemented by returning `transcriptErr` after completing turn-boundary work.
- MUST remaining: none.

### Validation
- `go test ./...`: 2,656 passed across 21 packages
- `go tool golangci-lint run --timeout=5m`: passed
- `git diff --check`: passed
- SQLite contention test: write fails within the host budget while the lock is held
- SIGTERM child-process test: retained payload survives
- dogfood failure injection: one retained record, `doctor hook-spool=warn`, payload not printed
- docs/i18n, changelog coverage, release manifests, and packaged integrations: passed
- CI at head `381efce0`: Docs, Test, Integrations passed; Lint running at review-comment time

### Structure-Behavior review
- Hook command handlers remain responsible for host-specific orchestration.
- The spool lifecycle is isolated behind one wrapper and one diagnostic scanner; core hook handlers do not contain persistence choreography.
- Deletion depends on observable handler success, never age.
- Raw local payloads are protected with mode 0600 and doctor emits metadata only.

### Verdict
APPROVE, pending final CI lint.

PR #1283, opened 2026-07-13
v0.22.0-10: detect and remediate duplicate Codex hook registration

### LINKED ISSUE #1271: v0.22.0-10: detect and remediate duplicate Codex hook registration
## Background

`docs/integrations/codex-plugin.md:70` warns that if the manual fallback hook entries remain registered when a Codex build with `plugin_hooks` enabled loads the packaged plugin hooks, "every session/prompt/transcript/audit event will be recorded twice". This is a real, currently reachable state: a dogfood environment has been observed with both manual hooks and plugin hooks registered simultaneously.

Today nothing detects or remediates this: doctor has checks for hook coverage gaps (missing hooks) but not for the opposite failure mode (double registration), and #1252 covers untrusted plugin hooks, not duplication.

## Scope

- Add a doctor check that detects Codex double registration (manual hook entries in the Codex config alongside active plugin hooks) and reports it as a WARN with the exact entries to remove.
- Offer `doctor --fix` remediation that removes the manual fallback entries when plugin hooks are confirmed active (leaving manual entries untouched when plugin hooks are unavailable).
- Document the migration step in `docs/integrations/codex-plugin.md` (upgrade path: manual → plugin).
- Optionally: consider write-side dedup (e.g. idempotency key per host event) as a defense in depth, as a separate follow-up if non-trivial.

## Acceptance

- `traceary doctor` flags the double-registration state with actionable output.
- `doctor --fix` (or a documented manual step) converges the environment to a single registration path.
- A regression test covers the detection logic.

duck8823:
## Motivation

Codex installations can retain Traceary manual fallback hooks after plugin-managed hooks become active, causing every matching host event to be recorded twice.

## Changes
- detect the confirmed dual-registration state when the Traceary plugin is enabled and `features.plugin_hooks = true`
- report the exact manual entries that must be removed
- make `doctor --fix` remove only Traceary-managed entries while preserving user hooks and top-level configuration
- retain manual fallback hooks when plugin hooks are false or unconfirmed
- document the manual-to-plugin migration in English and Japanese

## Validation
- `git diff --check`
- `go vet ./...`
- `go test ./...`
- `go tool golangci-lint run --timeout=5m` (0 issues)
- Windows compile-only: `GOOS=windows GOARCH=amd64 go test -exec=/usr/bin/true ./...`

Closes #1271

### DIFF
--- a/.ai/spec/1271.md
+++ b/.ai/spec/1271.md
@@ -1,0 +1,51 @@
+# Issue #1271: detect duplicate Codex hook registration
+
+## Structure-Behavior Design Note
+
+### Requirement summary
+- Detect the unsafe state where the Traceary Codex plugin has confirmed active `plugin_hooks` while manual Traceary entries remain in `~/.codex/hooks.json`.
+- Report a dedicated warning that names the manual entries to remove.
+- Make `doctor --fix` remove only Traceary-managed manual entries and preserve unrelated hooks and top-level configuration.
+- Do not remove fallback hooks when plugin hooks are disabled or their state cannot be confirmed.
+- Document the manual-to-plugin migration in English and Japanese.
+
+### Conceptual model
+| Concept | State | Behavior | Constraint / invariant |
+|---|---|---|---|
+| Codex plugin hook activation | plugin enabled plus explicit `features.plugin_hooks = true` | selects plugin-managed registration | absence of an explicit true value is not confirmation |
+| Manual Traceary registration | managed entries in `~/.codex/hooks.json` | invokes Traceary hook commands | non-Traceary entries are never owned by this remediation |
+| Duplicate registration | confirmed plugin activation plus manual registration | warns and offers cleanup | both conditions must hold |
+
+### Responsibility assignment
+| Responsibility | Owner | Reason to change | Not owner / reason |
+|---|---|---|---|
+| Parse Codex plugin activation | `doctor_codex_plugin.go` | Codex TOML contract changes | hook JSON inspector does not know Codex TOML |
+| Identify and remove managed JSON entries | `HooksInspector` filesystem adapter | hook document shape or managed-key recognition changes | doctor must not duplicate JSON parsing rules |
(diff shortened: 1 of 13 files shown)


duck8823:
## Multi-AI Review Results

### Reviewers
- Antigravity differential/structure reviewer: **APPROVE** after re-reviewing `e4ebdc70`
- Codex independent verifier: **APPROVE** after executing the full validation suite
- Current orchestrator integrator: **APPROVE**
- Claude Fable 5: unavailable (`quota_or_capacity`: usage credits exhausted)
- Grok: unavailable (`policy_or_permission_denied`: local session filesystem permission)
- External Codex CLI: unavailable (`command_failed`: local app-server initialization permission)

### Findings and resolution
- **MUST (resolved):** retained non-Traceary matcher/command extensions could be lost. Replaced typed reserialization with `json.RawMessage` filtering and added unknown-field regression coverage.
- **MUST (resolved):** cleanup used direct truncating writes. Added fd-pinned same-directory temporary writes, sync/close, atomic rename, permission preservation, and failure-preserves-original coverage.
- **SHOULD (resolved):** added the `plugin_hooks` unspecified regression case and propagated the caller context.
- **NIT (not adopted):** Antigravity suggested inner `MarshalIndent`; the outer `json.MarshalIndent(root, ...)` already formats nested `json.RawMessage`, so a second indentation pass adds no output value.

### Validation on latest head
- `git diff --check`: pass
- `go vet ./...`: pass
- `go test ./...`: pass (independent verifier: 2674 tests / 21 packages)
- `go tool golangci-lint run --timeout=5m`: 0 issues
- Windows compile-only (`GOOS=windows GOARCH=amd64 go test -exec=/usr/bin/true ./...`): pass
- Focused race tests for filesystem and doctor remediation: pass

### Structure review
- The doctor owns diagnostic policy, the application interface exposes the narrow remediation capability, and filesystem infrastructure owns JSON filtering and safe replacement.
- Tests assert observable behavior rather than call order or private implementation.

### Verdict
**APPROVE** — no unresolved MUST/SHOULD findings.

PR #1290, opened 2026-07-13
v0.22.0-5: automate stale active session GC

### LINKED ISSUE #1253: v0.22.0-5: automate stale active session GC
## Background

`traceary doctor` currently reports **5,627 active sessions older than 24h** (WARN `stale-active-sessions`). Dogfood DB evidence (last 14 days):

- codex: 5,104 of 5,119 sessions still open (host provides no session-end hook; Stop is a turn boundary)
- antigravity: 467 of 467 sessions still open (no session-end hook)
- claude: 32 of 508 open (SessionEnd hook mostly works)

The designed mitigation is manual `traceary session gc --stale-after 24h`, but the 5,627 backlog proves manual GC does not happen in practice. Stale actives degrade `session_status` / `get_context` default behavior (doctor says so explicitly) and inflate the sessions dashboard.

## Scope

Make stale-session GC happen without operator intervention. Candidate mechanisms (pick during design; they are not mutually exclusive):

1. **Opportunistic GC on write**: hook runtime triggers a bounded, rate-limited GC pass (e.g. at SessionStart, at most once per N hours) so any active host keeps the store tidy.
2. `doctor --fix` support for this check, consistent with the coverage-aware doctor direction from #1171.
3. Scheduled-task recipe in `docs/operations/scheduled-tasks*` as a fallback.

Constraints:

- GC must not close sessions that are genuinely long-running but active (respect recent-event activity, not just started_at age; codex multi-turn sessions were the subject of v0.21.0-1).
- Write-path additions must stay cheap (hooks run on every host event) and safe under concurrent hook invocations (WAL, per-process rate limit).

## Acceptance

- A fresh WARN-state store converges to pass=green within one day of normal host usage, with no manual gc invocation.
- Table-driven tests cover the activity-vs-age boundary conditions.

duck8823:
Closes #1253

## Motivation

Manual `traceary session gc` left thousands of Codex and Antigravity sessions open in dogfood data. Automatically closing by `started_at` alone would be unsafe, however, because a genuinely long-running session may still have recent events.

## Changes

- make stale-session count/apply queries activity-aware: an active session is stale only when it started before the cutoff and has no event at or after the cutoff
- run best-effort GC after normal hook session starts and Antigravity `PreInvocation`
- rate-limit successful passes to once every six hours per database with a cross-process file lease and durable success marker
- keep GC failures out of host hook failures and retry on the next eligible hook invocation
- update doctor messaging, CLI help, and bilingual operations docs; include a daily scheduled fallback for hook-disabled environments

## Safety

- no schema migration and no event deletion
- current session start is recorded before automatic GC, so it is protected by recent activity
- the apply query is idempotent and only sets `ended_at`
- removing the hook coordinator calls rolls back automation while retaining the safer manual activity predicate

## Verification

- `go test ./...`
- `go tool golangci-lint run`
- `go run ./cmd/repo-tooling docs verify-i18n`
- `python3 scripts/verify_release_manifests.py`
- `git diff --check`

All passed locally. Tests cover old/fresh session and old/recent-event boundaries, first/rate-limited/retry hook runs, and an end-to-end WARN-state convergence without closing a recently active long-running session.


duck8823:
## Final verification

- Antigravity review: approved. Its SHOULD finding was resolved by separating the 1-minute activity protection window from the 5-minute stale lease pruning window.
- Independent Codex verifier: no blocking findings on the activity-lease design and concurrent-session protection.
- Local validation at `b051fd2a`:
  - `go test ./presentation/cli -run 'Test.*(Activity|Lease|Garbage|Antigravity)' -count=1` — 81 passed
  - `go test ./...` — 2,687 passed
  - `go tool golangci-lint run` — no issues
  - `go vet ./...` — passed
  - `git diff --check` — passed
- Review packet: `/private/tmp/traceary-pr1290-final-review/packet.md` (`sha256:a6788a995469f57edb9fc2dd5beefff372530e21e552b1cbce944631530de1a9`)

Residual risks: system clock skew can affect lease freshness; unreadable activity state remains fail-safe and skips destructive GC. Claude review was stopped after a 401 authentication failure in accordance with the local failure policy; it was not retried.

duck8823:
Independent Codex final verifier: ✅ no regression at `b051fd2a`. The 1-minute protection decision is unchanged; only physical stale-lease deletion is delayed to 5 minutes. Targeted tests, race tests, and diff checks passed.

duck8823:
## Post-base-update final verification

Independent Codex verifier: ✅ **APPROVE** at `48b1838e`; no MUST/SHOULD/NIT findings.

- Linux CI exposed a real cross-process race: refreshing an existing activity lease with `os.WriteFile` briefly truncated it, allowing GC to observe an empty session ID and close an active session.
- The lease now publishes through a private same-directory temporary file (`0600`), close, and atomic rename.
- Independent verification passed the multi-process regression 200 times, targeted race tests, the full Go suite, fresh-cache lint, vet, integration/docs/release checks, and diff checks.
- CI run `29265216355`: Docs, Integrations, Lint, and Test passed.

Residual risks: lease freshness still depends on filesystem mtime/wall clock; atomic replacement is verified on macOS and Linux, the supported release targets.

PR #1530, opened 2026-07-24
fix: report missing Codex headless final turns

### LINKED ISSUE #1519: v0.32.1-10: report Codex headless final-turn coverage accurately
Parent epic: #1513

Found during refreshed-plugin dogfood.

## Problem

A new `codex exec --ephemeral` session using the 0.32.0 plugin emitted and persisted SessionStart and UserPromptSubmit, returned the expected assistant marker, but did not emit a visible Stop hook and persisted no transcript or session-end event for that session.

This may be a host-surface limitation rather than a Traceary ingestion defect, but current diagnostics should not imply complete final-turn coverage for that mode.

## Scope

- reproduce the exact Codex headless/ephemeral hook boundary matrix
- distinguish host non-emission from Traceary hook failure
- use an available stable local source only if it preserves privacy and session identity
- otherwise report the headless mode as partial coverage

## Acceptance criteria

- [ ] fixtures/live evidence identify which hooks Codex headless and interactive modes actually emit
- [ ] doctor does not infer Stop/transcript coverage from manifest presence alone
- [ ] a successful headless marker response is persisted when a supported stable source exists; otherwise the missing boundary is explicit
- [ ] no transcript scraping reads unrelated sessions or raw private history
- [ ] tests cover ephemeral and non-ephemeral headless modes without depending on private method call order

Priority: P1
Structure-Behavior risk: Medium.

duck8823:
## Motivation

Codex headless diagnostics could imply final-turn coverage from the installed manifest even when a prompt had no observed final boundary.

## Changes

- compute body-free prompt and uncovered-final-turn counts
- report partial coverage when a final boundary is absent
- add version-scoped interactive/headless/ephemeral fixtures

Closes #1519


duck8823:
Merge authorization recorded: the maintainer explicitly accepts the residual risk that the independent external reviewer family did not return a valid final verdict. Local independent review and same-head CI are recorded; merge proceeds under that explicit exception.

PR #1561, opened 2026-07-25
v0.33.0-3: avoid full body hydration in bounded MCP reads

### LINKED ISSUE #1552: v0.33.0-3: avoid full event-body reads for bounded MCP retrieval
## Problem

The MCP bounded-body response limit is applied after full event bodies have been selected, restored, and converted in Go. It caps emitted tokens but not database-to-process body I/O or allocation.

## Scope

- add a bounded event projection that avoids loading full stored bodies when only a prefix is needed
- preserve redaction, truncation provenance, and full/detail semantics
- verify behavior on a representative large-store fixture without reading private payloads in test evidence

## Acceptance criteria

- [ ] default bounded MCP retrieval does not select full body columns when the prefix is sufficient
- [ ] metadata, bounded, and full projections remain behaviorally distinct and documented
- [ ] body truncation provenance remains correct
- [ ] query-plan or instrumentation tests prove the bounded path
- [ ] existing compatibility tests pass

Priority: P0
Structure-Behavior risk: High (storage/query boundary and public output semantics).

duck8823:
## Motivation

Default bounded MCP retrieval truncated only after SQLite returned complete stored bodies and Go hydrated domain events. Large persisted payloads therefore still incurred full database-to-process transfer and allocation.

## Changes

- add a dedicated bounded event read model, usecase, and read-only SQLite hydration path
- select metadata or search candidate IDs before projecting a visible-text prefix in SQLite
- preserve canonical transcript visible-text semantics and independent response/ingest/storage truncation provenance
- keep canonical `body_blocks` as an explicit untruncated `list_events` fallback; omit the fallback from `search` and `get_context`
- route metadata, bounded, and full MCP projections before any event query

## Validation

- `go build ./...`
- `go test ./...`
- `go tool golangci-lint run`
- `go vet ./...`
- `git diff --check origin/main...HEAD`

Closes #1552

### DIFF
--- a/.ai/spec/1552.md
+++ b/.ai/spec/1552.md
@@ -1,0 +1,118 @@
+# #1552 Body-free bounded MCP event retrieval
+
+## Structure-Behavior Design Note
+
+### Requirement summary
+
+- **Purpose:** Make the default bounded `list_events`, `search`, and
+  `get_context` projections cap SQLite-to-Go body transfer and allocation, not
+  only the final MCP JSON.
+- **Current behavior:** MCP resolves an omitted body option to `bounded`, but
+  still calls the full `EventUsecase`. SQLite selects every complete stored
+  body, Go restores domain `Event` values, and presentation truncates only
+  after canonical visible-text extraction.
+- **Expected behavior:** Presentation selects `metadata`, `bounded`, or `full`
+  before issuing a query. Bounded reads return a dedicated application read
+  model whose visible body is truncated in SQLite. Search uses FTS only to
+  select event IDs and hydrates bounded content from authoritative event rows.
+  Full/detail reads retain the existing domain `Event` path.
+- **Compatibility:** The bounded visible-text projection preserves
+  `ExtractPlainBody` semantics for canonical transcript envelopes, legacy
+  plain text, foreign/malformed JSON, thinking exclusion, and text-block
+  joining. Response truncation and persisted ingest/storage truncation remain
+  separate facts.
+- **Explicit fallback:** `list_events` may fetch a full raw body only for a
+  canonical envelope whose visible-text response was not truncated, solely to
+  preserve `body_blocks`. `search` and `get_context` never perform this
+  fallback.
+- **Out of scope:** Schema migration, retention-policy changes, restoration of
+  payload bytes already removed at ingestion/storage, CLI projection changes,
+  or changing full event/detail semantics.
+
+### Conceptual model
+
+| Concept | State | Behavior | Constraint / invariant |
+| --- | --- | --- | --- |
+| Metadata projection | Body-free event facts | Returns identity, attribution, extent, and audit outcome | No stored body text crosses the query boundary |
+| Bounded event | Metadata plus a visible-text prefix and total visible rune count | Reports response truncation independently from persisted truncation | Returned prefix never exceeds the requested positive rune limit |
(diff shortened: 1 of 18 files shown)


duck8823:

I checked the new SQL projection line-by-line against `ExtractPlainBody` because that is where a bug would be invisible. The parity is good: non-object JSON, `{"blocks":null}`, a capitalized `Blocks` key, non-string `type`/`text`, the empty envelope, thinking-only bodies, and blank-text-block skipping all agree, and the whitespace code-point list matches `unicode.IsSpace` exactly. I also confirmed the two SQLite assumptions the projection rests on:

```
sqlite> SELECT length(substr('あいうえ',1,2));            -- 2   (code points, not bytes)
sqlite> SELECT typeof(group_concat(t, X'0A0A')) ...;      -- text (BLOB separator does not coerce)
```

Findings below.

### 1. `group_concat` ordering is not guaranteed (highest-value item)

`select_bounded_event_bodies.sql` relies on `ORDER BY CAST(key AS INTEGER)` inside the feeding subquery. SQLite documents the concatenation order of `group_concat()` as arbitrary; an ordering in the subquery is not a contract. Here the order decides which runes land in the visible-text *prefix*, so an out-of-order concat silently returns the wrong first N runes rather than failing.

`modernc.org/sqlite v1.53.0` embeds SQLite **3.53.2** (`lib/sqlite.go:4516`), so the ordered-aggregate form added in 3.44 is available:

```sql
group_concat(block_text, X'0A0A' ORDER BY CAST(key AS INTEGER))
```

Note this is not a regression — `search_event_metadata.sql` / `search_events.sql` have no ordering at all — but those use the concat as a match predicate, whereas this one is the returned value.

### 2. The canonical-envelope predicate is duplicated and the two copies differ

The 8-line `json_valid` / `json_type` / `NOT EXISTS` block appears twice, and they are not identical: the `canonical_envelope` copy also requires `body_availability = 'available'`, the `visible_body` copy does not. That is safe today only because migration `000026` constrains the column to `available` / `unavailable_retention` and the first `CASE` branch consumes the latter. A third availability value would make the two silently disagree. One CTE column would remove the coupling.

### 3. Hardcoded ellipsis

`convertBoundedEvents` (`presentation/mcpserver/server.go`) appends a literal `"…"`, while every other truncation path uses `apptypes.TruncationEllipsis` (`application/types/command_truncation.go:36`). Same value today; using the constant keeps the two paths from drifting.

### 4. Additive field growth on the default path — intentional?

Bounded rows now emit `body_stored_bytes` / `body_ingest_truncated` / `body_storage_truncated`, which the previous default `list_events` response did not. The schema description was updated correctly, so this is additive and safe, but it adds bytes to every default response in a sprint whose goal is shrinking them. Worth confirming it is deliberate.

### 5. Scope wording for #1552

SQLite still materializes the *complete* visible text (`group_concat` across all blocks) before `substr` applies. The PR body's claim is accurate as written — the database→process transfer and Go-side hydration are genuinely gone — but for the large-store case the per-row SQLite-side cost is unchanged. Worth one line in the issue so #1552 is not later read as "large bodies are no longer touched at all".

### 6. Minor

`LoadCanonicalBodies` opens its own read-only connection, so the fallback runs outside the caller's read snapshot. The revalidation makes this safe, and the doc comment says so — just noting the boundary is real rather than incidental.

Merge order: #1565 is stacked on this branch, so this lands first.

PR #1562, opened 2026-07-25
docs: stage history retrieval guidance

### LINKED ISSUE #1555: v0.33.0-6: require staged token-efficient retrieval in bundled skills
## Problem

Bundled Traceary retrieval skills recommend history tools without requiring a metadata-first, small-limit discovery step. Default bounded responses can therefore consume unnecessary host context.

## Scope

- update bundled skills and MCP documentation to require staged retrieval
- start with metadata plus small limits and narrow workspace/time filters
- request bounded bodies only for selected events, then explicit detail only when necessary

## Acceptance criteria

- [ ] session-history and related skills document the staged retrieval algorithm
- [ ] examples use `projection=metadata`, small limits, and narrowing filters first
- [ ] examples show when `bounded` and full/detail retrieval are justified
- [ ] docs do not encourage raw transcript or broad history reads as a first action

Priority: P1
Structure-Behavior risk: Low (documentation and prompt contract).

duck8823:
## Motivation

Prevent broad event-body reads by making staged metadata discovery, bounded inspection, and explicit detail retrieval consistent across all packaged skills.

## Validation

- go test ./...
- go tool golangci-lint run
- go build ./...
- go run ./cmd/repo-tooling integrations verify
- go run ./cmd/repo-tooling docs verify-i18n
- git diff --check

Closes #1555

### DIFF
--- a/.ai/spec/1555.md
+++ b/.ai/spec/1555.md
@@ -20,9 +20,9 @@
 
 | Concept | State | Behavior | Constraint / invariant |
 | --- | --- | --- | --- |
-| Discovery | Initial history question | Lists a small metadata-only candidate set | Scope by workspace and, when known, time or session; no body fields |
-| Inspection | Candidate IDs known | Reads a small body prefix only for relevant candidates | A positive 300–500 rune `body_limit`; never a broad full-body list |
-| Detail | One event selected | Retrieves the full stored event for a stated reason | Single event or equivalently narrow scope; CLI `traceary show` is preferred |
+| Discovery | Initial history question | Lists a small metadata-only candidate set | `list_events` accepts workspace/time/session filters; `search` accepts workspace/time/query but no `session_id`; no body fields |
+| Inspection | Narrow session/time scope known | Reads a small body prefix only for relevant candidates | A positive 300–500 rune `body_limit`; the example names its target tool; `get_context` narrows only by workspace/session/limit and has no `event_id` |
+| Detail | One event selected | Retrieves the full stored event for a stated reason | No MCP history read accepts `event_id`; CLI `traceary show` is the explicit event-detail path |
 | Session recap | Operator requested recap | Applies the same staged rules before summarising | Does not issue a bare `get_context` call |
 
 ### Responsibility assignment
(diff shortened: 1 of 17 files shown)


duck8823:

The staged Discovery → Inspection → Detail framing is the right shape, and the 6-copy byte parity plus bilingual pairing is enforced properly. My findings are all doc-accuracy issues: the guidance describes parameters that the tools do not actually accept. I checked against the tool registry on `main`:

```
get_context    [body_limit, full_body, limit, projection, session_id, workspace]
list_events    [agent, body_limit, client, from, full_body, kind, limit, offset,
                projection, session_id, source_hook, timezone, to, workspace]
search         [body_limit, from, full_body, limit, projection, query, timezone, to, workspace]
```

### 1. `get_context` cannot be scoped to a selected event

`get_context` has no `event_id` parameter. Both the skill ("Use `get_context` only with this selected and bounded scope", "after an event or equivalently narrow scope is selected") and `docs/mcp/README.md` ("`get_context` belongs here only after an event or equivalently narrow scope is selected") read as though an agent can point it at a chosen event. It can only be narrowed by `workspace` / `session_id` / `limit`. Suggest phrasing it purely as the session-or-time-slice case so an agent does not go looking for an argument that is not there.

### 2. `search` has no `session_id`

Discovery says "Add a time range or `session_id` when the question provides one", and then "Use this shape with `list_events`, or add a narrow literal `query` for `search`." An agent following that will send `session_id` to `search`. Recommend marking the session filter as `list_events` / `get_context` only.

### 3. The Inspection example does not name its tool

The Discovery block explicitly says which tools accept the shape; the Inspection block does not, and its example contains `session_id`, so it is valid for `list_events` / `get_context` but not `search`. Worth one clarifying line.

### 4. The semantic validators pin prose, including typography

`validateSessionHistorySkillContract` requires literal markers such as `` `300`–`500` `` (en dash) and `"limit": 5`. A harmless rewording will fail CI with "must include staged retrieval contract marker" rather than anything semantic. And `validateMemoryReviewRecapContract`'s negative check matches one exact historical sentence — a reworded reintroduction of bare-`get_context` guidance passes it. That is an acceptable trade for a docs guard; I am naming the limitation rather than objecting to it.

Nothing here blocks merge — items 1–3 are small text edits, and this PR is independent of #1561 / #1565 since `projection` and `body_limit` already exist on `main`.

PR #1563, opened 2026-07-26
test: budget MCP tool schema advertisements

### LINKED ISSUE #1557: v0.33.0-5: budget MCP tool-schema size and measure loading behavior
## Problem

The MCP tool registry contains multi-action schemas whose cumulative size can consume context before any Traceary data is requested. The actual host impact depends on tool discovery/loading behavior, which is not currently measured or budgeted.

## Scope

- establish deterministic per-tool and total schema-size budgets
- measure supported hosts' tool-loading behavior without prompts, transcripts, or credentials
- split or profile tools only if measurement proves the current registry materially harms startup/context usage

## Acceptance criteria

- [ ] golden/CI check reports total and per-tool schema bytes with a reviewed budget
- [ ] body-free host observations record whether schemas are eagerly or lazily loaded
- [ ] any registry split preserves current public operations or has a compatibility plan
- [ ] documentation records the measured limitation and chosen mitigation

Priority: P2
Structure-Behavior risk: Medium (public MCP discovery contract).

duck8823:
Closes #1557

## Motivation
Make MCP tool-schema size observable and budgeted before it consumes host context.

## Validation
- go build ./...
- go test ./...
- go tool golangci-lint run
- go vet ./...
- go run ./cmd/repo-tooling integrations verify
- go run ./cmd/repo-tooling docs verify-i18n


duck8823:

Measuring the runtime `tools/list` response rather than the source schema is the right call, and separating a hand-edited policy limit from a generated report is exactly the distinction that keeps a budget meaningful. The `isTracearyMCPCommand` extraction is a genuine improvement — the table test covering `["mcp-server", "--profile=read"]` pins the thing the docs promise.

### 1. The budget starts inside the warning band

`tool_schema_budget.golden.json` records `listToolsBytes: 41028` against `maxListToolsBytes = 45056` — that is **91.06%**, so `classifyToolSchemaBudget` returns `budgetWarning` and CI logs the WARNING on the very first run, with 4,028 B of headroom. A warning that fires from day one carries no signal; the next real growth event looks identical to the steady state.

Two reasonable resolutions: raise the hard limit so 90% is genuinely an early warning, or state plainly in `docs/operations/mcp-tool-schema-budget.md` that the surface currently sits in the warning band and what that means. Either is fine — silently shipping at 91% is the part I would avoid.

(The input-schema aggregate is comfortable: 13,636 / 18,432 = 74%.)

### 2. Cross-PR: #1565 will move both numbers

#1565 adds `continuation` to the input schemas of `list_events`, `search`, and `get_context`, and adds `coverage` / `partial` / `reasons` / `continuation` to the shared `eventsOutput` — which all three tools use, so the output-schema growth lands three times. #1561 also touches the registry golden.

Whichever of #1563 / #1565 merges second needs to regenerate both fixtures. The per-tool caps have room (`list_events` 5706/6500, `search` 5187/6000, `get_context` 4891/5500), so the number to re-check is the 45,056 aggregate at 91%.

### 3. Minor

- CI runs `TestServer_ToolAdvertisementBudget` twice: once as the dedicated `-v` reporting step and again inside `go test ./...`. Deliberate and cheap, since `t.Logf` output is only visible with `-v` — noting it so it does not read as an accident later.
- `TestServer_ToolAdvertisementBudget` correctly omits `t.Parallel()` because of the fixture write path.
- The `len(maxToolAdvertisementBytes) != len(report.Tools)` guard catches a stale budget entry for a removed tool, and the `!ok` branch catches a new tool with no budget. Both directions covered — good.

PR #1564, opened 2026-07-26
fix: classify exact hook denial audits

### LINKED ISSUE #1526: v0.33.0-8: classify hook-denied command audits
## 背景

dotfiles の PreToolUse hook（`check-gh-commands.sh`、`check-repo-scope.sh`、`business_data_guard.py` 等）は、拒否時に **stderr 先頭が `🚫 [hook]` で始まるメッセージ + exit 2** という deny 出力契約を持つ（owner: dotfiles、契約の正本は `conventions/ai/agent-hooks-observability.md`「hook deny の出力契約と計測経路」）。

一方 Traceary の `command_audits.failure_reason` は 2026-07-01 以降 106,860 件すべて `unknown` で、hook deny が 1 件も分類されていない。過剰ブロックの頻度を定量評価できない。

## 入力契約

- 対象イベント: command audit（PreToolUse hook 経由のコマンド実行）
- deny の判定: 終了コード 2 かつ stderr 先頭行が `🚫 [hook]` で始まる
- 分類: `failure_reason = 'hook_denied'`（新設）にマッピングする。既存の `unknown` との区別が付くこと

## 完了条件

- [ ] 上記契約に合致する command audit が `hook_denied` として記録・集計できる
- [ ] 契約に合致しない exit 2（stderr 接頭辞なし）が `hook_denied` に誤分類されない
- [ ] 回帰テスト（契約合致 / 不合致の 2 経路）

## 関連

- duck8823/dotfiles#243（hook の過剰ブロック観測。dotfiles 側の分類修正と出力契約の定義）
- dotfiles 側契約の検証: `tests/test_gh_command_policy.py::test_hook_deny_output_contract`（hook 実起動で stderr 接頭辞 + exit 2 を固定）

duck8823:
Closes #1526

## Motivation
Classify the documented dotfiles hook-denial contract without exposing audit payloads.

## Validation
- go build ./...
- go test ./...
- go tool golangci-lint run
- git diff --check

### DIFF
--- a/.ai/spec/1526.md
+++ b/.ai/spec/1526.md
@@ -4,7 +4,7 @@
 
 - Classify a hook audit as `hook_denied` only when the captured structured
   result has `tool_response.exitCode == 2` and a string `tool_response.stderr`
-  begins byte-for-byte with `🚫 [hook]`.
+  begins byte-for-byte with Traceary's own hook-runtime marker, `🚫 [hook]`.
 - Keep all other exit-code and failure classifications unchanged. In
   particular, do not trim, search, or stringify stderr or payload values.
 - Do not add a new copy of raw stderr or payload material to audits, reports,
--- a/.ai/spec/1526.md
+++ b/.ai/spec/1526.md
@@ -1,0 +1,50 @@
+# #1526 Hook-denial audit classification
+
+## Requirement summary
+
+- Classify a hook audit as `hook_denied` only when the captured structured
+  result has `tool_response.exitCode == 2` and a string `tool_response.stderr`
+  begins byte-for-byte with `🚫 [hook]`.
+- Keep all other exit-code and failure classifications unchanged. In
+  particular, do not trim, search, or stringify stderr or payload values.
+- Do not add a new copy of raw stderr or payload material to audits, reports,
+  cursors, or logs. Existing audit redaction and truncation remain the sole
+  storage policy.
+
+## Conceptual model
+
+| Concept | State | Behavior | Invariant |
+| --- | --- | --- | --- |
+| Hook denial evidence | absent or exact | Refines an exit-code-2 audit to `hook_denied` | Both typed fields and the exact byte prefix are required |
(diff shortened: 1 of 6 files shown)


duck8823:
**Claude (Opus 5) review — approve**

This is the cleanest of the five. The predicate is placed correctly in `hookPayloadFailureReason`: after the `is_interrupt` / `timed_out` checks and before the generic exit-code fallback, so a timeout that also carries the marker still classifies as `timeout` rather than being reclassified as a denial. The negative cases are the ones that matter and they are all there — non-string stderr, a leading newline before the marker, exit code 3, and exit code 0. The `strings.HasPrefix` on a type-asserted string with no trimming matches what the spec promises.

Three notes, none blocking.

### 1. The leak assertion is keyed on a test-case name

```go
if tc.name == "hook stderr marker with exit code two is hook denied" {
```

Branching on `tc.name` inside a table-driven loop is a coupling that breaks quietly if the name is ever reworded, and it does not extend to a second marker case. A `wantNoRawStderrLeak bool` field on the case struct keeps the table self-describing and makes the check reusable. The assertion itself is well chosen — counting the raw marker occurrences to prove exactly one pre-existing `tool_response` copy is a much stronger check than a bare `!strings.Contains`.

### 2. `"🚫 [hook]"` is an inline literal in three places

It appears in `hookPayloadHasHookDeniedMarker`, in each test payload, and in both `docs/hooks/contract.md` files. Since this is a byte-exact protocol constant, a named `const hookDenialMarker` in `hook_resolve.go` referenced from the test would make the contract greppable and keep code and test from drifting independently.

### 3. Documentation framing

`docs/hooks/contract.md` now states the rule alongside the Grok `PermissionDenied` mapping, which reads as though it were a host-published contract. It is the Traceary-managed hook runtime's own marker, so a third-party hook author will not emit it and will not get `hook_denied`. One sentence saying the marker is Traceary's own emission — opt-in for anyone writing their own hooks — would prevent that being read as a general host capability.

The report-level regression (`TestReportUsecaseGenerate_HookDeniedReasonIsBodyFree`) proving the marker never reaches the serialized report is a good addition; that is the failure mode worth pinning.

Issue #1551, opened 2026-07-25
v0.33.0-4: enforce aggregate MCP response budgets


duck8823:
## Problem

MCP `list_events`, `search`, and `get_context` accept an arbitrary positive `limit`. Their compatibility controls also permit `full_body=true`, so one request can materialize an unbounded number of full event bodies.

This can consume SQLite I/O, process memory, MCP transport capacity, and the host agent's context budget together.

## Scope

- define explicit maximum result-count and aggregate response-body budgets for event retrieval tools
- preserve a safe, paginated path for callers that need more history
- make limit/body-budget truncation observable in structured output
- keep event-detail retrieval explicit rather than silently returning a huge payload

## Acceptance criteria

- [ ] oversized `limit` and `full_body` requests have deterministic bounded behavior
- [ ] response metadata reports count/body-budget truncation distinctly
- [ ] pagination or event-ID detail retrieval preserves access to omitted records
- [ ] regression tests cover default, maximum, and adversarially large requests
- [ ] tool documentation states the effective limits

## Non-goals

- changing stored event bodies
- disabling explicit detail retrieval

Priority: P0
Structure-Behavior risk: High (public MCP contract and host compatibility).

duck8823:
## Delegation assignment

- Design: GPT-5.6 Sol (max) — aggregate budget model, MCP compatibility, and observable truncation behavior.
- Implementation: GPT-5.6 Terra — scoped budget enforcement and tests in a dedicated worktree.
- Review: Antigravity scout for MCP-surface drift; Codex verifier for contract tests, build, test, lint, and structure review.

Gate: limits must be explicit and deterministic; no silent partial result may be reported as complete.

duck8823:
## Review-lane correction

Implementation remains GPT-5.6 Terra. The required independent reviewer is Claude Opus 5. Codex test/lint/contract execution is implementation validation only and is not counted as an independent review. Antigravity is optional scout only.

PR #1565, opened 2026-07-26
feat: add bounded MCP response continuation

duck8823:
Implements the #1551 aggregate response budget and same-snapshot keyset continuation. Validated with go vet, go test, golangci-lint, and git diff check. Closes #1551.


duck8823:
**Claude (Opus 5) review — request changes**

The design is sound: fingerprint binding over tool + normalized filters + projection + limits, the `continuation` ⊗ `offset` rejection enforced independently at the usecase, datasource *and* presentation layers, `EventPageAnchorOf` rejecting zero values, machine-readable `reasons`, and an authenticated opaque cursor. The `(created_at_norm, id)` tuple is the correct key for `ORDER BY created_at_norm DESC, id DESC`, and pinning a resolved upper bound so concurrent inserts cannot shift later pages is the right invariant.

Three findings I think should be resolved before merge, then some smaller ones.

---

### 1. The keyset predicate defeats the index it exists to use

The anchor is added as a sentinel disjunction:

```sql
AND (? = '' OR e.created_at_norm < ? OR (e.created_at_norm = ? AND e.id < ?))
```

Because the sentinel is a bound parameter, SQLite cannot treat any of it as a range constraint, so an anchored page degrades to a full descending index **SCAN** that discards rows until it walks past the anchor. Page N then costs O(N × limit) — precisely what keyset pagination exists to avoid, on the store size this sprint targets. Verified locally against the same index shape:

```
-- current sentinel form
|--SCAN e USING INDEX idx_events_created_at_norm_id_desc

-- bare predicate when the anchor is present
|--SEARCH e USING COVERING INDEX idx_events_created_at_norm_id_desc (created_at_norm<?)

-- row-value form
|--SEARCH e USING COVERING INDEX idx_events_created_at_norm_id_desc ((created_at_norm,id)<(?,?))
```

`metadataTimeRangeQuery` already solves exactly this problem for `from` / `to` — it rewrites `(? = '' OR ...)` into a bare `AND e.created_at_norm >= ?` when the value is present, specifically to keep the index seek. `metadataPageQuery` should do the same; it is already rewriting `LIMIT ? OFFSET ?` in that function, so the hook exists. `modernc.org/sqlite v1.53.0` embeds SQLite 3.53.2, so the row-value form is available and is the clearest expression of the tuple comparison.

Related: `TestGeneralMetadataListAndContextQueryPlansUseNormalizedTimestampIndexes` only asserts plans for the **unanchored** case (`apptypes.EventPageAnchor{}` / `""` args). An anchored case in that table would have caught this, and is the regression guard I would want alongside the fix.

While there: `metadataTimeRangeQuery` matches against the named constants `metadataOptionalFromPredicate` / `metadataOptionalToPredicate`, whereas `metadataPageQuery` matches the inline literal `"LIMIT ? OFFSET ?"`. All ten SQL files happen to contain it verbatim today, but a reformat makes `strings.Replace` a silent no-op while `metadataLimitOffsetArgs` still drops the offset argument — an arity mismatch at runtime. A named constant matches the existing convention.

### 2. The membership query runs twice per bounded/full page

`loadEventPage` calls `loaders.candidates` (→ `eventMetadata.List`) to select the page, then calls `loaders.bounded` (→ `eventBounded.List`) with `len(returnedCandidates)` and the same offset. After #1561, `ListRecentBounded` internally re-runs `queryRecentEventMetadataTx` before hydrating bodies. So one default `list_events` issues the filtered+ordered metadata query twice, on two separate read-only connections — and with finding 1, each of those is a SCAN once a continuation is in play.

#1561 already exposes the right seam: `hydrateBoundedEvents(ctx, tx, metadata, bodyRuneLimit)` takes the selected metadata. Passing the already-chosen candidates into body hydration removes the second pass and also closes the two-snapshot window. The shared `to` upper bound does cover concurrent inserts, so this is currently a cost problem rather than a correctness one — but the anchor is built from `returnedCandidates` while the bodies come from an independently re-executed query, so the two only agree by construction, not by contract.

### 3. Body-budget exhaustion drops bodies that continuation will never return

`applyEventAggregateBudget` appends *every* event. Once `used` reaches the budget, `truncateEventForAggregateBudget(event, <= 0)` nils `Body` and `BodyBlocks` but the event is still emitted and still counted in `returned_count`. The continuation is then anchored at the **last** returned event, so following it resumes *after* the body-less ones.

Concrete case: 20 events whose bodies total 80 KiB. Events ~15–20 come back as metadata-only, `partial=true`, `reasons=["aggregate_body_budget"]`, and the continuation starts at event 21. Those six bodies are unreachable through the paging contract — only a fresh per-ID call retrieves them. `returned_count == candidate_count` in that response, so `coverage` does not reveal the gap either.

`partial=true` currently means two different things with two different recovery paths, and the anchor only helps `more_results`. I would expect either: stop including events once the budget is exhausted and anchor at the last event that kept its body; or keep the current shape but have `returned_count` count only body-carrying events and make the distinction explicit in `coverage`.

---

### Smaller items

**4. The new hard caps are not advertised.** `NewEventResponseBudget` makes `limit > 100` and `body_limit > 16384` hard errors, but the `limit` schema still reads only `"result limit (default: 20)"` and `body_limit`'s description is unchanged. A client that previously sent `limit=200` now gets an error with no discoverable cause. The maxima belong in the jsonschema strings, not only in `docs/mcp/response-budget.md` — the schema is what the agent actually reads.

**5. Ellipsis cliff in the binary search.** `truncateEventForAggregateBudget` finds the largest prefix that fits, *then* appends `"…"`, then drops the body entirely if that pushed it over. A body one byte from the limit becomes `null` instead of one rune shorter. Folding the ellipsis into the search predicate removes the cliff. (Also, `"…"` is hardcoded here rather than `apptypes.TruncationEllipsis` — same note as on #1561.)

**6. Capability sniffing in the application layer.** `event_usecase_impl.go` selects the page implementation with `u.eventQuery.(queryservice.EventPageQueryService)`. A stub that does not implement it silently takes the old path, so a test can pass against behavior production never executes. `server.go` already models the alternative well with explicit `WithEventBounded` injection.

**7. Cursor lifetime error message.** `loadEventContinuationAEAD` mints a random key per process via `sync.OnceValues`, so every server restart makes outstanding cursors fail as `"continuation is malformed or has been modified"`. `docs/mcp/response-budget.md` documents that tokens are process-scoped, but the error text reads as tampering. A distinct message — "issued by a previous server process; restart paging" — would save real debugging time on the most common cause.

**8. Double decode.** `resolveEventContinuationSnapshot` and `resolveEventPageRequest` each decrypt and unmarshal the same token independently, two AEAD opens per request. Easy to fold into one decode.

**9. Response-size interaction with #1563.** `coverage` and `partial` are non-`omitempty`, so all three tools grow every response by a constant ~120 B, and the shared `eventsOutput` schema growth lands three times in `tools/list`. #1563 pins `listToolsBytes` at 41,028 / 45,056 = 91% of the hard limit, so please re-check the regenerated budget after rebasing.

---

Merge order: stacked on #1561, so that one lands first.

duck8823:
**Claude (Opus 5) re-review of `12b4b532` — previous blockers resolved**

All nine items from my earlier review are addressed, and the three I flagged as blocking are fixed at the right layer rather than papered over. Details of what I verified:

### 1. Keyset seek — fixed, with the regression guard

The sentinel disjunction is replaced by a `/* traceary:event-page-anchor */` marker substituted with `AND (e.created_at_norm, e.id) < (?, ?)`, both named constants. `TestAnchoredMetadataPageQueryPlanUsesCompositeKeysetSeek` asserts the plan literally contains `(created_at_norm,id)<(?,?)`, and `TestMetadataPageQueriesRenderCompositeAnchorWithoutSentinelOrOffset` proves all ten queries carry exactly one marker and render without sentinel or `OFFSET`. Making the unanchored case a SQL comment also means a missing marker degrades to a valid query rather than an argument-arity mismatch. This is a better fix than what I suggested.

### 2. Double membership query — fixed

`HydrateBounded(ctx, metadata, bodyRuneLimit)` performs no membership query, and `hydrateBounded` asserts both count and per-index event-ID identity against the supplied page, so a silent drift becomes an error rather than a wrong page. `TestEventBoundedUsecase_RejectsHydrationThatChangesCandidateOrder` pins it.

### 3. Body-budget prefix — fixed

`applyEventAggregateBudget` now breaks instead of emitting body-stripped events, `hasUnreturned` drives the continuation, and the anchor comes from the last *emitted* event rather than the last candidate. `TestAggregateBudgetContinuationRecoversFirstUnreturnedEvent` covers exactly the case I described. Progress is guaranteed — the first event of a page always has the full 64 KiB available, so a page can never come back empty with a continuation.

### 4–8 — all addressed

Maxima are in the jsonschema strings (with `event_input_schema_test.go` pinning them), the ellipsis is folded into the binary-search predicate and uses `apptypes.TruncationEllipsis`, `EventReadQueryService` makes page-aware reads a construction-time requirement instead of a runtime type assertion, the authentication error now names the server-restart case, and `resolveEventContinuation` decodes once.

### 9 — resolved on the #1563 side

I test-merged this branch into #1563 locally to check the interaction rather than guess at it. The recalibrated bands there (`warning 48 KiB / hard 52 KiB`, with a comment stating they account for this PR's schema expansion) hold comfortably:

| | merged | warning | hard |
|---|---:|---:|---:|
| `tools/list` | 44,753 | 49,152 | 53,248 |
| input schemas | 14,130 | 16,384 | 18,432 |
| `list_events` | 6,969 | — | 8,192 |
| `search` | 6,418 | — | 7,680 |
| `get_context` | 6,122 | — | 7,168 |

**One mechanical note for whoever merges second:** the merge conflicts in `presentation/mcpserver/testdata/tool_registry.golden.json` (both branches add `outputSchema` to the snapshot). `tool_registry_snapshot_test.go` auto-merges cleanly. Resolving with a regeneration is sufficient — after

```sh
go test ./presentation/mcpserver -run TestServer_ToolRegistrySnapshot -update
```

`go build ./...` and `go test ./...` both pass on the merged tree, and the budget test passes without needing its fixture regenerated.

---

### Remaining non-blocking items

**a. The binary search is non-monotonic at the top end.** Adding the ellipsis when `middle < len(runes)` means `f(len)` (no ellipsis) can be *smaller* than `f(len-1)` (with ellipsis), so the predicate is not monotonic and the search can miss the untruncated case. Concretely, with a 100-rune ASCII body and `available = overhead + 101`: the search probes 99 → `overhead+102` → false → `high=98`, settles on `best=98`, and returns a truncated body even though the full 100 runes would have fit. It never exceeds the budget and `body_truncated` / `body_length` are set honestly, so this is suboptimal rather than wrong — but a `middle == len(runes)` check before the loop (or seeding `best` from the untruncated body) removes it for free.

**b. The legacy branch cannot report `more_results`.** `NewServer` requires `eventBounded` but not `eventMetadata`, so the `!metadataAvailable` fallback is constructible. That branch fetches exactly `ItemLimit()` items with no `+1` probe and hardcodes `hasMore=false`, so a full page never emits `more_results` or a continuation even when more rows exist — paging silently ends early. `main.go:217` wires `WithEventMetadata`, so this is configuration-only today; validating it in `NewServer` next to the `eventBounded` check would make the degraded path unreachable rather than silent.

**c. Cosmetic.** `metadataPageQuery` now uses the named `metadataPageAnchorMarker` but still matches the inline `"LIMIT ? OFFSET ?"` literal. The new rendering test asserts no `OFFSET` survives in any of the ten queries, so the risk I raised is now covered by test — only the constant asymmetry remains.

**d.** The EXPLAIN-plan assertion covers the workspace-scoped query; the other nine get string-level checks. Given the predicate is textually identical across all ten, that seems like the right amount of coverage.

None of a–d blocks merge. CI is green on `12b4b532` (Test, Lint, Docs, Integrations). Still stacked on #1561, so that lands first.

Issue #1556, opened 2026-07-25
v0.33.0-7: bound memory-hygiene scan and similarity work


duck8823:
## Problem

Memory hygiene loads every matching memory and compares candidate pairs within a scope. Similarity and duplicate suggestion work can grow quadratically and emit an unbounded suggestion set.

## Scope

- define bounded scan/result behavior for large memory scopes
- introduce deterministic prefiltering, continuation, or both without hiding actionable findings
- preserve accepted/candidate trust separation and dry-run semantics

## Acceptance criteria

- [ ] scan work and returned suggestions have explicit per-invocation bounds
- [ ] continuation is deterministic and does not duplicate or skip candidates
- [ ] duplicate and validity-overlap semantics remain covered by tests
- [ ] large-scope benchmark/regression test guards against quadratic blowups
- [ ] operator output explains incomplete/continued scans

Priority: P1
Structure-Behavior risk: Medium (query/use-case behavior and operator workflow).

duck8823:
## Delegation assignment

- Design: GPT-5.6 Sol (max) — bounded scan and similarity model, privacy boundary, and performance invariants.
- Implementation: GPT-5.6 Sol — scoped memory-hygiene changes in a dedicated worktree.
- Review: Claude independent reviewer for privacy and algorithmic boundary; Codex verifier for benchmarks, tests, lint, and structure review.

Gate: work is bounded by rows, bytes, and time; incomplete scans are surfaced rather than inferred complete.

duck8823:
## Review-lane correction

Implementation remains GPT-5.6 Sol. The required independent reviewer is Claude Opus 5. Codex benchmark and test execution is implementation validation only and is not counted as an independent review.

duck8823:
Implemented and merged via PR #1567 (merge commit 6641c6d7): bounded scans, authenticated MCP continuations, best-effort revision reporting, and fail-closed Apply revalidation.

PR #1567, opened 2026-07-26
fix: bound and resume memory hygiene scans

duck8823:
## Motivation\n\nLarge durable-memory hygiene scans must be explicitly bounded and resumable so they do not transfer or retain an unbounded store snapshot.\n\nCloses #1556\n\n## Changes\n\n- add revision-stable, cursor-resumable bounded scan phases\n- expose finite CLI/MCP budgets, completion state, and safe previews\n- revalidate only requested memory IDs and all same-scope peers before Apply; partial or stale revalidation fails closed before mutation\n\n## Validation\n\n- go build ./...\n- go test ./...\n- go tool golangci-lint run\n- go test -race ./application/usecase ./infrastructure/sqlite -run MemoryHygiene

### DIFF
--- a/.ai/spec/1556.md
+++ b/.ai/spec/1556.md
@@ -11,18 +11,22 @@
   similarity checks. `apply` repeats that full scan before changing selected
   IDs. CLI and MCP return complete raw facts without a response-size bound.
 - **Expected behavior:** Each invocation has explicit row, scanned-byte,
-  returned-suggestion-byte, comparison, and wall-time ceilings. A partial
-  result carries an opaque continuation cursor and a stable stop reason.
-  Continuing an unchanged store visits every phase/key exactly once.
+  returned-suggestion-byte, comparison, and wall-time ceilings. MCP partial
+  results carry an opaque continuation cursor and a stable stop reason.
+  Standalone CLI scans are deliberately single-invocation and report bounded
+  re-run guidance instead of exposing an unusable cross-process cursor.
+  Continuing an unchanged MCP scan visits every phase/key exactly once.
 - **Cursor trust:** A process-owned AES-GCM key authenticates and encrypts every
   continuation. The scan creation time is selected by the CLI/MCP boundary and
   restored only from authenticated cursor plaintext. Legacy checksum cursors,
   modified cursors, and cursors from an earlier process require a new scan.
 - **Concurrent-write behavior:** A revision change no longer discards progress
-  back to phase one. It returns a partial `best_effort` result with
-  `revision_changed`, preserving the keyset in a continuation rebound to the
-  current revision. The consistency downgrade remains visible on every later
-  page. Apply does not use this relaxation.
+  back to phase one. The use case binds the current revision, permanently
+  downgrades the chain to `best_effort`, and retries the same source page
+  inside the original invocation and budget. A successful retry reports the
+  actual bound that stops it. `revision_changed` is returned only when revision
+  churn consumes the invocation duration before the page can advance. Apply
+  does not use this relaxation.
 - **Pair semantics:** Exact duplicates remain exhaustive and emit one
   suggestion per duplicate member. Similarity and validity-overlap detection
   exhaustively visit every unordered same-scope pair across continuations,
@@ -48,14 +52,14 @@
 
 | Concept | State | Behavior | Constraint / invariant |
 | --- | --- | --- | --- |
-| Memory revision | Monotonic integer in a singleton SQLite table | Increments transactionally after every memory insert/update/delete | A mismatch returns an explicit best-effort continuation at the same keyset and current revision instead of restarting phase one |
+| Memory revision | Monotonic integer in a singleton SQLite table | Increments transactionally after every memory insert/update/delete | A mismatch rebinds and retries the same page as best-effort without restarting phase one; only budget exhaustion may return a continuation at that keyset |
 | Scan budget | Positive row, scan-byte, result-byte, comparison, and duration limits | Accounts for each source unit and returned suggestion | Zero-value caller input resolves to finite defaults; no unlimited mode |
 | Scan phase | `accepted_rows`, `exact_duplicates`, `similarity_pairs`, `candidate_rows` | Owns one deterministic kind of traversal | Accepted and candidate phases never share status filters |
 | Keyset | Last completed memory ID, or current anchor/peer position for a pair phase | Resumes strictly after completed work | Contains identities/counters only, never fact text |
 | Opaque cursor | Encrypted version, memory revision, criteria digest, fixed scan time, phase, keyset, and consistency state | Restores deterministic continuation | AES-GCM authentication, payload shape, and criteria must validate before any row is read; keys never leave the process |
 | Source unit | One row, exact-duplicate anchor, or prefiltered unordered pair | Advances one observable keyset step | Unit byte extent is known before raw text is returned |
 | Similarity pair traversal | Every unordered same-scope pair, ordered by anchor ID then partner ID | Stops and resumes at `(anchor_id, partner_after_id)` | A completed continuation chain has no approximation, duplicate, or skipped pair; each invocation remains bounded even though exhaustive total work can be quadratic |
-| Scan result | Suggestions, counters, resource usage, completion, stop reason, consistency, consistency reason, next cursor | Makes partial and mixed-revision coverage operator-visible | `complete=true` has no cursor; `best_effort` remains sticky after the first revision change |
+| Scan result | Suggestions, counters, resource usage, completion, stop reason, consistency, consistency reason, and an MCP continuation | Makes partial and mixed-revision coverage operator-visible | `complete=true` has no cursor; `best_effort` remains sticky after the first revision change; CLI suppresses cursor input/output |
 | Targeted revalidation | Requested row plus exact peer and bounded neighboring peers | Recomputes current suggestion immediately before apply | It never scans unrelated scopes or every memory |
 | Safe preview | Sanitized fact prefix plus truncation marker | Gives enough operator context to identify a suggestion | Raw fact bytes are absent from public result and cursor |
 
@@ -69,7 +73,7 @@
 | Read revision and deterministic bounded source units | SQLite `MemoryHygieneScanSource` | Transactions, indexes, keysets, byte lengths, and peer ordering are storage concerns | Generic `MemoryQueryService.List` is not a safe scan primitive |
 | Maintain revision and hygiene indexes | additive migration 33 and triggers | Store mutation and index compatibility must be transactional | Runtime migration code must not rewrite memory facts |
 | Revalidate selected IDs and mutate via existing lifecycle methods | hygiene use case plus narrow source query | Apply owns orchestration; existing domain/repository methods own status transitions | Scan remains read-only and cannot share a boolean apply switch |
-| Render completeness, usage, cursor, and previews | CLI/MCP presentation | Protocol shape and localized text belong at the edge | Presentation does not receive raw facts |
+| Render completeness, usage, continuation/guidance, and previews | CLI/MCP presentation | MCP owns resumable paging while each standalone CLI process owns one bounded scan and re-run guidance | Presentation does not receive raw facts; CLI never exposes cursor input/output |
 
 ### Boundaries / interfaces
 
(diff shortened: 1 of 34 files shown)


duck8823:
**Claude (Opus 5) review of `bd798569`**

The four-phase traversal, the revision-stable page contract, and `Apply` re-validating every requested ID at one revision before the first mutation are all sound. Three things I'd want resolved before merge.

### 1. A scan can hand back a cursor that makes no progress

`application/usecase/memory_hygiene.go:143-167`: inside the unit loop, `keyset` is only advanced at `:167`, *after* the result-byte check at `:159-161`. If the first unit of an invocation produces suggestions larger than the remaining result budget, the function returns at `:160` with `keyset` still equal to the cursor that was passed in — zero suggestions, `stop_reason=result_byte_limit`, and a continuation identical to the input. Resuming reproduces it exactly, forever.

Reachable because `max_result_bytes` is client-settable down to 1: `presentation/mcpserver/server.go:1359-1360` copies the input verbatim and `application/types/memory_hygiene_scan.go:53` only rejects `< 1`. Under the 256 KiB default a single unit cannot exceed it (previews are capped at 240 bytes), so this needs a deliberately small budget — but "bounded and resumable" ought to mean the first unit of every invocation is always admitted. That is precisely the guarantee #1565 makes for the first event of a page, and it's the reason a page there can never come back empty with a continuation. The in-loop time check at `:144-146` has the same shape with `max_duration_ms=1`.

Either admit the first unit unconditionally and charge the overflow to the next one, or treat "returned cursor == input cursor" as an error.

### 2. The `ScanAt` guard never runs, and the checksum isn't an authentication boundary

`:78-80` rejects a cursor whose `ScanAt` disagrees with `criteria.Now` — but only when `criteria.Now` is non-zero, and neither production caller sets it: `presentation/mcpserver/server.go:1265-1270` and `presentation/cli/memory_hygiene.go:219-225` both construct `MemoryHygieneScanCriteria` without `Now`. So `:81` takes `now` from the client's cursor verbatim, and `now` is the reference for staleness classification at `:292`.

The cursor checksum is an unkeyed SHA-256 truncated to 16 bytes (`memory_hygiene_cursor.go:126-133`), so anyone can compute a valid one for any payload; `subtle.ConstantTimeCompare` at `:82` gives it the shape of authentication without the substance. A hand-written cursor therefore picks which memories look stale.

The blast radius is limited, and I checked before writing this: `Apply` recomputes `now` from `time.Now()` (`:565-568` — neither caller sets `Now` there either), so a forged `ScanAt` cannot drive a mutation; the affected memory simply fails with "no current hygiene suggestion for this memory". So this misleads a review UI rather than deleting anything.

Still, #1565 in the same sprint authenticates its continuation with AES-GCM under a process key. Shipping two cursor formats in one release with different trust models should be a deliberate decision, not an accident of authorship. Either way, the dead `criteria.Now` comparison should go or `Now` should be wired at both call sites — as written it reads like a check that runs.

### 3. A global revision counter versus hook-driven memory writes

`schema/sqlite/migrations/000033_add_memory_hygiene_revision.sql` bumps one singleton on every `memories` insert/update/delete regardless of scope. A bump mid-scan makes the source return `ErrMemoryHygieneRevisionChanged`, which `Scan` converts into a hard error at `:134` — no partial result, no cursor, restart from phase 1.

That is the correct default for correctness. The concern is that this repo's own memory extraction writes during normal agent activity, so on a store large enough to need pagination the scan may not have a window in which to finish, and each retry hits the same wall. Nothing in the diff bounds how often the stated goal is unreachable. Worth either narrowing the counter (per scope kind) or offering an explicit best-effort resume that reports "revision changed, results span two snapshots" instead of failing.

### Non-blocking

- `idx_memories_hygiene_exact` indexes `fact` itself, duplicating every fact's text into the index. A hash column gives the same exact-duplicate lookup at a fraction of the size — relevant given this sprint's store-size motivation.
- `MaxResultBytes` is accepted down to 1 while `MaxRows` requires ≥ 2 (`memory_hygiene_scan.go:47-55`). The asymmetry is what makes item 1 reachable.
- The criteria digest is computed after defaults are applied (`:88`), so changing `--expiry-days` mid-pagination is correctly rejected rather than silently mixing two definitions of stale. Good.
- `validMemoryHygieneCursorKeyset` (`memory_hygiene_cursor.go:96-124`) enforcing phase-specific keyset shape plus the `AnchorMemoryID < AfterPartnerID` invariant makes most malformed cursors non-decodable regardless of the checksum question. Also good.

CI is green (Test / Lint / Docs / Integrations) on `bd798569`.

duck8823:
**Claude (Opus 5) re-review of `3560a9e6`**

All three items are addressed, and two of them better than I proposed. Three new things, one of which is blocking.

### What's resolved

**Non-progressing continuation** — `24acafc1` rejects rather than papers over it: `finishPartialMemoryHygieneResult` now compares the outgoing `(phase, keyset)` against the values the invocation started with and returns `ErrMemoryHygieneContinuationCannotProgress` naming the budget that stopped it. That is a better answer than "always admit the first unit", because it tells the caller their budget is unusable instead of quietly exceeding it. The `finishPartial` closure also removes the twelve-argument call sites, which is a real readability win.

**Cursor authentication** — `3560a9e6` replaces the unkeyed checksum with AES-GCM under a per-process key, with `traceary:memory-hygiene-cursor:v2` as additional data and a version bump to 2, so a v1 cursor no longer decodes. `ErrMemoryHygieneRescanRequired` names the restart case explicitly. This is now consistent with #1565, which was the point.

**Revision changes** — the `MemoryHygieneRevisionChangedError` / consistency-downgrade design is more careful than what I suggested. Preserving the keyset while permanently marking the chain `best_effort`, carrying the downgrade inside the cursor so it survives every later page, surfacing `consistency` / `consistency_reason` on both CLI and MCP, and keeping `Apply` failing closed via `Unwrap` — that separates "the read may span snapshots" from "the write must not" exactly where it belongs. The best-effort anchor-vanished handling in `scanMemoryHygienePairs` (treat the old anchor as completed rather than restart the phase) is the right recovery, and the comment explains why.

### Blocking: CI Lint is red

Run [30241760788](https://github.com/duck8823/traceary/actions/runs/30241760788) fails with three `wrapcheck` violations on the new constructor:

- `infrastructure/sqlite/memory_hygiene_scan_source.go:81`
- `infrastructure/sqlite/memory_hygiene_scan_source.go:423`
- `application/usecase/memory_usecase_import_test.go:104`

The code these replaced used `xerrors.Errorf("%w", queryservice.ErrMemoryHygieneRevisionChanged)`, which is why it passed before. Wrapping the constructor call the same way keeps `errors.As(&revisionChanged)` working, since `As` unwraps.

### 1. The CLI ships a `--cursor` flag its own process can never honour

`docs/cli/README.md` states it outright: "A standalone CLI command starts a new process for each invocation, so its cursor cannot be reused by a later standalone invocation; use the MCP surface when a scan requires multiple process-authenticated pages."

So `traceary memory admin hygiene scan` prints `next_cursor=…` and accepts `--cursor`, and passing the value it just printed back to the next invocation always fails. On `main` the CLI has no continuation surface at all (`git grep continuation -- presentation/cli` is empty), so #1565's process-bound key was only ever used where a long-lived process makes it work. This PR is the first to put one on a one-shot command.

Documenting a dead flag is not the same as not shipping one. Two ways out, both cheap:

- **Persist the key in the store.** `raw_body_retention_store_identity` already establishes the pattern of a random per-store row. Anyone who can read that key can already read the memories directly, so keeping it in the same database costs nothing against the actual threat model (a caller fabricating a cursor to steer classification) and makes CLI pagination work.
- **Or drop `--cursor` and `next_cursor=` from the CLI** and print the stop reason with "re-run with a larger budget", so the surface stops advertising something it cannot do.

I'd take the first — a bounded scan whose only paginating client is MCP is a smaller feature than the issue asked for. Either way this is a UX call, so it's yours; I'm flagging that the current state ships a control that never works.

### 2. A revision change costs a whole round trip and can repeat forever

On `ErrMemoryHygieneRevisionChanged` the use case rebinds the revision and returns immediately (`memory_hygiene.go`, the `errors.As` branch) — zero rows scanned, zero suggestions, a new cursor. The source's top-level `ExpectedRevision` check at `memory_hygiene_scan_source.go:80-82` is *not* relaxed for `best_effort`, so the next invocation compares against the freshly bound revision and rebinds again if another write landed in between.

That means with one memory write per client round trip — which is what hook-driven extraction on an active session looks like — the chain never advances. It is strictly better than the old hard error, since the keyset survives; but it is also harder to notice, because a fresh nonce makes every returned cursor different, so a client cannot detect "I got the same position back". The non-progress guard from `24acafc1` is explicitly bypassed for this case, by design.

Retrying the page in-loop after rebinding — `continue` instead of `return`, still charged to the same budget — would make a revision change cost one extra page read instead of an entire round trip, and would only surface `stop_reason=revision_changed` when the budget genuinely ran out. The consistency downgrade would still be permanent and still be reported.

### 3. Minor

`loadMemoryHygieneCursorAEAD` is a `sync.OnceValues`, so a key-generation failure is cached for the process lifetime — every subsequent scan returns the same error with no recovery short of restart. `crypto/rand` failing is close to unheard of, so this is a note, not a request.

Everything except the Lint failure is green (Test / Docs / Integrations) on `3560a9e6`.

duck8823:
**Claude (Opus 5) re-review of `acf164d5` — all three items resolved**

### Lint

Green. The three `wrapcheck` sites now wrap the constructor with `xerrors.Errorf("…: %w", …)`, and the added context strings ("revision changed before reading the source page" / "pair anchor changed before continuation") distinguish the two mismatch origins in the error text. `errors.As` still reaches `*MemoryHygieneRevisionChangedError` through the wrap, which is what the retry branch depends on.

### 1. CLI cursor — removed rather than documented

`--cursor` is gone from the flag set, `next_cursor` is gone from both the text and JSON output, and `rerun_guidance` takes its place on partial results. That closes the gap: the CLI no longer prints a value it cannot consume, and the guidance names the three real options (narrow `--workspace`, raise the finite bounds, or use MCP for paging). `docs/cli/README.md:632-635` states the split plainly.

Worth noting for anyone reading later: this does not cap what the CLI can do, because `MemoryHygieneScanBudgetFrom` enforces only lower bounds — a CLI user can raise `--max-scan-rows` / `--max-duration` arbitrarily and complete a large scan in one invocation. The CLI gives up pagination, not completeness.

### 2. Revision changes now retry in-loop

The rebind happens inside a retry loop around `ScanMemoryHygienePage`, bounded by the same `scanCtx` deadline and `budget.MaxDuration()`, and `stop_reason=revision_changed` is reserved for churn that actually consumed the duration before any page progress. A successful retry reports whichever bound later stops the invocation, which is the honest answer. The `revisionChangedWhileRetrying` flag reclassifying a subsequent deadline as `revision_changed` rather than `time_limit` is the right attribution — it tells the caller *why* the budget went.

I checked that the permanent `best_effort` downgrade is accurate rather than merely conservative, in both paths that can reach it:

- **Mid-invocation** — page 1 is read at R, page 2 requests `ExpectedRevision=R`, a write lands, the retry reads page 2 at R'. The chain genuinely spans two revisions.
- **On a continuation** — units from earlier invocations were read at R, this invocation resumes at R'. Same.

A fresh scan cannot reach it at all, since `ExpectedRevision` is unset for its first page. So the flag is never set for a scan that actually saw one snapshot.

### Non-blocking

**a. The retry loop is a busy-wait with no attempt cap or backoff.** Failed attempts charge nothing against rows/bytes/comparisons, so the only exit is the clock. Each attempt is a full `sql.Open` + `PingContext` + `Close` (`infrastructure/sqlite/database.go:115-137`, called per page), which means sustained write churn can turn one 2-second default budget into a large number of connection open/close cycles against a store that is already under write pressure. It terminates and it converts into progress the moment a window opens, so this is a politeness issue rather than a correctness one — but a small backoff, or an attempt counter surfaced in `usage`, would keep a hot store from being hammered and would make the churn visible instead of inferred from `stop_reason`.

**b. Residual, and correctly documented:** with a deliberately tiny `--max-duration` / `max_duration_ms`, the first attempt can mismatch and the deadline check can fire before any retry, returning a rebound cursor at the same `(phase, keyset)`. The non-progress guard permits that because the revision changed, so a client can still loop without advancing. `docs/cli/README.md:625-627` describes exactly this case, and the default 2 s budget absorbs it, so I don't think it needs code — noting it so the behaviour isn't rediscovered as a bug.

**c. Carried forward, unchanged:** `loadMemoryHygieneCursorAEAD` is a `sync.OnceValues`, so a `crypto/rand` failure at first use is cached for the process lifetime. Still a note, not a request.

CI is green on `acf164d5` (Test / Lint / Docs / Integrations). Nothing blocking from me.

One process note: #1568 is based on this branch, so its committed evidence artifact is now three commits stale against its own base. Covered in my comment there.

Issue #1558, opened 2026-07-25
v0.33.0-9: dogfood bounded retrieval on a multi-GB store


duck8823:
## Scope

Validate the v0.33.0 bounded retrieval changes on a multi-GB metadata fixture and a body-free dogfood store observation.

## Acceptance criteria

- [ ] run `TRACEARY_RUN_MULTI_GIB_BENCHMARK=1 go test ./infrastructure/sqlite -run '^$' -bench BenchmarkMetadataDirectRangeMultiGiB -benchtime=1x` on a host that allows completion
- [ ] record benchmark environment, `page_count * page_size`, event count, non-NULL body metadata count, `SUM(body_stored_bytes)`, target, and measured p95
- [ ] list, context, and search meet the documented bounded p95 targets
- [ ] MCP aggregate response budgets and truncation metadata are observable
- [ ] FTS/backfill and index migrations are healthy and rollback guidance is verified
- [ ] staged retrieval examples work across supported host integrations
- [ ] no event body, prompt, transcript, credential, or spool payload is placed in QA evidence

Priority: P0 release QA
Structure-Behavior risk: Medium (verification only; it validates high-risk child changes).

duck8823:
## Delegation assignment

- Design: GPT-5.6 Sol (max) — multi-GB acceptance matrix, measurement method, and release evidence boundary.
- Implementation: Codex QA — reproducible fixture, body-free measurement harness, and verification artifact in a dedicated worktree.
- Review: Claude independent reviewer for evidence quality and release-risk assessment; second Codex verifier for reproducibility and payload exclusion.

Gate: use an opt-in copied or synthetic store; never attach event bodies, prompts, transcripts, credentials, or raw production rows to the Issue or reviewer packet.

duck8823:
## Review-lane correction

Implementation remains Codex QA. The required independent reviewer is Claude Opus 5. Codex reproducibility checks are implementation validation only and are not counted as an independent review.

duck8823:
## Planning update — Claude Fable 5

**Wave:** 0 / pre-tag.
**Dependency:** #1574 must be corrected first.
**Release evidence:** independently review PR #1572 against multi-GB p95, migration/rollback evidence, and fixture representativeness before merge.
**Exit:** merge PR #1572 and close this Issue before the v0.33.0 tag.

PR #1568, opened 2026-07-26
test: add v0.33.0 multi-GiB release evidence

duck8823:
## Motivation

v0.33.0 needs reproducible release evidence that metadata discovery, bounded inspection, indexed search, migration/backfill, and aggregate MCP responses stay bounded on a multi-GiB SQLite store without exposing event content or local identities.

## What changed

- added a maintainer-only `repo-tooling release` runner with a 12 GiB capacity preflight, owner-only scratch lifecycle, capped in-memory child output, strict metrics-only schema, and blocked-artifact validation;
- converted the existing opt-in direct metadata benchmark to content-free SQLite extents and added a numeric Phase-A marker;
- added a copied pre-31 store benchmark that measures migrations 31/32, integrity, foreign keys, resumable FTS backfill, and separate 25-run metadata/bounded list/context/search series for incomplete and complete FTS states;
- added an MCP aggregate-budget scenario for 100 multibyte canonical-block events and continuation traversal;
- recorded six-host MCP manifest parity and bilingual release documentation;
- committed the metrics-only evidence artifact and the Structure-Behavior Design note.

## Release evidence

**Status: BLOCKED.**

- Phase A: 2,418,733,056 managed bytes, 2,147,483,648 stored-body bytes, 8 events, 0 missing metadata, 25 runs; p95 **605.756875 ms**, above the fixed **250 ms** release target.
- Phase B: source remained byte-identical; integrity passed; foreign-key violations were 0; the first FTS batch stopped incomplete at 128/130 documents and the resumed batch completed at 130/130.
- Phase C: all eight 25-run projection/FTS series completed; metadata projections returned 0 body bytes. Their latency values are observations only because #1558 does not define Phase-C latency gates.
- Phase D: all 100 events appeared exactly once over three pages; each page stayed within 100 items and 65,536 encoded body/block bytes; multibyte text and canonical blocks were observed.
- Phase E: all six packaged hosts passed the shared MCP command verifier.

Do not tag v0.33.0 from this evidence. A separate performance fix and a full rerun are required before release readiness can pass.

## Privacy boundary

The committed artifact contains fixed enums, booleans, counts, byte sizes, and durations only. It has no bodies, prompts, responses, event/session identities, workspaces, local paths, cursors, continuations, raw child output, or digests. The private scratch directory was removed before the artifact was written.

## Dependency

This branch is intentionally based on the unmerged memory-hygiene implementation in PR #1567 at `bd798569`. Keep that dependency until #1567 is merged, then retarget/rebase without dropping its commits and rerun the evidence workflow if the dependency head changes.

## Validation

- `go build ./...`
- `go vet ./...`
- `go test ./...`
- `go tool golangci-lint run --timeout=5m` (`0 issues`)
- `go run ./cmd/repo-tooling docs verify-i18n`
- `go run ./cmd/repo-tooling docs verify-host-coverage`
- `go run ./cmd/repo-tooling integrations verify`
- `go run ./cmd/repo-tooling release run-v0.33.0-evidence --output docs/release/v0.33.0-multigib-evidence.json` (expected nonzero exit: `phase_a_failed`; complete sanitized blocker written)
- `go run ./cmd/repo-tooling release verify-body-free-evidence --input docs/release/v0.33.0-multigib-evidence.json --allow-blocked`
- readiness verification without `--allow-blocked` rejected the blocked artifact as expected
- `git diff --check`
- forbidden-key scan of the committed JSON passed
- scratch-residue scan passed

Closes #1558

### DIFF
--- a/.ai/spec/1558.md
+++ b/.ai/spec/1558.md
@@ -131,3 +131,40 @@
 Proceed with repo-tooling-owned orchestration, test-only SQLite/MCP probes, and
 strict metrics-only serialization. Do not add a user-facing command, read a
 live store, or assign Phase-C latency pass/fail targets.
+
+## Phase-A blocker diagnosis checkpoint
+
+The accepted-host run proved that the Phase-A failure is not setup time leaking
+into the measurement. Fixture creation, migration, checkpointing, extent
+verification, and query-plan inspection all occur before the 25 manual timers.
+Each timer covers only `QueryContext`, row scanning, and `Rows.Close`.
+
+The production query plan uses
+`idx_events_workspace_created_at_norm_id_desc`, seeks both timestamp bounds,
+and creates no temporary order-by tree. It is not a covering plan. SQLite's
+`events` record stores `body` before `created_at`, client/workspace metadata,
+and the body-extent metadata. SQLite's documented table-lookup and linked
+overflow-page mechanics, combined with the controlled result below, support
+the inference that restoring those selected columns follows each 256 MiB
+body's overflow chain even though `body` is absent from the SELECT list.
+
+A local diagnostic created one temporary covering workspace index on the same
+2 GiB fixture outside the measurement timer. The identical 25-run query
+measured p95 0.088750 ms. Production-schema runs measured 605.756875 ms and,
+under heavier host I/O pressure, 4,159.414709 ms. This diagnostic is not release
+evidence because the temporary index is not part of the production schema. It
+isolates the non-covering table lookup as the blocker while showing that host
+I/O conditions affect the magnitude.
+
+### Options considered
+
+| Option | Decision | Reason |
+| --- | --- | --- |
+| Change the fixture, timer, or 250 ms target | reject | The fixture and timing match the accepted #1553/#1558 contract; weakening them would hide the physical body read |
+| Rebuild the five ordered indexes as covering indexes | follow-up candidate | It preserves query membership and ordering and fixes every documented scope, but rebuilding five wider indexes on existing stores has material disk, write-lock, and migration-duration impact |
+| Add one metadata-cover index plus self-join query variants | follow-up alternative | It reduces index duplication, but changes every list/context SQL boundary and needs explicit compatibility tests for pre-FTS fixtures |
+| Add a mirrored metadata table | reject for the release fix | It needs backfill and ongoing triggers and creates a larger consistency boundary |
+
+No runtime migration belongs in this evidence PR without an independently
+approved migration budget and rollback plan. The release remains blocked and
+the proposed runtime optimization must use a separate Issue/PR.
--- a/.ai/spec/1558.md
+++ b/.ai/spec/1558.md
@@ -1,0 +1,133 @@
+# #1558 Multi-GiB release evidence
+
+## Structure-Behavior Design Note
+
+### Requirement summary
+
+- **Purpose:** Produce release evidence that metadata discovery, bounded event
+  inspection, indexed search, and aggregate MCP responses remain bounded on a
+  multi-GiB-class SQLite store without publishing event content or local
+  identities.
+- **Dependency:** The evidence branch is based on the unmerged #1556/#1567
+  durable-memory-hygiene branch (`bd798569`). The event projection, FTS, and
+  aggregate response dependencies from #1551 and #1552 are already present in
+  that base. This PR must not be rebased directly onto `main` until #1567 is
+  merged or its commits are otherwise retained.
+- **Phase A:** Run the existing opt-in direct metadata benchmark with eight
+  synthetic 256 MiB extents. Both SQLite-managed bytes and stored-body metadata
+  must reach 2 GiB, all eight rows must have body metadata, 25 measurements are
+  required, and p95 must remain below 250 ms.
+- **Phase B:** Create a private synthetic pre-migration store, copy it, apply
+  migrations 31 and 32 on the copy, and record only migration duration,
+  integrity/foreign-key results, scratch extent, FTS progress, and whether the
+  source stayed byte-identical. The first 128-document backfill must remain
+  observably incomplete before a resumed batch completes it.
+- **Phase C:** Record 25-run p95 observations for metadata and bounded
+  list/context/search projections. Search is measured separately while FTS
+  backfill is incomplete and complete. These surface observations have no
+  release pass/fail latency target until the Issue fixes one; only Phase A's
+  250 ms p95 is a gate.
+- **Phase D:** Exercise the MCP aggregate response with a requested limit of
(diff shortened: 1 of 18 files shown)


duck8823:
**Claude (Opus 5) review of `9120808d`**

The artifact schema, the forbidden-key scan, the metrics-only discipline, and the decision to commit a *blocked* result rather than tune the target are all right. The evidence document itself is unusually honest — the Phase-A diagnosis separating the covering-index diagnostic from the release measurement is the best part of this PR. Three things to fix before merge.

### 1. The PR description's headline number contradicts the committed artifact

The body says:

> Phase A: … 25 runs; p95 **605.756875 ms**, above the fixed **250 ms** release target.

The artifact says `"p95_ms": 4159.414709` (`docs/release/v0.33.0-multigib-evidence.json:23`), and the document's own Phase-A table agrees (`docs/release/v0.33.0-multigib-evidence.md:69`). 605.756875 ms is the lighter-host-I/O production-schema observation described at `:95-96`, not the canonical measurement.

That is the difference between "2.4× over target" and "16.6× over target" — a materially different read on how far v0.33.0 is from release-ready, in the one paragraph a release handoff will actually read. The document is correct; please fix the body to match it.

### 2. Nothing enforces the gate

`verify-body-free-evidence` exists, but `git grep` finds it only in `cmd/repo-tooling/release_evidence.go` and four documentation files — no workflow, no Makefile target. Releases are a manual `git tag vX.Y.Z` per `CLAUDE.md`, so a committed `status: blocked` artifact does not actually stop anyone from tagging v0.33.0; the PR's closing instruction ("Do not tag v0.33.0 from this evidence") is enforced by memory alone.

A tag-triggered job running `release verify-body-free-evidence --input docs/release/v0.33.0-multigib-evidence.json` without `--allow-blocked` would turn this from a document into the gate the PR describes. That is a small addition and it is what makes the rest of the work load-bearing.

### 3. The base-branch dependency looks unnecessary

Base is `fix/v0.33.0-7-bounded-memory-hygiene`. The only reference to it anywhere in this branch's own files is prose at `.ai/spec/1558.md:12` — no tooling, benchmark, or test in the diff touches memory hygiene. I checked with a repo-wide grep across the changed paths.

The cost is real: this PR cannot merge before #1567; GitHub's diff hides #1567's 3,739 lines from anyone reviewing #1568; and by this PR's own rule ("rerun the evidence workflow if the dependency head changes"), any change to #1567 during review invalidates the committed artifact. I have just filed findings on #1567 that would require changes there. Rebasing onto `main` removes all three problems and costs nothing.

### Non-blocking

**4. `covering_index` is emitted but never validated.** `cmd/repo-tooling/release_evidence.go:70` carries the field; `validateBodyFreeEvidencePhaseA` (`:633-642`) checks `OrderedIndex` only. The artifact records `false`, and the document's diagnosis (`:83-100`) identifies the non-covering table lookup as *the* blocker — which is what #1570 sets out to fix. Once #1570 lands, gating `CoveringIndex == true` would stop a future "pass" artifact from silently re-measuring the un-fixed shape.

**5. Phase C is noise-dominated; label it so nobody quotes it later.** `search|bounded|complete` (214.79 ms) is *faster* than `search|metadata|complete` (294.95 ms), and the ordering inverts at `incomplete` (607.97 vs 395.93). The document is already clear that #1558 defines no Phase-C gate; one sentence saying the series is not powered to compare projections would prevent it being cited as evidence that the metadata path is or isn't faster.

**6. Phase A's fixture is 8 events holding 2 GiB** — the few-huge-rows shape. That is the right fixture for the overflow-chain hypothesis and the document says so. It says nothing about the many-small-rows shape, which is the other way a store reaches multiple GiB. Worth stating explicitly under "what this evidence does not cover".

**7. `scratch_cleanup_failed` overwrites an earlier reason.** `:339` assigns unconditionally while every other reason goes through `setBlockReason` (`:295-299`), so a phase failure plus a cleanup failure reports only the latter. Separately, when cleanup does fail the artifact deliberately records no path — leaving a maintainer with several GiB somewhere under `os.TempDir()` and no pointer to it. Echoing back the operator-supplied `--scratch-root` stays metrics-only in spirit while being actionable.

**8. `cappedEvidenceBuffer` silently drops output past 2 MiB** (`:399-415`). If a `-v` benchmark ever exceeds that, the marker line is lost or truncated mid-JSON and the run reports `phase_*_evidence_missing` rather than the real result. It fails closed, so this is a diagnosability note, not a correctness one.

**9. The tooling is version-welded.** `run-v0.33.0-evidence` hardcodes the version in the command name and the validators pin fixture counts (`Events != 8`, `!= 130`, `InitialFTSDocuments != 128`). It will need renaming or deleting for v0.34.0; a `--target-version` flag plus a fixture descriptor would age better.

CI is green (Test / Lint / Docs / Integrations) on `9120808d`, and the PR is correctly still a draft.

duck8823:
**Claude (Opus 5) re-review — no change on this branch, but its base moved twice**

`9120808d` is unchanged, so my earlier three items stand as written. Two things happened around it that matter here.

### The dependency head moved

`fix/v0.33.0-7-bounded-memory-hygiene` advanced `bd798569` → `24acafc1` → `3560a9e6`, and `3560a9e6` is a substantial change: the hygiene cursor is now AES-GCM encrypted, a new `revision_changed` stop reason and `consistency` field were added to the MCP output, and `tool_registry.golden.json` / `tool_schema_budget.golden.json` were regenerated. By this PR's own rule — "rerun the evidence workflow if the dependency head changes" — the committed artifact is already stale against the branch it is based on, twice over. That base branch also currently has a red Lint check.

This is the concrete cost I was pointing at in item 3 of my first review: nothing in the diff needs #1567, so the artifact is being invalidated by changes it does not depend on. Rebasing onto `main` stops that from recurring.

### The Phase-A blocker looks resolved by #1570

#1570's `5738de98` records the same Phase-A probe at p95 **0.07242 ms** against the 250 ms gate, versus the **4,159.414709 ms** in `docs/release/v0.33.0-multigib-evidence.json:23`. That is consistent with this PR's own diagnosis (`docs/release/v0.33.0-multigib-evidence.md:83-100`): the non-covering table lookup following the body's overflow chain was the mechanism, and #1570 is what removes it.

So the practical sequence is: land #1570, rebase this onto `main`, rerun `release run-v0.33.0-evidence`, and the artifact should flip from `blocked` to `pass` — at which point `covering_index` becomes `true` and my item 4 (gate it in `validateBodyFreeEvidencePhaseA`) becomes worth doing in the same pass rather than later.

Items 1 (the PR body quoting 605.76 ms where the artifact says 4,159.41 ms) and 2 (nothing in CI enforces `verify-body-free-evidence`) are unaffected by any of this and still apply. Item 1 in particular should be fixed even though the number is about to be replaced, because the rerun will produce a new headline figure and the body should be trustworthy when it does.

PR #1570, opened 2026-07-26
v0.33.0-10: add persistent covering metadata access

### LINKED ISSUE #1569: v0.33.0-10: avoid event-body overflow reads in metadata queries
## Motivation

The multi-GiB release probe proves that metadata-only range queries still perform table lookups through large event-body overflow chains. The canonical Phase-A p95 is 4159.414709 ms against the 250 ms release target, despite ordered index seeks and no temporary sort.

A diagnostic temporary covering index lowers p95 to 0.088750 ms. The release target requires a durable schema/query solution that avoids body overflow reads while preserving metadata query semantics.

## Acceptance criteria

- Add a migration-safe persistent access path that covers metadata list/context projections without selecting event bodies.
- Verify query plans use the access path and avoid temporary sorts.
- Re-run the sanitized multi-GiB Phase-A evidence and meet p95 < 250 ms.
- Measure migration time, write amplification, scratch-space needs, and rollback compatibility on a copied store.
- Do not publish raw event bodies, identifiers, paths, or query values in evidence.

duck8823:
## Motivation

Metadata list and context SQL used ordered event indexes but still had to open
the body-bearing event record for non-key metadata. Large bodies could
therefore make a body-free query traverse event overflow pages.

## Summary

- add migration 34 with a persistent payload-free event metadata projection
- backfill existing events and optional command-audit metadata
- preserve the two historical source-hook fallbacks as fixed classifications
- maintain projection rows transactionally with event and audit triggers
- route list, context, latest timestamp/kind, and metadata hydration through
  the projection while preserving filters, ordering, offsets, and keyset pages
- add copied-store, trigger-maintenance, plan, lock, rollback, and operational
  benchmark coverage

## Migration and rollback

Migration 34 is additive. Older binaries ignore its table and indexes, while
the persisted triggers maintain projection rows for their writes. A code
rollback therefore leaves authoritative events readable and writable without
removing the projection.

The migration backfill and index creation use one transaction. A competing
write lock returned a busy outcome after 1,012.467 ms with zero partial
projection objects; retry succeeded after lock release.

## Dependencies

This branch is based on current `main`. The durable-memory branch in pull
request 1567 reserves migration 33, so this change uses migration 34 and has no
runtime dependency on that branch. The release-evidence branch in pull request
1568 must rerun its evidence after this change; its unrelated tooling is not
copied here.

## Validation

- `go build ./...`
- `go vet ./...`
- `go test ./... -count=1` — 4,055 tests passed in 23 packages
- `go tool golangci-lint run` — 0 issues
- `git diff --check`

Copied-store migration benchmark on Go 1.26.3, macOS 26.5, darwin/arm64,
Apple M4:

- 8 event/projection rows
- migration: 309.6 ms
- source/copy main file: 302,714,880 bytes before and after
- main-file growth: 0 bytes
- peak scratch: 605,528,480 bytes
- post-checkpoint scratch: 605,429,760 bytes
- integrity: passed
- foreign-key violations: 0
- source byte-identical: true

Multi-GiB Phase A:

- managed bytes: 2,418,753,536
- stored body bytes: 2,147,483,648
- event/projection rows: 8/8
- measurements: 25
- projection-only plan: true
- missing body metadata: 0
- returned body bytes: 0
- p95: 0.2921 ms (target: below 250 ms)

All benchmark fixtures were synthetic and temporary. Only metrics and fixed
booleans are included here.

## Residual risk

The copied-store migration benchmark is 256 MiB-class; Phase A proves runtime
behavior on a multi-GiB store but does not claim a multi-GiB migration-time
bound. Large-store rollout still requires free-space and writer-quiescing
guidance.

Closes #1569

### DIFF
--- a/.ai/spec/1569.md
+++ b/.ai/spec/1569.md
@@ -1,0 +1,171 @@
+# #1569 Persistent covering metadata access
+
+## Structure-Behavior Design Note
+
+### Requirement summary
+
+- **Purpose:** Make event metadata list and context reads independent of the
+  `events` table payload so a large event body cannot make a body-free read
+  traverse its overflow chain.
+- **Current behavior:** Migration 31 supplies ordered indexes, but the metadata
+  SQL still selects non-key columns from `events`. Those indexes are therefore
+  not covering, and SQLite must visit the body-bearing table record.
+- **Expected behavior:** An additive, persistent metadata projection is
+  backfilled from existing events and command-audit metadata, maintained in the
+  same transaction by triggers, and used by every list/context metadata query.
+  Result membership, ordering, filters, body-extent facts, command-failure
+  facts, offset paging, and keyset paging remain unchanged.
+- **Legacy hook behavior:** The two supported body-prefix fallbacks are reduced
+  to a fixed classification in the projection. Runtime metadata reads compare
+  that classification and never evaluate the event body.
+- **Migration evidence:** A copied pre-migration synthetic store must upgrade
+  without changing the source copy, pass integrity and foreign-key checks,
+  preserve row semantics, and expose sanitized duration, database growth,
+  temporary write extent, and lock observations.
+- **Performance gate:** The sanitized Phase-A fixture contains at least 2 GiB
+  of SQLite-managed and stored body bytes. Twenty-five direct metadata range
+  observations must have p95 below 250 ms and the query plan must use the
+  projection without opening the body-bearing event table.
+- **Privacy:** Durable design, commit, PR, and benchmark summaries contain
+  metrics and fixed booleans only. They do not contain bodies, event/session
+  identities, workspaces, local paths, search text, SQL text, continuations, or
+  raw command output.
+- **Out of scope:** Moving bounded/full event reads to the projection, changing
+  event or command-audit retention, changing FTS membership, removing legacy
+  indexes, or making migrations run online without a write lock.
+
+### Dependency decision
+
+- This branch starts at `origin/main` commit `a37ad721`; PR #1567 and PR #1568
+  are not in that base.
+- PR #1567 reserves migration 33 for durable-memory revision state. This issue
+  therefore uses migration 34. The migration runner supports applying a
+  previously absent lower version, so #1567 and this issue have no runtime code
+  dependency and may merge in either order.
+- PR #1568 is release-evidence work based on #1567. It diagnoses the
+  non-covering Phase-A plan that this issue fixes. Its evidence must be rebased
+  or rerun after this change; this PR does not copy #1568 tooling or memory
+  hygiene changes.
+
+### Conceptual model
+
+| Concept | State | Behavior | Constraint / invariant |
+| --- | --- | --- | --- |
+| Authoritative event | Existing `events` row including body | Owns the durable event and every full/bounded body read | Projection never becomes an event write API |
+| Metadata projection | One row per event with metadata, normalized order key, legacy-hook class, and optional command-audit facts | Serves list/context metadata reads without opening the authoritative event table | No body, command text, input, or output column exists; projected row count and identity match events |
+| Legacy hook class | `subagent_stop`, `pre_compact`, or NULL | Preserves the two historical body-prefix fallbacks | It is derived transactionally on backfill and relevant body/kind/source-hook writes; arbitrary body text is never copied |
+| Projection maintenance | Event and command-audit triggers | Inserts, replaces, clears, or deletes the matching projection row in the writer transaction | A committed authoritative mutation and projection state cannot diverge |
+| Projection ordering path | General, workspace, session, workspace-session, and source-hook indexes | Provides the existing `(created_at_norm DESC, id DESC)` order and keyset seek | List/context plans do not open `events` or create an order-by temporary tree |
+| Migration observation | Counts, bytes, elapsed durations, lock classifications, integrity booleans | Describes copied-store upgrade cost without content | Metrics-only; the original copied source remains byte-identical |
+
+### Responsibility assignment
+
+| Responsibility | Owner | Reason to change | Not owner / reason |
+| --- | --- | --- | --- |
+| Projection schema, backfill, indexes, and maintenance | migration 34 | These are persistent SQLite compatibility and writer-order concerns | Application types must not know table or trigger names |
+| Metadata list/context SQL | SQLite metadata datasource SQL | It owns read plans and storage projection hydration | Use case retains criteria validation and orchestration only |
+| Result semantics | Existing `EventMetadata` application type and metadata use case | It is the consumer-visible body-free contract | Migration does not introduce a second DTO |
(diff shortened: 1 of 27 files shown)


duck8823:
**Claude (Opus 5) review of `2dbb5b13` — one blocking defect**

The projection is the right shape and the read-path change is a real improvement, but the update trigger reintroduces a body-derived computation that `events` deliberately does not do, and it diverges on the retention path.

### Blocking: the projection contradicts `events.body_stored_bytes` after a retention prune

`schema/sqlite/migrations/000034_create_event_metadata_projection.sql:261-265` (VALUES) and `:294-298` (ON CONFLICT) recompute the stored size from the new body whenever the body changed:

```sql
CASE WHEN NEW.body IS NOT OLD.body
     THEN length(CAST(NEW.body AS BLOB))
     ELSE NEW.body_stored_bytes END
```

`events` does not. `000026_add_raw_body_retention.sql`'s `events_body_metadata_after_body_update` is guarded by `WHEN NEW.body_availability = 'available'`, so when `infrastructure/sqlite/raw_body_retention_datasource.go:237` replaces the body with the 37-byte `[traceary:body-unavailable:retention]` marker and flips availability, `events.body_stored_bytes` intentionally keeps the original size. The projection has no such guard and overwrites it with 37.

Reproduced with stock `sqlite3`, all migrations from this branch applied in order, then the exact `UPDATE` from `raw_body_retention_datasource.go:237`:

```
after insert  ev=5    pj=5
after prune   ev=5    availability=unavailable_retention
after prune   pj=37
```

This is user-visible as of this PR, because `infrastructure/sqlite/sql/select_recent_event_metadata.sql:3,7` now reads `body_stored_bytes` from the projection. `list_events` with `projection=metadata` will report 37 for every pruned event — `presentation/mcpserver/output.go:84` documents that field as "stored event body byte count" — while the `events`-backed paths keep reporting the original size. `body_stored_bytes` is how an operator sees what a prune reclaimed, so after this change the metadata projection reports the marker's size instead of the size that was freed.

Suggested fix — mirror migration 26's guard in both places:

```sql
CASE WHEN NEW.body IS NOT OLD.body AND NEW.body_availability = 'available'
     THEN length(CAST(NEW.body AS BLOB))
     ELSE NEW.body_stored_bytes END
```

On the available→available path the recompute branch still runs (verified separately: a body edit from 5 to 13 bytes gives `ev=13 / pj=13` today, and that path keeps taking the recompute branch under the added condition). I verified the divergence, not the patch.

**Why the tests didn't catch it:** `infrastructure/sqlite/event_metadata_projection_migration_test.go` seeds retention rows with `body_availability='unavailable_retention'` already set at INSERT time. The `available` → `unavailable_retention` transition — the only path that produces the divergence — is never executed. A regression test should run the real prune statement rather than construct its end state.

### Non-blocking

**1. Backfill cost on a multi-GiB store.** The `INSERT ... SELECT ... FROM events e LEFT JOIN command_audits` at `:38-88` is a single non-resumable statement, and the `legacy_source_hook` CASE reads `e.body`. #1568's Phase B measured 246,959 ms for migrations 31/32 alone on a 2.4 GiB store, and its own Phase-A diagnosis explains the mechanism: `events.body` precedes the other columns in the record, so reading them follows the overflow chain even when the projection excludes `body`. So a first command after upgrading to v0.33.0 will block for minutes on a large store, with no progress output, and an interrupt rolls the whole transaction back and starts over. Migration 32's FTS backfill is batched and resumable (#1568 Phase B: 128/130, then 130/130); this one is not. At minimum the one-time cost belongs in the release notes.

**2. No drift detection anywhere.** `grep -rl event_metadata_projection` returns only the migration, the queries, and tests — `doctor` gains no check. The defect above is exactly the class a count/aggregate comparison between `events` and the projection would surface, and the projection is now the sole source for the metadata read path.

**3. `body_availability` is not carried into the projection**, so the read path cannot distinguish a small body from a pruned one without going back to `events`. Combined with (1), that's an argument for storing it.

**4. Migration numbering.** This branch adds 34 while #1567 adds 33 and neither contains the other. `infrastructure/sqlite/migrate.go:57` tracks applied versions as a set, so a gap is safe and out-of-order application still works — noting it only so nobody "fixes" the numbering during a rebase.

**5. Good change:** `select_recent_event_metadata_by_source_hook_with_legacy.sql` no longer evaluates `e.body LIKE '[phase:subagent]%'` at read time and uses the precomputed `legacy_source_hook` column. That is the right place to pay that cost.

CI is green (Test / Lint / Docs / Integrations) on `2dbb5b13`.

duck8823:
**Claude (Opus 5) re-review of `5738de98` — blocker resolved**

The fix is the right one and lands at the right layer: the projection now mirrors migration 26's rule instead of re-deriving the extent from whatever body happens to be present.

I re-ran the reproduction against a fresh database built from this branch's migrations (stock `sqlite3`, all 34 applied in order), using the exact statements from `raw_body_retention_datasource.go:238` and `:438`:

| transition | `events` | projection |
|---|---:|---:|
| insert (`hello`) | 5 | 5 |
| prune → `unavailable_retention` | 5 | 5 |
| restore to a 42-byte body | 42 | 42 |
| plain available-body edit | 13 | 13 |
| insert already-pruned row | 37 | 37 |

Parity holds on every path, including the two the new test does not cover (a plain body edit, and inserting a row that is already `unavailable_retention` — the latter works because migration 21's insert trigger is unconditional on both sides). Trigger firing order between `events_body_metadata_after_body_update` and the projection trigger is unspecified in SQLite, but both orderings converge here because the restore sets `body` and `body_availability` in one statement (`raw_body_retention_datasource.go:438`), so the guard sees the final availability either way.

`assertStoredBodyBytesParity` asserting both rows against one expected value — rather than asserting the projection alone — is the right shape for this invariant; it fails if either side drifts.

### The four non-blocking items are all addressed

1. **Backfill cost** — `docs/architecture/index-backed-metadata-list.md` now states plainly that it is a one-time non-resumable scan, that legacy-hook classification inspects bodies, that an interrupt restarts migration 34 from the beginning, and that operators should quiesce writers and reserve disk. That is what I was asking for.
2. **Drift detection** — declined with a recorded rationale (an unbounded projection-to-events comparison in the normal `doctor` path is not a minimal safe addition for multi-GiB stores), and the invariant is instead pinned by migration/trigger integration tests. Reasonable trade, and the decision is now discoverable in `.ai/spec/1569.md` rather than implicit.
3. **`body_availability` not projected** — same, with the reasoning that no `EventMetadata` consumer reads it and the trigger already has the authoritative `NEW` value. Agreed, given (1) — one fewer column to keep in sync.
4. **Migration numbering** — noted in the spec.

### One observation, not a request

The updated Phase-A figure in the doc (p95 **0.07242 ms** against the 250 ms gate, versus 4,159.41 ms in #1568's committed artifact) means this PR resolves what #1568 recorded as the sole v0.33.0 release blocker. Worth saying so explicitly in the merge order: #1568's evidence has to be regenerated after this lands, and at that point it should flip from `blocked` to `pass`. I've noted the same on #1568.

CI is green (Test / Lint / Docs / Integrations) on `5738de98`. No remaining objections from me.

Issue #1682, opened 2026-08-07
v0.34.0-42: make the existing dedupe repair runnable at full store size


duck8823:
## What changed: the repair command already exists

`traceary store dedupe content-events` (`presentation/cli/store_dedupe.go`, `infrastructure/sqlite/content_event_dedupe_datasource.go`) already does what this issue was written to build:

- targets exactly `events` where `client='hook'` and `kind IN ('prompt','transcript')`; command audits are never touched
- groups by `(kind, client, agent, session_id, workspace, source_hook, TRIM(body))`
- keeps the earliest row per cluster, moves the rest into `event_content_dedupe_archive`
- dry-run by default, reversible via `--restore <run-id>`

Deletion also propagates correctly: `event_search_documents.event_id` is `REFERENCES events(id) ON DELETE CASCADE`, `foreign_keys(1)` is in the standard DSN, and `event_search_documents_after_delete` removes the FTS row. So removing an event does reclaim its share of the 15.83 GiB index.

**This issue is therefore not "build a repair tool". It is "make the existing one able to run at our size, and then run it."**

## Decision (resolved 2026-08-07)

**Delete the redundant rows** — option (1). Counts become truthful. Historical event totals change, which is correct and must be called out in the release notes. Keeping a 23x inflation to protect the appearance of stability makes every statistic in the 記録 pillar a lie.

Quarantine-then-purge satisfies this: the row leaves `events`, so counts are immediately truthful, and the archive is a bounded rollback window rather than permanent storage.

## Why it cannot run today

Measured against the live store, 2026-08-07:

| | |
|---|---|
| eligible rows | **246,657** |
| eligible stored bytes | **3.31 GiB** (decoded plaintext is larger) |

`DedupeContentEvents` rejects `MaxScanRows > 0` when `Apply` is set (`content_event_dedupe_datasource.go:106`), so an apply run must:

1. `loadDedupeCandidates` every one of those 246,657 rows into a Go slice, each with its **decoded** body (`loadEventPlaintext` per row)
2. hold all of it in one `BeginTx` transaction
3. perform ~200,000 INSERT+DELETE pairs before a single `Commit`

Three consequences:

- **Memory.** Several GiB of live Go heap in one pass. Untested at this scale.
- **Not resumable.** A kill, an OOM, or a laptop sleep rolls the entire run back. The issue's own "interruptible" requirement is unmet by construction.
[...]

PR #1694, opened 2026-08-07
fix: make Kimi transcript recording idempotent per turn

### LINKED ISSUE #1681: v0.34.0-41: make transcript recording idempotent per turn
## Problem

`events.body` is **57.7% exact duplicate** (2.91 GiB of 5.05 GiB, measured 2026-08-07 on the live store), and almost all of it is Kimi:

| agent | rows | distinct | duplicate |
|---|---|---|---|
| **kimi** | 175,614 | 7,445 | **96.6%** (2,468 MiB) |
| antigravity | 709 | 494 | 28.9% |
| codex | 32,253 | 17,382 | 24.2% |
| claude | 1,755 | 1,687 | 1.5% |
| grok | 673 | 663 | 0.0% |

## Diagnosis

The extraction logic is **correct**. `presentation/cli/hook_kimi_transcript.go:181` returns `blocksByTurn[lastTurn]` — the final turn only, not an accumulation.

The measured shape says something else:

- 100% of duplicate groups are within a **single agent name** — not a subagent path-resolution problem
- 96.9% are within a **single session**
- The most-repeated body appears **40,382 times**, 36,552 of them in one session
- Ratio: 175,614 rows / 7,445 distinct turns = **23.6x**

So `Stop` fires roughly 24 times per completed turn, and each firing re-reads and re-records the same unchanged final turn. **Traceary does not guard against re-recording an unchanged turn.** That is the defect — not the extraction.

## Fix direction

`kimiWireTurnID(row.Event.Turn)` already yields a turn identifier. Make transcript recording **idempotent per (session, wire turn)**: if that turn has already been recorded for this session, skip. Content hashing is a fallback for hosts with no turn identifier — prefer the identifier where one exists.

Existing dedup machinery lives alongside `hook_dedup_test.go`; extend it rather than adding a parallel path.

## Also classify (do not blind-fix)

Codex 24.2% and Antigravity 28.9% are far milder and may be legitimate (identical short replies, retries). Measure the same way — same agent? same session? repeat ratio? — before changing either. Record the finding in this issue.

## Done when

- A fixture where `Stop` fires 3 times against an unadvanced `wire.jsonl` records **one** transcript event
- A fixture where the turn advances between firings records one event per turn, with distinct bodies
- Newly ingested data measures < 5% exact-duplicate body bytes
- The Codex / Antigravity classification is written up here

Parent: #1606

duck8823:
## Problem

Kimi's `Stop` hook fires roughly two dozen times per turn, and Traceary recorded a transcript event every time. Measured on the live store: **175,983 kimi rows for 7,644 distinct bodies — a 23.0x repeat ratio**, one body appearing 40,382 times. That is 2.5 GiB of phantom events, and every per-event statistic over Kimi history is inflated by the same factor.

## What was actually wrong

Not the extraction. `readKimiWireTranscriptBlocks` already returns only the last turn's blocks (`blocksByTurn[lastTurn]`), and that is unchanged here.

The gap is that nothing prevented re-recording a turn that had not changed. Traceary already has an exact-redelivery ledger, but it never engaged for Kimi transcripts: `provenHookDeliveryIDFields("kimi")` keys on `tool_use_id`, a field Kimi's `Stop` payload never carries — `Stop` is a turn boundary, not a tool call. So `resolveHookDeliveryNativeID` always returned `""`, `attachHookDelivery` returned early without recording evidence, and every redelivery was inserted as a new row.

## The fix

Idempotency keyed on `(session_id, wire turn ID, content fingerprint)`. A firing is skipped only when **both** the turn ID and a SHA-256 fingerprint of its extracted blocks match the marker.

The fingerprint is not redundant. `Stop` is **not** a completed-turn boundary — it re-fires while a turn is still streaming, so a later firing can carry strictly more content under the same turn ID. Measured over 52,475 consecutive same-session pairs (2026-08-01..07):

| | pairs |
|---|---|
| identical bodies | 49,718 |
| **same turn gained blocks** | **2,053** |
| of those, gained the turn's only `"type":"text"` block | 153 |
| unrelated (a genuinely new turn) | 704 |

2.34 MB of content exists only in the later snapshot. Keying on the turn ID alone would write the marker on the first, incomplete firing and silently discard everything the turn produced afterwards — trading 23x duplication for silent transcript truncation, which is strictly worse than the bug being fixed. (Bodies are `{"blocks":[...]}` JSON, so a grown snapshot is not a byte-prefix of the earlier one; a naive prefix test reports zero growth and is what made this easy to miss.)

State is a per-session marker file under the existing `resolveHookStateDir()`, holding `<turnID>\t<fingerprint>`. A marker that is not exactly two nonempty tab-separated fields — including one written by an earlier turn-ID-only build — decodes as "not recorded" and falls open toward recording rather than risk a false match.

Live duplicates arrive as little as **0.14 s** apart, including effectively concurrently, so check-then-record-then-mark runs inside an exclusive lock. That lock is `github.com/gofrs/flock` (already used by `hook_memory_extract_queue.go`), not the mkdir idiom: the critical section runs a migration check plus an insert against a store that can be tens of GB, inside Kimi's 5 s host timeout, and a mkdir lock's deferred cleanup does not run on SIGKILL. A stale mkdir lock would have had no PID check, no TTL, and no cleanup — every later firing in that session would spin the full budget, time out, and fall open, permanently disabling the guard. flock is released by the kernel on process death.

**The marker is written only on confirmed persistence.** The guard originally treated a nil error from the shared recorder as proof a row had been written. It is not: the recorder fails soft — nil error, nothing persisted — on several conditions unrelated to whether extraction succeeded (unknown client, no extractable content, an unresolved session). Marking on any of those would have burned the marker for a turn that was never stored, and every later redelivery of that unchanged turn would then match the marker and be dropped forever — the exact permanent-loss failure this guard exists to avoid. The recorder now reports `(recorded bool, err error)`, and the guard marks only when `recorded` is true.

**The wire log is read once per firing, not twice.** The guard extracted the turn to compute the fingerprint, then the recorder independently re-extracted it from `wire.jsonl`. Besides doubling a full `bufio` scan of the session wire log inside Kimi's 5 s budget, it meant the fingerprint keyed one read while the persisted body came from a second, later read of a file that can change between the two — precisely the window the CRITICAL above needed. `runHookTranscriptWithBlocks` now accepts the already-extracted blocks; supplying none falls back to the registered extractor, so every other client's path is byte-for-byte unchanged.

Everything fails open. An unresolvable turn, an uncomputable fingerprint, an unavailable lock, or an unwritable marker all fall through to recording unguarded: a hook must never fail the host's turn or silently drop a genuinely new turn over housekeeping state.

## What this deliberately does not do

It does not supersede the earlier, shorter row a growing turn leaves behind — that needs a delete port the hook does not have. Recording each growth snapshot collapses ~24 firings to roughly 2–3 rows: a ~95% reduction with **zero content loss**. Filed as #1697.

Note those growth snapshots are not collapsible by `store dedupe content-events` (#1682) either, since a prefix is not byte-identical to its extension.

## Scope

Kimi only. Measured repeat ratios: **kimi 23.0x**, codex 1.8x, claude 1.1x, antigravity 1.3x, grok 1.0x. Kimi is the only pathological host; the others sit within the range explained by legitimately identical short content, and their behaviour is deliberately unchanged.

## Tests

`presentation/cli/hook_kimi_transcript_idempotency_test.go`:

- three `Stop` firings against an unadvanced `wire.jsonl` → exactly one transcript event
- the turn advances between firings → one event per turn, distinct bodies
- **the same turn gains a block between firings → two events, the second carrying the appended content** (the case the first design lost)
- **eight goroutines firing concurrently on the same turn → exactly one event** (verified under `-race`, repeated 10x)
- **a failed record leaves the marker unwritten → the next firing retries the turn** (pins the invariant that would otherwise silently drop a turn forever)
- **the recorder skips without error → the marker is left unwritten and the next, healthy firing records** (verified by mutation: marking unconditionally on nil error makes this subtest fail)
- **lock acquisition times out under real flock contention → records unguarded, and the marker is absent** (the earlier assertions did not distinguish this from the uncontended happy path)
- marker state unavailable → records unguarded, exit 0
- event store error → exit 0

## Known gap (follow-up, not this PR)

`traceary hook transcript kimi` is unwired in every plugin manifest, and the Kimi spool replay path routes through `replayKimiSpoolRecord` → `runHookKimiStop`, which **is** guarded. #1696 tracks the remaining generic entry point.

## Verification

```
go build ./...                     exit 0
go vet ./...                       exit 0
go test ./...                      exit 0
go test -race -run Kimi ./presentation/cli/   exit 0
go tool golangci-lint run          0 issues.
```

Closes #1681

### DIFF
--- a/presentation/cli/export_test.go
+++ b/presentation/cli/export_test.go
@@ -6,6 +6,8 @@ import (
 	"path/filepath"
 	"strings"
 	"time"
+
+	"github.com/duck8823/traceary/domain/types"
 )
 
 // ResolveDBPath exposes resolveDBPath for tests.
@@ -123,6 +125,22 @@ func AntigravityWorkspaceCwd(payload []byte) string {
 // a conversation/step pair so tests can age or inspect it directly.
 func AntigravityPendingCommandPath(conversationID, stepIdx string) (string, error) {
 	return antigravityPendingCommandPath(conversationID, stepIdx)
+}
+
+// SetResolveHookTranscriptSessionIDFunc overrides the session-ID resolver
+// runHookTranscriptWithBlocks uses. Tests use it to force the "session
+// resolution yielded nothing" fail-soft skip (recorded=false, err=nil) in
+// isolation from any particular caller's own upstream preconditions — e.g.
+// Kimi's idempotency guard (#1681), whose own turn-resolution check would
+// otherwise make that skip unreachable through payload manipulation alone.
+func SetResolveHookTranscriptSessionIDFunc(f func([]byte, string) (types.SessionID, error)) {
+	resolveHookTranscriptSessionIDFunc = f
+}
+
+// ResetResolveHookTranscriptSessionIDFunc restores the default session-ID
+// resolver for runHookTranscriptWithBlocks.
+func ResetResolveHookTranscriptSessionIDFunc() {
+	resolveHookTranscriptSessionIDFunc = resolveHookSessionID
 }
 
 // SetDetectRepoContextFunc replaces the work-context resolver for tests.
--- a/presentation/cli/hook_kimi.go
+++ b/presentation/cli/hook_kimi.go
@@ -191,6 +191,17 @@ func (c *RootCLI) runHookKimiStop(ctx context.Context, input io.Reader, dbPath s
 // firing is skipped only when both the turn ID and the fingerprint of its
 // extracted blocks match the marker.
 //
+// The blocks extracted here for the fingerprint are the SAME blocks passed
+// to runHookTranscriptWithBlocks below — the shared recorder never
+// re-extracts from the wire log on this path. Two independent reads of
+// wire.jsonl (one for the fingerprint, one inside the recorder) used to run
+// per firing; besides doubling the per-firing cost of a bufio scan of the
+// whole session wire log, it meant the fingerprint keyed one read while the
+// persisted row came from a second, later read of a file that can change
+// between the two (rotated, truncated, rewritten). Passing the blocks
+// through closes that window: the fingerprint and the persisted body are
+// now guaranteed to describe the same content (#1681 HIGH finding).
+//
 // This deliberately does not delete or supersede the earlier, shorter row
 // that a growing turn leaves behind — that needs a delete port this hook
 // does not have. Recording each growth snapshot still collapses ~24
@@ -200,7 +211,13 @@ func (c *RootCLI) runHookKimiStop(ctx context.Context, input io.Reader, dbPath s
 // The check-then-record-then-mark sequence runs inside
 // withKimiTranscriptTurnStateLock so two racing firings for the same
 // (turn ID, fingerprint) cannot both observe "not yet recorded" and both
-// write a row.
+// write a row. Marking is additionally gated on the recorder's own
+// recorded=true return value, never inferred from a nil error: the shared
+// recorder fails soft on several unrelated conditions (e.g. an unresolved
+// session), and treating any of those nil-error returns as "a row was
+// persisted" would let a turn that was never actually stored get marked as
+// done — silently discarding every later redelivery of that turn forever
+// (#1681 CRITICAL finding).
 //
 // When the wire turn cannot be resolved (missing index entry, missing wire
 // log, escaped path, ...), the fingerprint cannot be computed, or the turn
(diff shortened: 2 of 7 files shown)


duck8823:
**Codex verifier review** (read-only, full branch diff)

Two HIGH findings. Both checked against the code before acting on them.

### 1. Lock-timeout fall-open reproduces the duplication — **accepted, fixed in f70c32c**

Codex's line reference was wrong (it pointed at `copyKimiHookField`), but the claim held once traced to the real call site.

The acquisition budget `kimiTranscriptTurnLockTimeout` is 1s. `sqliteBusyTimeout` in `infrastructure/sqlite/database.go:403` is also 1000ms, and the insert inside the critical section is bounded by it. So a store under write contention holds the turn lock for about as long as a contending firing is willing to wait — exhausting the wait is not an anomaly, it is what a contended store does. On that path the hook recorded unguarded, i.e. the guard stopped working under exactly the load it exists for.

Losing the race now returns `errKimiTranscriptTurnLockBusy` and the firing yields. Safe here in a way falling open is not: Kimi redelivers Stop for the same turn ~24 times, so a yielding firing is retried ~0.14s later, and if the holder succeeds its marker makes those redeliveries correct no-ops. A lock no firing can take at all (unresolvable state dir, failed mkdir) still falls open — yielding there would drop the turn outright.

Mutation-verified: reverting the branch makes the test fail with `transcript log calls = 1, want 0`.

### 2. Marker not keyed by target DB — **verified, declined**

Real but much narrower than described. Codex says a stale marker "permanently suppresses" recording; the marker holds only the **last** `(turnID, fingerprint)` pair for a session (`kimiTranscriptTurnAlreadyRecorded`, `hook_kimi_transcript.go`), so a `--db-path` switch mid-session costs at most that one turn's final snapshot, and every later turn records normally because its turn ID differs.

Trigger is switching the target DB or recreating it in the middle of a live session. Not filing an issue: keying the marker on DB identity adds state-format surface for a one-turn loss in a scenario that is not a normal operation. Recorded here so the reasoning is findable if it ever does bite.

Local gates on `f70c32c`: `go build ./...`, `go vet ./...`, `go test ./...`, `go tool golangci-lint run` — all green.

Issue #1185, opened 2026-06-18
RFC3339Nano variable-width timestamps break lexical TEXT comparisons in dedup and period-filter SQL


duck8823:
## Problem

`events.created_at` (and `sessions.started_at` / `ended_at`) are stored as **variable-width RFC3339Nano** via `formatTimestamp` (`infrastructure/sqlite/database.go:156`), which trims trailing zeros. Variable-width RFC3339Nano is **not** lexically ordered the same as real time: `"2026-05-31T01:00:00.5Z"` sorts *before* `"2026-05-31T01:00:00Z"` because `'.'` (0x2E) < `'Z'` (0x5A).

Several SQL paths compare these columns as TEXT against a bound (or against each other). Near a fractional-second boundary they can **drop a row that is temporally in range, or include one that is out of range**.

This was discovered while fixing #1167: the new hook content-event dedup guard (`hookContentEventDuplicateExists`) hit exactly this bug and was fixed there by comparing parsed timestamps in Go after a fixed-width whole-second-prefix coarse filter. PR #1184 deliberately scoped the fix to the prompt/transcript guard only (1 issue = 1 branch = 1 PR; command audits are an explicit non-goal of #1167). This issue tracks the remaining sites.

## Affected sites (from Gemini scout review on PR #1184 + follow-up audit)

1. **Command-audit dedup guard** — `hookCommandAuditDuplicateExists` (`infrastructure/sqlite/event_datasource.go`) uses the same `formatTimestamp(created_at ± window)` TEXT range. Same latent miss/over-suppress near fractional-second boundaries. Apply the same coarse-prefix + Go-comparison shape as the content-event guard.
2. **Read-side period filters** — `e.created_at >= ?` / `<= ?` in `search_events.sql`, `list_timeline_blocks.sql`, `select_recent_events*.sql`, etc. A `--from` / `--to` bound at a whole second can exclude/include an event in that same second depending on fractional width.
3. **Session late-event comparisons** — `list_sessions.sql` (`late_ev.created_at > s.ended_at`) and `find_latest_session.sql` (`row_later.created_at > ended_row.ended_at`, `later_ev.created_at > ended.created_at`), added in #1172. Two stored variable-width timestamps with the same whole-second prefix but different fractional widths can compare in the wrong direction, so a late event ~sub-second after the end marker could be missed.

## Possible approaches

[...]

PR #1695, opened 2026-08-07
fix: make the dedupe repair runnable at full store size

### LINKED ISSUE #1682: v0.34.0-42: make the existing dedupe repair runnable at full store size

duck8823:
## Problem

`traceary store dedupe content-events --apply` already existed and was already correct — it just could not be run.

An apply had to load **every** eligible row with its decoded body into memory and archive all of them in a single transaction. On the live store that is **246,657 rows and 3.31 GiB of stored payload**, held in one Go heap and one `BeginTx`. Any interruption — OOM, a kill, a laptop sleep — rolled the entire repair back. And `event_content_dedupe_archive` had no purge, so a successful apply relocated the duplicates rather than reclaiming them.

## Changes

**Identification no longer retains bodies.** It streams the eligible rows, keys each into its identity group by the SHA-256 of the trimmed body, and keeps only the id and timestamp. That is all the plan needs: choosing the canonical row and clustering by proximity are decided from timestamps alone. Payload metadata columns are probed once and selected inline, replacing the two extra queries `loadEventPlaintext` issued per candidate.

**Apply commits in bounded batches** (`--batch-size`, default 1000), re-reading a body only inside the batch that archives it.

**A cluster is never split across two commits.** Proximity clustering measures the gap between *consecutive surviving* rows, so archiving part of a cluster widens the gaps inside it and can split what was one cluster into several, each keeping its own canonical row. A cluster at t=0s, 9s, 18s under the 10 s window is one cluster keeping only t=0; archive t=9 alone and the remaining 0→18 gap exceeds the window, leaving two singletons that no re-run will ever collapse. Batching therefore packs **whole clusters**, and a cluster larger than the batch size becomes its own transaction — the clustering decision outranks the size target.

That is what makes resumption need no checkpoint state: an interrupted run leaves every cluster either fully archived or untouched, its archived rows are gone from `events`, and a re-run reproduces exactly what a clean run would have decided.

**`--list-runs`** reports the quarantine runs still held in the archive, newest first. Because batches commit before the run id is printed, an apply killed after its first commit had quarantined rows under an id the operator never saw — invisible in `events`, and reachable by neither `--restore` nor `--purge`.

**A strict group is splittable; a proximity group is not.** Cluster atomicity is a property of *proximity* grouping alone: its membership is read from the gaps between surviving rows. `--strict` ignores gaps entirely — one plan entry is the whole identity group, and its canonical row is the earliest, which archiving duplicates never removes — so a partially archived strict group re-plans to exactly the same decision. Treating strict groups as atomic would have put an entire identity group into one transaction; on the live store the largest is **36,552 rows**, re-creating the very thing batching exists to prevent. The plan therefore carries the distinction explicitly and only atomic groups resist a batch boundary.

**Retention-emptied rows are excluded.** `store retention apply` replaces a body with one fixed marker string for every row it prunes, regardless of client or kind, so two prompts that were never duplicates of each other hash to the same identity once emptied. The filter probes `body_availability` so stores predating migration 26 behave as before. (Zero rows on the live store are emptied today; the path is live the moment retention runs.)

**Retention-held rows are excluded too.** A row restored from retention keeps its `raw_body_retention_entries` row — restore only sets `restored_at` — and that row holds an `ON DELETE RESTRICT` reference to `events(id)`. Nothing else distinguishes it: body available, body intact. Archiving one raised a raw SQLite foreign-key error that aborted the batch mid-apply. Eligibility now excludes ledger-held rows, probing for the table so stores predating migration 26 behave as before.

**`--list-runs` orders in Go, not in SQL.** `archived_at` is RFC3339Nano, whose text form is not order-preserving — `'.'` precedes `'Z'`, so a whole second sorts before a fractional one inside the same second. Ordering by the parsed instant is what makes "newest first" true for an operator choosing which run to restore. (Same lexical hazard class as #1185.)

**`--purge <run-id>`** ends a run's rollback window and reports the bytes released, with a pointer to `VACUUM` — SQLite returns the pages to its free list, not to the filesystem. Purge is deliberately separate from apply: reversibility is the point of the archive.

**`--client kimi`** is now a named scope. It is the host that needs the repair.

## What did not change

The identity tuple, the 10 s proximity window, canonical-row selection, malformed-timestamp skipping, `--restore`, and the eligibility filter's client/kind predicate (`client='hook'`, `kind IN ('prompt','transcript')` — command audits are never touched). Every pre-existing test passes unmodified.

One correctness subtlety worth flagging for review: the archive stores the **decoded** body, because `RestoreContentEventDedupeRun` re-encodes it as identity. The batch re-reads and decodes rather than copying the raw `body` column, which would corrupt a codec-encoded payload on restore.

## Scope decisions

**Not reusing the payload-rehearsal harness.** The issue originally said to. Having read it, that is the wrong call: it is a 1,046-line WAL-budgeted, lease-held, shadow-table harness for whole-store codec rewrites with a rollback digest. Coupling dedupe to it would drag in leases, schema-SHA guards, and shadow tables that a row-level repair does not need. What dedupe needed turned out to be smaller than a checkpoint table anyway — see cluster atomicity above.

**`MaxScanRows` still cannot be combined with `Apply`.** That is a *sampling* bound (`ORDER BY created_at DESC LIMIT`), not a batch size; applying on a partial view could pick a non-earliest canonical row. Batching is orthogonal and is expressed separately as `BatchSize`.

**Codex and Antigravity are untouched.** Repeat ratios: kimi 23.0x, codex 1.8x, claude 1.1x, antigravity 1.3x, grok 1.0x. Only Kimi is pathological.

**The default proximity window is adequate.** Median gap between consecutive kimi duplicates is 0.14–2.28 s, and 66–98% of consecutive gaps fall inside the existing 10 s window. `--strict` stays opt-in.

## Tests

- `TestPartitionDedupeTargets` — the batch-partitioning decision is a pure function so the cluster-atomicity invariant is pinned directly rather than inferred: clusters pack until the batch fills, an overflowing cluster starts a new batch, a cluster larger than the batch is never split, and an oversized cluster does not drag the next one along.
- `BatchSizeDoesNotChangeOutcome` — batch sizes 1, 2, 1000, and 0 all land on the identical surviving-event set. Batching is a durability and memory decision, not a semantic one.
- `ResumesAfterPartialApply` — a hand-archived cluster stands in for a batch an interrupted run committed; the re-run moves the rest and converges on exactly the clean-run state. The fixture archives a cluster *in full* because that is the only shape an interruption can leave behind.
- `SkipsRetentionEmptiedBodies` — two prompts that were never duplicates of each other, emptied by retention into the same marker text, must not be grouped. Verified to fail without the filter.
- `PlanContentEventDedupe_StrictGroupsAreSplittable` / `_ProximityGroupsAreAtomic` — pins the planning→partitioning link: strict plans mark groups splittable and a `--batch-size` of 2 really does yield two transactions, while proximity plans stay atomic.
- `SkipsRetentionLedgerRows` — a restored, ledger-held row must not reach the plan, and the rest of the apply must still commit. Verified to fail without the filter, with exactly the foreign-key abort described above.
- `SortDedupeRunsNewestFirst` — fractional vs whole seconds, differing UTC offsets, ties falling back to the run id, and an unparsable timestamp still landing deterministically.
- `ListContentEventDedupeRuns` — empty archive, newest-first ordering, per-run row and byte counts, and disappearance after purge.
- `PurgeContentEventDedupeRun` — drops the archive, reports count and bytes, leaves surviving events untouched, and makes a subsequent restore fail.
- CLI: `--client kimi`, `--batch-size` pass-through and default, `--purge` text and JSON output (asserting the VACUUM guidance is present), `--list-runs` text/JSON/empty output, and a table of rejected mode combinations.

## Expected yield and worst-case transaction size

Measured read-only against the live store (`mode=ro`, no repo build involved), reproducing the default 10 s proximity clustering over `agent='kimi'`:

| | |
|---|---|
| removable rows | **154,451** of 175,983 |
| reclaimable body | **2,230.9 MiB** of 2,572.5 MiB |
| largest single proximity cluster | **887 rows** |
| clusters larger than the default batch size | **0** |

The last two answer the obvious objection to cluster-atomic batching: no cluster on this store exceeds the default `--batch-size` of 1000, so nothing becomes an oversized transaction in practice. Even if one did, memory is bounded independently — `archiveDedupeBatch` re-reads and decodes one body at a time rather than holding the batch's bodies resident.

## Not verified here

The before/after numbers against a rehearsal copy of the real store. That belongs to running the repair, not to landing it, and must never be run from a repo build against the live store — the live store is at migration 34 while `main` carries 35–44 unreleased.

## Verification

```
go build ./...            exit 0
go vet ./...              exit 0
go test ./...             exit 0
go tool golangci-lint run 0 issues.
```

Closes #1682


duck8823:
**Codex verifier review** (read-only, incremental diff `7eac4a5..082c0e8`)

One HIGH finding, accepted and fixed in 03a0e87.

### Migration 45 was classified `constant_in_place`

`MigrationConstantInPlace` means "bounded data-independent installation". Migration 45 builds an index over every existing `raw_body_retention_entries` row, so its cost scales with the ledger. `000035_add_session_lookup_projection_index.sql` sets the precedent directly — also a `CREATE INDEX` over a populated table, also classified `data_dependent_offline` — while the `constant_in_place` migrations around it only create empty tables.

Consequence: a store already at version 44 built a plan reporting `Offline=false` and would have skipped the prepared-migration path. A store at 34 was unaffected because 35, 41 and 42 already force the plan offline, which is why no existing test caught it.

Reclassified, and the rationale is now in the migration body so the next reviewer does not have to re-derive it. The SQL body changed, so the catalog digest was recomputed.

Local gates on `03a0e87`: `go build ./...`, `go vet ./...`, `go test ./...`, `go tool golangci-lint run` — all green.

### Follow-ups filed rather than folded in

- #1699 — identification aborts on a single undecodable body
- #1700 — replace the mode flags with subcommands (user-approved shape; deliberately after this PR)
- #1701 — the `content-event-reliability` diagnostic no longer agrees with this repair
- #1702 — a failed apply drops the run id that would recover its committed batches

Issue #1683, opened 2026-08-07
v0.34.0-43: record orphan ranges and fall back to mechanical summaries


duck8823:
## Problem

```
stop(folded) -- turns -- stop(under threshold) -- compact
                        |___ never folded ___|
```

No data is lost — hooks recorded every turn. What is lost is the cheap opportunity to fold, because the material left the agent's context. And an unfolded range **blocks body discard**, so it keeps disk forever.

By construction the range is smaller than the threshold (64 KiB).

## Scope

1. New `session_orphan_ranges`: on a compact marker (available on 5 of 6 hosts; Antigravity has no compact hook and relies on the threshold alone), record the unfolded range
2. **At gc** — if still unfolded, produce a mechanical summary with no LLM: when, which event kinds, how many, which commands. Set `degraded=1`. "Why" is lost; the target is under 64 KiB

This removes the deadlock where an unfolded tail can never be discarded.

## Removed from scope 2026-08-08: fold on access

An earlier revision had a second path — fold the tail when the session is opened. Dropped.

Writing a summary needs an LLM, and Traceary never calls one, so a fold can only be *requested* from a path whose caller is known to be an LLM. `traceary show` may be run by a human, so it cannot fold anything. That leaves the stop hook, which #1674 already owns.

And the range is by definition under the 64 KiB threshold, so the next stop that crosses the threshold folds it anyway. A second request path would double the surface for material the first path already reaches — and would make a read path write.

## Done when

- A compact marker after an unfolded range creates an orphan row
- gc on a stale orphan produces a `degraded=1` refinement and permits discard
- No read path writes a refinement

Spec: [`.ai/spec/1606.md`](../blob/main/.ai/spec/1606.md) (landed by #1671)

Parent: #1606

duck8823:
**Design correction from the #1671 review (2026-08-08).**

This issue's problem statement says the orphan range is "by construction smaller than the threshold (64 KiB)". That is not an invariant — it is the common case. Two paths exceed it:

- **An ignored fold request.** #1674 ships with no repeat-request suppression, so pressure stays above the threshold until the agent complies. A compact arriving first leaves an orphan range larger than the threshold.
- **A single oversized turn.** Pressure is evaluated only at stop, so one turn that records more than the threshold in one go crosses it with no intervening evaluation point.

[...]

Issue #1674, opened 2026-08-07
v0.34.0-34: trigger consolidation from the stop hook


duck8823:
## Supersedes the previous design

The earlier plan was to request refinement at `SessionStart`. That is wrong: handing the agent raw material at wake time puts that material in the fresh session's context for its whole lifetime — the exact pollution this sprint exists to avoid.

Consolidation must happen **while the material is already in the agent's context**. Verified against primary sources, that means the **stop hook only**:

| host | stop can request work | compact can |
|---|---|---|
| Claude | exit 2 -> "continues the conversation" | **no** — PreCompact has no additionalContext; PostCompact has "No decision control" |
| Codex | `{"decision":"block","reason":...}` -> continuation prompt | **no** |
| Kimi | exit 2 -> "a message can be appended to let the model continue" | **no** — "return values are completely ignored" |

## Scope

At every stop (2,811/day measured), compute pressure from the DB alone — **no LLM, one row update**:

```
pressure = unrefined body bytes for this session
due      = pressure >= threshold
```

When due, exit 2 / `decision:block` with a reason naming the session. The agent already holds the material, so the marginal cost is output tokens only.

**Threshold = 64 KiB.** The threshold bounds the *orphan range* — the material that gets compacted before being folded — and therefore the cost of recovering it later (~16k tokens). It is not a call-count knob.

## Reason payload

The reason is the only channel that reaches the agent, so it carries everything the agent needs to write a *merged* refinement (#1673) rather than a fresh one:

- the session id
- the previous summary, if one exists
- the `covers_to` of that summary

This is why #1673 needs no separate read command: the agent is handed the previous summary at the moment it is asked to fold.

`main.go` already routes a `cliExitCoder` error to `os.Exit(n)` after writing it to stderr, which is exactly the channel the three hosts above read back. No new mechanism.

## Safety

- Any failure in the check exits 0. A consolidation decision must never stop the user's work.
- The threshold must be configurable, including a value that disables the trigger entirely.

## Not in scope — decided 2026-08-08

**No suppression of repeat requests.** If the agent ignores the request, pressure stays above the threshold and the next stop asks again.

[...]

Issue #1673, opened 2026-08-07
v0.34.0-33: add session refinements with re-summarisation


duck8823:
## Scope

The materialised L2 that the whole design depends on. Today session summaries exist for **4 of 26,728** sessions.

New table `session_refinements`:

| column | meaning |
|---|---|
| `session_id` | |
| `generation` | monotonic; a later refinement supersedes an earlier one |
| `covers_from` / `covers_to` | the event range this summary accounts for |
| `summary` | the text |
| `keywords` | for the session-tier index |
| `produced_by` | which agent wrote it |
| `produced_at` | |
| `degraded` | 1 when produced mechanically (no LLM) |

Write port: `traceary session refine <session-id> --summary <text> [--keywords ...] --covers-to <event-id>`.

## Who merges — corrected 2026-08-08

An earlier revision of this issue said the write port "takes the previous summary plus new material and produces one merged summary". That is wrong, and it contradicts the responsibility table in `.ai/spec/1606.md`: *what to write* is owned by the agent, and Traceary never calls an LLM.

The shape is:

```
hook records -> stop asks the agent to fold -> the agent writes the merged summary
             -> the agent hands it to `traceary session refine`
```

The port stores what it is given and owns generation and coverage. It does not compose text.

**Re-summarisation, not last-write-wins.** A session compacts up to **116 times** (measured). First-write-wins (today's `SetSummaryIfEmpty`) keeps only the opening; last-write-wins loses it. Since the agent composes the merge, it must be handed the previous summary before it writes.

**How the previous summary reaches the agent.** Through the #1674 stop-hook reason, which is already the one channel the host feeds back to the model. No separate read command: the agent receives the previous summary and its `covers_to` at the moment it is asked to fold.

**Surface.** CLI only. The MCP surface was cut from 24 tools to 8 in v0.10 and is not being grown; if this is ever wanted over MCP it belongs as an action of the existing `manage_session` tool, not as a ninth tool.

## Done when

- Writing twice for overlapping ranges yields one summary at `generation=2` whose `covers` spans both
- Idempotent: refining the same range twice does not create a second row
- Retention never deletes a refinement
- The stop-hook reason carries the previous summary and its `covers_to` (verified together with #1674)

Spec: [`.ai/spec/1606.md`](../blob/main/.ai/spec/1606.md) (landed by #1671)

Parent: #1606

Issue #1621, opened 2026-08-01
v0.34.0-11: prepare and verify the release candidate


duck8823:
## Objective

Prepare and verify the v0.34.0 release candidate without tagging or publishing it.

Parent: #1606

## Note on the rewrite (2026-08-11)

The original acceptance criteria described the superseded hot/cold segmented-history design — "Hot retention", "immutable segments", "partial historical search", "covered-row deletion", "abandoned migrations 44-48". None of that machinery exists in v0.34; the strict segment migration machinery was removed in aa20e167. The criteria below are rewritten against what actually shipped: per-row zstd compression of event bodies and command audit text, content dedupe, session refinements folding old events into summaries, a bounded search projection, and `store compact`.

Criteria informed by the #1620 gate measurement and a gpt-5.6-sol review of it. They are stated as observable properties, not as patches.

## Blocking status

#1620 is measured. Three rows miss, and two of them are release-blocking as of 2026-08-11:

- **index family** — #1793 / #1794 / #1796. The projection cannot be built to completion on a real store, and its byte ceiling does not bound the build.
- **monthly growth** — #1797. Body discard is transcript-only; the undiscardable sources alone are 0.79 GiB/month against a 0.5 GiB gate.

This issue cannot complete until those are resolved or the corresponding release claims are formally withdrawn.

**Third finding, 2026-08-12: the recent full-text tier is write-only (#1842).** Found while cutting the release, not by the gate. The tier is built, bounded, evicted, reclaimed and measured — and read by nothing. `8cae0ab0` (#1718) removed the last reader when it replaced FTS candidate selection with the newest-first decode walk, and no surface noticed. Verified three ways: no production reader in Go or embedded SQL; the deleted `buildProjectionFTSEventIDsQuery` in `8cae0ab0`'s diff; and empirically — emptying `search_projection_recent_documents` on a store with a complete generation changes no search result. Consequences: `--index-family-bytes` buys retention, not searchable reach; the 9.0 GB `search_projection_recent_fts_data` in the #1620 measurement is the largest object in the file and returns nothing. The claims it falsified are corrected in #1841 (merged); its fate — delete or re-wire search to read it — is #1842, deferred to v0.35.

[...]

PR #1703, opened 2026-08-07
docs: land the two-pillar spec and correct the three-layer model

### LINKED ISSUE #1671: v0.34.0-31: land the two-pillar spec and remove the segmented-history docs
## Scope

Docs only.

1. Land `.ai/spec/1606.md` (written; this issue commits it)
2. Remove `docs/architecture/hot-cold-segmented-history.md` and its `.ja.md` pair — the external-storage design is withdrawn
3. Update `README.md` three-layer table: L2 is **materialised by consolidation**, not "recomputed when the next session starts"
4. Fix the contradiction in `README.md:38` ("L3 only grows when the operator says so") against `docs/memory/README.md` (auto-extraction fires on session end). Reality: 33,014 memories, 29,757 created in July
5. Resolve the duplicate use of the phrase "three layers" — `docs/memory/README.md:238` uses it for an unrelated activation stack

## Blocker on step 1: the spec is gitignored

`.gitignore:10` ignores `.ai/` in full, so `git add .ai/spec/1606.md` is a no-op today and every `../blob/main/.ai/spec/1606.md` link in the #1606 sub-issues currently 404s.

A negative pattern under an ignored directory has no effect — git never descends into it — so the entry has to become:

```
.ai/*
!.ai/spec/
```

`.ai/evidence/`, `.ai/handoff/` and the `review-job-*` working directories stay ignored. Only `.ai/spec/` starts being tracked, which also makes the ~90 existing design notes there available to reviewers.

## The two pillars

1. **記録** — capture works; summarise, compress, and evict from the index
2. **記憶** — the same, plus **automatic use of the summary** at wake time

## Done when

- `.ai/spec/1606.md` is in the repo and the sub-issue spec links resolve
- No file describes hot/cold segmented history as the plan
- The three-layer description matches measured behaviour

Parent: #1606

duck8823:
Docs only. First PR of v0.34 wave 1.

## What changed

**The spec is now in the repo.** `.ai/spec/1606.md` has been steering this sprint from an untracked file, so the `../blob/main/.ai/spec/1606.md` link in every #1606 sub-issue resolved to nothing. It is force-added (`git add -f`); `.ai/` stays ignored, so `evidence/`, `handoff/` and the `review-job-*` working directories are unaffected. A comment above the `.gitignore` rule records why the exception exists.

Landing it also meant making it true. Three decisions taken after it was written are folded in:

- the write port stores what the agent hands it and never composes text, so the previous summary and its `covers_to` travel in the stop hook's reason payload (#1673)
- the stop trigger ships without repeat-request suppression (#1674) — no Traceary hook returns a non-zero exit today, so the failure mode is unobserved and the mitigation is dogfooding before the tag
- fold-on-access is withdrawn (#1683) — a fold can only be *requested* from a caller known to be an LLM, and the stop hook already is one

**The withdrawn architecture doc is gone.** `docs/architecture/hot-cold-segmented-history.{md,ja.md}` described an external-storage design that is no longer the plan. Nothing else referenced it.

**The README three-layer description now matches measured behaviour.** Both claims after the first were wrong:

| claim | reality |
|---|---|
| "L2 is recomputed when the next session starts" | recomputing at wake time is exactly what this design rejects — raw material handed to a fresh session pollutes its whole lifetime. L2 is consolidated at stop, while the material is already in context |
| "L3 ... only grows when the operator says so" | hook-driven auto-extraction fires on session end. What keeps L3 small is that those memories land as `status=candidate` and only human review promotes them |

The `Working memory` row also no longer presents the Claude `PreCompact` digest as the mechanism.

**`docs/memory/README.md` stopped using "three layers" for two different things.** The activation stack is a set of paths by which an accepted memory reaches a host, not a layering of context, so it is now "three activation paths".

## Verification

```
make docs/check   # documentation i18n check passed / Antigravity docs status check passed
```

No Go files touched.

## Notes

The README now describes v0.34 consolidation before #1673/#1674 implement it. v0.34 does not tag until the wave 5 gates (#1620, #1621), so no release carries the forward-looking wording.

Implementation delegated to Grok 4.5 (high). Reviewed by Claude Opus 5; two rounds of corrections applied — the private-store memory counts were removed from the public README, a run-on sentence was split, and two literal-translation phrases in `README.ja.md` were rewritten.

Closes #1671


duck8823:
**Independent review** — gpt-5.6-sol (Codex, read-only), full PR content including the newly tracked spec.

Four HIGH findings, no CRITICAL. All four checked against the sources before acting; all four were real and all four are fixed in 4aca92f.

### 1. The orphan range is not bounded by the threshold — the substantive one

The spec asserted in four places that an orphan range is 定義上閾値未満. It is not an invariant, and the reviewer is right that later work leans on it.

Two paths exceed the threshold:

- **An ignored fold request.** #1674 ships with no repeat-request suppression, so pressure stays above the threshold until the agent complies. A compact arriving first leaves an orphan range larger than the threshold. This hole is created directly by that decision.
- **A single oversized turn.** Pressure is evaluated only at stop, so one turn recording more than the threshold in one go crosses it with no intervening evaluation point.

64 KiB is now stated as the design target, and the two consequences are recorded where the work that needs them will find them: the gc mechanical summary (#1683) must accept an unbounded range, and the wake-time injection budget (#1684) must be enforced on its own terms rather than inferred from the threshold. Noted on both issues.

### 2. `body_state` inside an immutable skeleton

The conceptual model called the event skeleton "written once, never changed" while listing `body_state` in it and requiring it to move to `discarded` / `compressed` elsewhere in the same document. Not implementable as written. The invariant is now split: identity, kind, session, workspace, agent, timestamps and `body_bytes` are immutable; `body_state` moves along the defined transitions and nothing else does.

### 3. "なぜ" was promised unconditionally

Two places promised that discarding bodies still leaves いつ・何を・何回・なぜ permanently, while the degraded gc path in the same document says なぜ is lost. Corrected: なぜ survives only for ranges an agent actually folded.

### 4. Both READMEs overstated how L2 is built

"L2 is consolidated at stop" is the normal path, not the only one — unfolded material is reduced to a mechanical summary instead. One clause added per language, no gc terminology imported into the README.

### Verification

`make docs/check` — documentation i18n check passed, Antigravity docs status check passed. No Go files touched by this PR.

Issue #1714, opened 2026-08-08
v0.34.0 follow-up: verify the Antigravity PreInvocation output contract before injecting at wake


duck8823:
Follow-up from #1684 (wake-time injection).

#1684 injects finished summaries through the hook's stdout, which every other host's wrapper passes through verbatim (`preserve_stdout=1` in the plugin wrappers). Antigravity cannot use that channel: its `PreInvocation` stdout is parsed as JSON (`writeAntigravityJSON`, `presentation/cli/hook_antigravity.go:120`), and the only output vocabulary this repository has verified is `decision` (`allow` / `""`) plus "an empty object so Antigravity injects no steps".

The spec's host table (`.ai/spec/1606.md`) records `Antigravity | ● PreInvocation` for wake injection, but neither the spec nor `docs/integrations/antigravity.md` names the field that carries injected context. Implementing it today would mean guessing a field name and shipping a hook that silently injects nothing — worse than not shipping it, because `doctor` would report the hook as installed.

So #1684 leaves Antigravity out and this issue owns it.

## Done when

- The `PreInvocation` output field that carries context into the model's input is confirmed against Antigravity's official hook documentation (or, failing that, against an observed `agy` run), and the finding is written into `docs/integrations/antigravity.md` alongside the existing `decision` vocabulary
- If a field exists: `hook antigravity pre-invocation` emits the same selected summaries #1684 already computes, through the shared wake-injection component, guarded by the same once-per-session marker (`PreInvocation` fires before every model call)
- If no such field exists: the spec's host table is corrected, and `docs/integrations/antigravity.md` states that Antigravity receives no wake injection and why

No new command, subcommand, or MCP tool. This reuses the selection query, the budget, and the marker from #1684.

Parent: #1606

duck8823:
## The field exists: `injectSteps`

Confirmed against the vendor's own hook documentation, shipped with the Antigravity CLI install at `~/.gemini/antigravity-cli/builtin/skills/agy-customizations/docs/hooks.md`. Its `PreInvocation` section is explicit about the purpose — *"Use to inject context or instructions before the model runs"* — and gives the output contract:

```json
{
  "injectSteps": [
    { "ephemeralMessage": "Remember to check for lint errors before proposing changes." }
  ]
}
```

`injectSteps` is an optional array; the supported step types are `{"toolCall": {...}}`, `{"userMessage": "..."}` and `{"ephemeralMessage": "..."}`, the last documented as a transient system message.

[...]

PR #1715, opened 2026-08-08
feat: inject finished session summaries at wake

### LINKED ISSUE #1684: v0.34.0-44: inject the finished summary at wake time
## Problem

This is the missing half of the product. `additionalContext` and `hookSpecificOutput` appear **zero times** in the codebase; MCP `get_context` was called **once** in 2.5 months. Nothing carries a summary forward automatically.

## Scope

At wake, inject **only finished summaries**. Never raw material.

| host | injection point |
|---|---|
| Claude, Codex, Grok, Gemini | `SessionStart` — the hook's stdout, which the plugin wrappers already pass through (`preserve_stdout=1`) |
| **Kimi** | **first `UserPromptSubmit`** — Kimi's `SessionStart` is fire-and-forget and ignores return values |
| ~~Antigravity~~ | out of scope — its `PreInvocation` stdout is JSON and this repository has not verified a field that carries context. Split out as #1714 |

The table has two rows; the code has one mechanism. Injection is emitted **at most once per session**, guarded by a per-session marker file in the hook state dir, mirroring `hookSessionEndMarkerPath` / `markHookSessionEnded`. Kimi needs that guard structurally (`UserPromptSubmit` fires on every prompt); the `SessionStart` hosts need it because `SessionStart` fires again on resume and on compact. Do not build a per-host strategy layer: the host-specific part is which hook carries the payload, nothing else.

### The session-start hook already owns that stdout

`runHookSession` prints a bare session-id line on `start` (`presentation/cli/hook_session.go:106` and `:147`), and the wrappers deliver it to the host — so a stray session id is **already** being injected into every Claude / Codex / Gemini session's context, and nothing in this repository consumes it. `git log -S` traces it to `1a3d7543` and `80bcb039`; neither commit names a consumer, and no test asserts it. That line is removed as part of this issue: you cannot own the injection channel without owning what is already on it.

## What gets injected

Consolidation owns *making* the summary. It cannot own *choosing the audience*: it runs per session at stop / gc time and cannot know which workspace will wake next, or when. So the wake path selects — with a fixed deterministic query, no scoring and no relevance model:

1. same `workspace` as the waking session
2. top-level sessions only (`parent_session_id` empty) — subagent work is already subsumed by its parent's summary
3. excluding the waking session itself
4. `degraded = 0` only. A mechanical summary exists so gc can discard bodies without losing the skeleton, and its own text says the *why* is gone. Spending wake budget on it teaches the agent nothing
5. newest `started_at` first, backed by `idx_sessions_repo_started_at`
6. take summaries whole, newest first, until the next one would exceed the budget — then stop. Older summaries are dropped entirely; a summary is never cut mid-text

Measured on the live store for `github.com/duck8823/traceary`, 2026-08-01..08: 269 sessions, of which **81 top-level** (188 were subagents). Of those 81, **45 carried at least 64 KiB of material** — about 5.6 consolidation-worthy top-level sessions per day. At an 8 KiB budget that is roughly the last two to three days of work.

## Budget

`wake_injection.budget_bytes` in `config.json`, default **8192**. Explicit `0` disables injection. A config file that is present but unreadable or malformed also disables it — the same policy as `consolidation.threshold_bytes` (#1674), so a broken file cannot switch on something an operator turned off.

The injection budget is independent of the consolidation threshold: changing one must not change the other's behaviour.

## Done when

- Three eligible summaries -> the hook's stdout carries the summaries and **no event body**
- Zero eligible summaries, or only `degraded = 1` ones -> the hook emits nothing (not an empty string, not a bare session id)
- Kimi -> silent at `SessionStart`, injects at the first `UserPromptSubmit` only
- A second `SessionStart` for the same session (resume / compact) injects nothing
- Injected bytes never exceed the budget
- Same summary set and same 8 KiB budget, consolidation threshold varied between 64 KiB and 1 MiB -> byte-identical output
- A store still on migration 46 (no `session_refinements` table) injects nothing rather than failing the hook

Spec: [`.ai/spec/1606.md`](../blob/main/.ai/spec/1606.md) (landed by #1671).

Parent: #1606

duck8823:
Closes #1684

## What this is

The half of the product that was missing. Consolidation makes summaries; nothing carried them forward. `additionalContext` / `hookSpecificOutput` appeared **zero times** in the codebase and MCP `get_context` was called **once** in 2.5 months.

At wake, the hook writes finished summaries for the same workspace to stdout, and the host injects them into model context. Only finished summaries — never raw event bodies.

## The channel was already open

`runHookSession` printed a bare session id on `start`, and the packaged wrappers run the command with `preserve_stdout=1` — so that line was **already** landing in every Claude / Codex / Gemini session's context, with no consumer anywhere in this repository. `git log -S` traces it to `1a3d7543` and `80bcb039`; neither commit names a reader. One test asserted it (`hook_runtime_test.go`), and that assertion is now inverted.

The channel carries summaries or nothing. Not an empty line — an empty line is an injection of nothing.

## Two entry points, one mechanism

| host | where |
|---|---|
| Claude, Codex, Gemini, Grok | `SessionStart` stdout |
| Kimi | first `UserPromptSubmit` — its `SessionStart` ignores return values |

Injection fires at most once per session, guarded by a marker file beside the existing session-end markers (`SessionStart` re-fires on resume and compact; Kimi's prompt hook fires on every prompt). Spool replay passes a **nil** writer rather than `io.Discard` — a discarded write would still consume the marker and silence the live firing.

**Antigravity is out of scope** and split to #1714. Its `PreInvocation` stdout is JSON and the field that carries context is unverified in both the spec and the vendor docs this repository cites; guessing a field name would ship a hook that silently injects nothing while `doctor` reports it installed.

## Selection

Fixed query, no scoring: same workspace, top-level only, excluding the waking session, `degraded = 0`, newest first, summaries taken **whole** until the next would exceed the budget. A degraded row's own text says the *why* is gone — spending wake budget on it teaches the agent nothing.

`ORDER BY` goes through `ts_norm(started_at)`: `started_at` is variable-width on the live store (24–27 chars, measured read-only), so raw lexical DESC misorders whole-second against sub-second timestamps (#1185).

Budget arithmetic is deliberately **not** in SQL — it lives in one Go function so the rule is unit-testable without a database and no summary is ever cut mid-text.

## Budget

`wake_injection.budget_bytes`, default **8192**. Explicit `0` disables. Present-but-unusable config disables rather than defaulting — the same policy #1674 established for `consolidation.threshold_bytes`, so a broken file cannot re-enable something an operator turned off.

Independent by construction: a test pins byte-identical output with the consolidation threshold at 64 KiB and at 1 MiB.

## Sizing

Live store, `github.com/duck8823/traceary`, 2026-08-01..08 (read-only): 269 sessions, 81 top-level, of which 45 carried ≥64 KiB of material — ~5.6 consolidation-worthy top-level sessions/day. At 8 KiB that is roughly the last two to three days of work.

## Surface

Zero new commands, subcommands, flags, or MCP tools. One config key, one removed stdout line.

## Verification

```
go build ./...        ok
go vet ./...          ok
go test ./...         ok
golangci-lint run     0 issues
```

Tests cover: three eligible summaries injected with no body text; zero eligible → empty; degraded-only → empty; other workspace / subagent / the waking session excluded; oversized single summary → nothing; two-fit-three-don't; second `SessionStart` silent; Kimi silent at `SessionStart` and injecting at the first prompt only; threshold-independence; missing `session_refinements` table; unusable config.

Implementation delegated to Grok 4.5 (high). Review, the `ts_norm` ordering fix, the spool-replay `nil`-writer fix, and dropping the unused `StartedAt` field are mine.


duck8823:
**gpt-5.6-sol (Codex verifier)**: 2 HIGH, `ready_to_merge: false`. Budget arithmetic, the nil-writer spool path, config independence, SQL ordering, failure containment and the removed stdout line all came back clean.

Both findings verified against the code. Disposition:

**HIGH 1 — `sanitizeHookStateKey` is not injective** (`hook_state.go:92`). Correct: `a/b` and `a:b` both become `a_b`. But this is not introduced here — `hookSessionEndMarkerPath` and `hookActiveSubagentStatePath` have keyed on the same function since long before #1684, so fixing it means changing the keying contract for three features at once. Every supported host emits UUID-shaped session ids, which sanitize to themselves, so no collision is reachable on the live store today. Filed as #1716 rather than widened into this PR.

**HIGH 2 — TTL not checked on read** (`hook_state.go:605`). The scenario requires a host to reuse a session id more than 30 days later; none does. But the reviewer was reading a real contradiction: the constant said "how long markers are kept" while nothing consults mtime on read, and pruning only runs on the next marker write. The comment was the wrong half, not the code — fixed in 97f6534 to state that the TTL is disk hygiene and that presence means "already injected" at any age. Adding an mtime check would be logic for an unreachable state.

Also split out during review: **#1714** (Antigravity `PreInvocation` injection — its stdout is JSON and the context-carrying field is unverified in both the spec and the vendor docs this repo cites; shipping a guess would mean a hook that silently injects nothing while `doctor` reports it installed).

Verification after the fix: `go build` / `go vet` / `go test ./...` / `golangci-lint run` all clean.

Issue #1722, opened 2026-08-08
v0.40-5: orphan consolidation opens one read-only connection per candidate


duck8823:
## Problem

`SessionOrphanRangeDatasource.LoadMaterial` calls `openForDiscovery` — which calls `Database.openReadOnly` — **once per candidate**, and closes it again on return. `openReadOnly` is not a handle lookup: it builds a fresh `*sql.DB`, runs `PingContext`, and then runs `checkStoreCompatibility` (a `sqlite_master` probe plus a `store_format_state` read).

So consolidating N ranges pays N × (connection setup + ping + two compatibility queries) before doing any of the two range queries that are the actual work.

`Database` already has the mechanism to avoid this — `sharedReadOnly`, which short-circuits both `open` and `openReadOnly`. But it is only ever populated by `NewImmutableReadDatabase`, which exists for benchmark orchestration. Production `NewDatabase` leaves it nil, so every gc candidate takes the slow path.

## Why it matters now

#1719 paces consolidation by a wall-clock budget. Per-candidate overhead therefore converts directly into fewer candidates folded per `gc` run, and a longer drain on exactly the stores that need it most (the measured store has 26,884 unfolded sessions).

It is also not specific to orphan consolidation — any read path that loops over rows and calls a datasource method per row pays the same toll. This issue is the orphan path because that is where it was measured, but the fix should be at the `Database` level.

## Scope

- Give read-heavy passes a way to reuse one read-only handle for their duration, without leaking `*sql.DB` above the infrastructure layer (the domain repository interface must not gain a connection parameter)
- The likely shape is a scoped read session owned by the datasource or by `Database`, entered and left around a pass, with the existing per-call behaviour as the default so nothing else changes
- Confirm the compatibility check still runs once per scope rather than being skipped entirely — it is a correctness guard, not overhead to delete

## Not in scope

Changing `Database.open` (the write path) or the DSN/journal settings.

## Done when

- A consolidation pass over N candidates opens O(1) read-only connections, not O(N)
- The store-compatibility guard still runs for each pass
- No connection type appears in a `domain` or `application` signature

Found while implementing #1719.

Issue #1721, opened 2026-08-08
v0.40-4: let gc delete during a long consolidation drain by folding oldest ranges first


duck8823:
## Problem

#1719 bounds orphan consolidation and, because `delete_old_events.sql` deletes purely by cutoff with **no coverage check**, gates the deletion half of `gc` on the consolidation pass being complete:

```
deletion runs only when  !HasMore && Skipped == 0
```

That gate is correct — deleting events whose range was never folded is the irreversible mistake. But it has a cost: on a store with a large unfolded backlog, `gc` **deletes nothing at all** until the backlog drains over several runs. On the measured store (26,884 sessions, essentially zero refinements) that is a handful of invocations during which `gc` does not do the thing an operator ran it for.

## The observation that fixes it

Deletion only ever touches events older than the retention cutoff (default 90 days). A session whose events are **all newer than the cutoff** cannot be harmed by deletion, so it does not need to be folded before deletion runs.

So the safety condition is not "the backlog is empty". It is:

> no unfolded range contains an event older than the retention cutoff

## Scope

- Order discovery **oldest-first** instead of by `session_id`, so a bounded pass always folds the ranges deletion is about to touch
- Carry enough information out of the pass to evaluate the real condition — the earliest event time of the first *unprocessed* candidate, and of any *skipped* candidate
- Replace the `Complete()` gate in `presentation/cli/gc.go` with that comparison against the cutoff `gc` already computes
- Keep the conservative behaviour as the fallback: if the earliest time cannot be determined, do not delete

Ordering oldest-first also changes the keyset pagination cursor in `session_orphan_range_datasource.go` from `session_id` to the ordering key, so the cursor has to be a composite that is unique — an ordering key alone is not.

## Why it was not done in #1719

#1719's job was to stop the pass being unbounded and all-or-nothing. Adding an age-ordered deletion gate on top would have widened it into a change to what `gc` deletes and when. Splitting keeps that reviewable on its own.

## Done when

- A bounded pass on a large backlog folds the oldest ranges first
- `gc` deletes on a store whose unfolded backlog is entirely newer than the cutoff, even with `HasMore` true
- `gc` still refuses to delete when an unfolded or skipped range reaches past the cutoff
- The pagination cursor is unique under the new ordering

Depends on #1719.

Issue #1720, opened 2026-08-08
v0.39-5: gc retention comparisons are lexical, not ts_norm (#1185 class)


duck8823:
## Problem

Every retention comparison in `gc` compares timestamps **lexically**, not through `ts_norm`. The stored columns are variable-width RFC3339Nano, so the comparison is wrong at the boundary — the #1185 hazard, in the one place where being wrong deletes data.

The cutoff is bound as `formatTimestamp(before)` = `time.RFC3339Nano`, which **strips trailing zeros** from the fraction, so cutoff width varies too.

Affected SQL, all with a bare `< ?`:

| file | column |
|---|---|
| `delete_old_events.sql` | `created_at` |
| `delete_empty_sessions.sql` | `COALESCE(ended_at, started_at)` |
| `count_empty_sessions_after_event_gc.sql` | same, plus `events.created_at >= ?` |
| `delete_old_memories.sql` | `updated_at` |
| `delete_old_memory_edges.sql` | `valid_to` |
| `count_old_memory_edges_after_memory_gc.sql` | `valid_to`, `memories.updated_at` |
| `delete_stale_extracted_candidates.sql` | `updated_at` |

## What actually goes wrong

`'.'` is 0x2E and `'Z'` is 0x5A, so a sub-second timestamp sorts **below** the whole-second form of the same instant.

- Cutoff `2026-05-10T00:00:00Z`, event `2026-05-10T00:00:00.5Z` — the event is *after* the cutoff and must be retained, but `.` < `Z` at index 19 makes it compare as less than the cutoff, so it is **deleted**.
- Cutoff `2026-05-10T00:00:00.5Z`, event `2026-05-10T00:00:00Z` — the event is *before* the cutoff and should be deleted, but it compares as greater, so it is **retained**.

## Magnitude, stated honestly

The error window is under one second at a 90-day boundary, and the cutoff is `now - keepDays` so it almost always carries a fraction — which puts real stores in the *under*-deletion case (a few rows survive one extra second). This is not eating anyone's data today.

It is worth fixing anyway because it is the same defect class as #1185 in the code path with the least reversible consequence, and because the correct form is already used elsewhere in the codebase (`list_ended_or_stale_sessions_for_orphan.sql`, `select_latest_event_id_for_session.sql`).

## Scope

- Wrap both sides in `ts_norm` in each file above
[...]

PR #1723, opened 2026-08-08
fix: bound orphan consolidation and stop aborting on the first failure

### LINKED ISSUE #1719: v0.34.0-50: traceary gc orphan consolidation is unbounded and aborts on the first failure
## Problem

`OrphanConsolidationUsecase.Consolidate` (#1683, merged) processes **every** discovered candidate in one pass and stops at the first error.

- `OrphanConsolidationInput` carries only `StaleAfter` and `DryRun` — no limit, no batch size, no cursor (`application/usecase/orphan_consolidation_usecase.go:11`)
- `DiscoverCandidates` returns the whole candidate set, and the loop walks all of it (`orphan_consolidation_usecase_impl.go:69`)
- `produceDegraded` returning an error aborts the run: `return apptypes.OrphanConsolidationResult{}, err` — one unreadable session kills the pass
- `presentation/cli/gc.go` then aborts the whole gc run for the `events` / `sessions` / `all` targets, because deleting events after the summary failed to land is the irreversible half

The only bound anywhere on this path is `orphanMechanicalSummaryMaxBytes`, which caps a single summary's text.

## Why this is a spike, not a steady load

Measured on the live store 2026-08-08 (`mode=ro`): **26,884 sessions, 548,841 events, and 4 rows carrying a summary** — all of them 34-byte placeholders. `session_refinements` does not exist yet (migration 046 has not shipped).

So on the first `traceary gc` after upgrading to v0.34, essentially **every session in the store is a candidate at once**. That single command walks ~26.8k ranges, reads each range's events through `LoadMaterial`, and writes ~26.8k refinements — with no progress output, no checkpoint, and no way to ask for less.

Every subsequent run is small, because the refinements persist. The spike is one-time per store, and it lands on every existing installation on the same upgrade.

Two things make it worse than a slow command:

1. **All-or-nothing.** A single failing session aborts the pass *and* the deletion half of gc. The operator gets no partial progress reported, even though refinements written before the failure do persist — so re-running does make progress, but nothing says so.
2. **It is on the critical path for the release gate.** #1676 can only discard a transcript body once its range is folded. Until this pass completes for a session, none of that session's bodies are discardable. The storage reduction in Wave 3 is gated behind this one unbounded command.

It does **not** fire from a hook: `runOpportunisticSessionGC` calls `CloseStaleSessions` only, and `runOpportunisticArchiveThenGC` is gated on `retention.mode = archive_then_gc`, which is off by default. The spike is confined to an explicit `traceary gc`, which is the right place for it — it just has to be survivable there.

## Scope

- A bound on how many ranges one invocation consolidates, with the remainder left for the next run. Default it to something a laptop finishes without the operator wondering whether it hung
- Per-candidate error tolerance: a session that cannot be read is skipped and counted, not fatal. Only a failure that indicates the mechanism itself is broken (repository unavailable, refinement writes rejected) aborts
- Report what happened: consolidated, skipped, remaining. `--dry-run` already counts candidates; make the same numbers available after a real run
- Progress for a long pass, since the first one on an existing store is by definition long

No new command or subcommand. This is the shape of the step that already runs inside `store gc`.

## Done when

- A store with 25k unfolded sessions completes `traceary gc` in bounded work, and re-running continues where it left off
- One unreadable session is skipped and reported; the rest of the pass and the deletion half still run
- The result reports consolidated / skipped / remaining counts
- A mechanism-level failure (repository unavailable) still aborts before any deletion

Found while scoping Wave 3 (#1606). Blocks #1676.
## Design decisions (settled 2026-08-08, reviewed with gpt-5.6-sol)

**The bound is wall-clock, not just a count.** Discovery is capped by a candidate limit (protects against an open-ended scan); the processing loop is capped by a time budget (what an operator actually experiences when a command has not returned). Both are constants — **no new flag, no new subcommand**.

**Discovery must filter coverage in SQL.** A `LIMIT` on the unfiltered ended/stale list would return the same already-folded sessions on every subsequent run and make zero progress forever. The coverage predicate moves into `list_ended_or_stale_sessions_for_orphan.sql`, deliberately over-inclusive, with Go re-checking the exact "latest strictly after covers_to" rule.

**`HasMore` is computed from valid candidates, never from raw row counts.** The datasource paginates by keyset until it has `limit+1` *valid* candidates or the scan is exhausted. Deriving it from row counts would let an over-inclusive filter report "more remains" forever while producing nothing — which, given the deletion gate below, would block deletion permanently.

**Error tolerance: skip per candidate, abort after 3 consecutive failures.** Isolated failures are counted and logged; three in a row means the mechanism is broken rather than one session's data. Context cancellation and discovery errors abort immediately and are never treated as skippable.

**Deletion is gated on a complete pass.** `delete_old_events.sql` is `DELETE FROM events WHERE created_at < ?` — it checks no coverage at all. So bounding consolidation without gating deletion would be a regression: `gc` would delete events whose ranges were never folded. Deletion runs only when `!HasMore && Skipped == 0`, for the targets consolidation protects.

## Split out rather than bundled

- **#1721** — the gate above means `gc` deletes nothing until a large backlog drains. Folding oldest ranges first lets deletion resume as soon as nothing unfolded reaches past the retention cutoff
- **#1720** — every gc retention comparison is lexical rather than `ts_norm` (#1185 class), including `delete_old_events.sql`
- **#1722** — `LoadMaterial` opens a fresh read-only `*sql.DB` per candidate, which converts directly into fewer candidates folded per time budget

duck8823:
Closes #1719

## What was wrong

`OrphanConsolidationUsecase.Consolidate` walked **every** discovered candidate in one pass and `return`ed on the first per-candidate error. On a store that has never been consolidated — 26,884 sessions and effectively zero refinements on the measured store — that is essentially every session at once, and one unreadable session aborted the pass *and* the deletion half of `gc`.

## What changed

**Bounded.** Discovery takes a candidate limit (5,000); the processing loop takes a wall-clock budget (2 min). Both are constants — **no new command, subcommand, or flag**. The budget is the bound that matters: an operator experiences "has this hung", not a row count.

**Discovery filters coverage in SQL.** A `LIMIT` over the unfiltered ended/stale list would return the same already-folded sessions on every later run and make zero progress forever. The coverage predicate moves into `list_ended_or_stale_sessions_for_orphan.sql`, deliberately over-inclusive, with Go still applying the exact "latest strictly after `covers_to`" rule.

**`HasMore` is computed from valid candidates, never from row counts.** The datasource paginates by `session_id` keyset until it has `limit+1` *valid* candidates or the scan is exhausted. Deriving it from raw rows would let an over-inclusive filter report "more remains" forever — which, given the deletion gate below, would block deletion permanently.

**Failures are survivable.** A candidate that fails is skipped, counted, and logged. Three *consecutive* failures abort with an error instead, on the assumption that the mechanism is broken rather than one range's data. Context cancellation and discovery errors abort immediately and are never treated as skippable.

**Deletion is gated on a complete pass.** `delete_old_events.sql` is `DELETE FROM events WHERE created_at < ?` — it checks no coverage at all. So bounding without gating would be a regression: `gc` would delete events whose ranges were never folded. Deletion now runs only when `!HasMore && Skipped == 0`, for the targets consolidation protects:

```
Orphan refinements: 5000
Orphan ranges skipped: 2
Deletion skipped: orphan ranges are not fully consolidated; re-run gc to continue
```

Exit 0 — the pass made real progress and its refinements are durable. Documented in `docs/cli/README.md` and `.ja.md`.

## One change beyond the issue, with the measurement

The discovery query reads `events.created_at_norm` instead of calling `ts_norm(created_at)`. This is inside the SQL the fix already rewrites, and it decides whether the fix works: discovery runs the latest-event subquery for **every session it scans**, so an unindexable ordering there costs a sort per session — in the query whose entire purpose is to be cheap enough that the budget goes to folding.

Measured with `EXPLAIN QUERY PLAN` on a fixture store:

| | `ts_norm(e2.created_at)` | `e2.created_at_norm` |
|---|---|---|
| latest-event subquery | `SEARCH e2 USING COVERING INDEX idx_events_session_created_at_id_desc (session_id=?)` + **`USE TEMP B-TREE FOR ORDER BY`** | `SEARCH e2 USING COVERING INDEX idx_events_session_created_at_norm_id_desc (session_id=?)`, no sort |
| activity probe | `SEARCH e ... idx_events_session_created_at (session_id=?)` | `SEARCH e ... idx_events_session_created_at_norm_id_desc (session_id=? AND created_at_norm>?)` |

`created_at_norm` is backfilled and trigger-maintained by migration 031, and `migrate_test.go` already pins `created_at_norm = ts_norm(created_at)`. `sessions` has no such column, so `started_at` still goes through `ts_norm`. `TestSessionOrphanRangeDatasource_DiscoveryPlanUsesNormalizedTimestampIndex` asserts the plan has no temp B-tree.

## Tests

Usecase: bounded pass reports `HasMore`; a single failure is skipped and counted; three consecutive failures abort; two failures separated by a success do not; budget exhaustion sets `HasMore` without an error (stub clock, no real time); a cancelled context errors rather than skipping; dry run respects the limit and skips a failing `LoadMaterial`.

Datasource: limit is respected and `HasMore` is set; **progress across pages** — consolidate page 1, discover again, and the second call must return the *next* candidates, which is the failure mode the SQL coverage filter exists to prevent; sessions with no events are excluded; the pre-migration-47 store test still passes.

CLI: an incomplete consolidation does not call `CollectGarbage` and prints the skip line; a complete one deletes; dry-run still reports counts when incomplete.

## Split out rather than bundled

- **#1721** — the gate means `gc` deletes nothing until a large backlog drains. Folding oldest ranges first lets deletion resume as soon as nothing unfolded reaches past the retention cutoff
- **#1720** — every gc retention comparison is lexical rather than `ts_norm` (#1185 class), including `delete_old_events.sql`
- **#1722** — `LoadMaterial` opens a fresh read-only `*sql.DB` per candidate, which converts directly into fewer candidates folded per budget

## Verification

`go build ./...`, `go vet ./...`, `go test ./...`, `go tool golangci-lint run` — all clean (0 lint issues).

Implementation delegated to Grok 4.5 (high); design fork reviewed with gpt-5.6-sol; reviewed, corrected, and verified by Claude Opus 5. Grok's pass left the duplicated failure-handling block and the `ts_norm` function call in the subquery; both were fixed here.


duck8823:
## gpt-5.6-sol adversarial review — 3 HIGH, dispositioned

Each was checked against the code rather than accepted on assertion.

### 1. `Complete()` does not guarantee deletion safety — ✅ real, pre-existing, now named + filed as #1724

Correct. Discovery sees ended-or-stale sessions plus recorded markers. A session that stays continuously active for longer than the retention window is never discovered, so `HasMore` is false, `Complete()` reports a clean pass, and `gc` deletes its pre-cutoff events unfolded.

This is **not introduced by this PR** — `main` has the same ended-or-stale predicate in discovery and the same cutoff-only rule in `delete_old_events.sql`. What this PR did was make the safe cases explicit, which made the unsafe one visible.

Fixing it means changing what discovery covers, which is a separate design question (and #1724 notes that making deletion itself coverage-aware may be the smaller total change, closing this and #1721 at once). Filed as **#1724**. The `Complete()` doc comment overclaimed — it said an undiscovered range is one whose events would be deleted with nothing summarising them, as if that were prevented. It now states the actual guarantee and points at #1724.

### 2. `created_at_norm` NULL or divergent → false negative — ✅ valid concern, invariant now pinned

The mechanism is right: `ORDER BY created_at_norm DESC` sorts NULL last in SQLite, so a NULL row is never picked as the session's latest event. Discovery would read an older event as the latest, `covers_to` would match it, and the session would leave the candidate set — after which `Complete()` reports clean and deletion removes events that were never folded.

This PR does introduce the dependency, so it should not rest on "should be impossible":

- migration 031 backfills every existing row and installs `AFTER INSERT` / `AFTER UPDATE OF created_at` triggers, whose `WHEN` clause fires precisely when `created_at_norm IS NULL` or disagrees
- no migration after 031 rebuilds `events` (the only rebuilds are 001 and 006)
- `migrate_test.go` already pins `created_at_norm = ts_norm(created_at)` for the backfill

The untested half was the trigger path. `TestSessionOrphanRangeDatasource_InsertedEventsCarryNormalizedTimestamp` now pins it across whole-second, sub-second, and nanosecond widths — the three renderings whose differing widths are what breaks lexical ordering in the first place.

Worth noting the alternative was rejected deliberately: `COALESCE(created_at_norm, ts_norm(created_at))` would be NULL-safe but unindexable, which defeats the reason for the change (measured: the function form plans `USE TEMP B-TREE FOR ORDER BY` per scanned session).

### 3. `orphanConsolidation == nil` is fail-open — ✅ fixed

`main.go:283` always wires it, so nil is unreachable in the shipped binary — but the consequence of a future wiring regression is irreversible deletion of events nothing summarises. Now fails closed with an explicit refusal, and `CollectGarbage` is not called. `memories` / `memory_edges` still prune without it, since they hold nothing consolidation preserves.

The existing test `without orphan consolidation still reports zero orphans` pinned the fall-through; it is replaced by `missing orphan consolidation refuses to delete` plus a case proving the unprotected targets are unaffected.

---

Verification after the fixes: `go build ./...`, `go vet ./...`, `go test ./...`, `go tool golangci-lint run` — all clean, 0 lint issues.

Issue #1718, opened 2026-08-08
v0.34.0-49: retire the legacy full-corpus search index


duck8823:
## Problem

Now that #1717 moved reads onto the bounded projection, the migration-032 index family is dead weight — and it is the single largest thing in the store.

Measured on the live store (`mode=ro`, schema 34):

```
event_search_fts_data      11.04 GiB
event_search_documents      5.11 GiB
                           ------
                           16.15 GiB   = 65% of a 24.7 GiB store
```

`event_search_documents` is a full second copy of the body text; `event_search_fts_data` is the trigram index over it at roughly 2.16x. Neither has a reader now that #1733 and #1735 made the tiered path answer in every projection state — there is no state in which tiered needs a fallback.

The v0.34 gate is 24.71 GiB -> ≤4 GiB. **This issue is where most of that comes from.**

## Upgrade shape (decided)

Released schema tops out at 34, so the first v0.34 startup applies 17+ migrations at once. `infrastructure/sqlite/migrate.go:34-41` applies every pending migration unconditionally — `MigrationExecutionClass` is read only by `prepared_migration_catalog.go` when building a prepared-publication plan, and does **not** gate normal startup. A `DROP TABLE` migration on the 16 GiB family would therefore run as one uninterruptible transaction at the user's next `traceary` launch, and roll back entirely if interrupted.

So the reclaim is split:

- **the migration does only constant-cost work** — it removes the five legacy writer triggers and the view, so the family stops being maintained and stops growing. Startup keeps succeeding exactly as before.
- **the 16 GiB removal stays an explicit, resumable, opt-in command.** Stores whose operator never runs it do not shrink; the v0.34 gate is met by stores that do.
- **`traceary doctor` surfaces it** so the operator does not have to know it exists.

### 1. Migration (`constant_in_place`)

Drop the five source-side triggers `BeginSearchRetirement` drops today (`infrastructure/sqlite/search_maintenance.go:164-169`):

```
events_search_after_insert
events_search_after_body_update
command_audits_search_after_insert
command_audits_search_after_update
command_audits_search_after_delete
```

and `DROP VIEW event_search_projection`.

The three `event_search_documents_after_*` triggers are left alone: once the source triggers are gone nothing inserts into `event_search_documents`, so they are inert, and SQLite drops them implicitly with their table.

[...]

Issue #1685, opened 2026-08-07
v0.34.0-45: apply the payload codec to the existing corpus


duck8823:
> **Rescoped 2026-08-09 after design review.** The original scope was "move event bodies to a side
> table **and** compress them". The move half rested on a premise that measured false, and the
> compress half turned out to already have an end-to-end wiring waiting for it. This issue is now the
> compression backfill. The side table moved to #1743 (measure first), and the `command_audits` half
> of the gate contribution moved to #1742. Full reasoning, measurements and design:
> [`.ai/spec/1685.md`](../blob/main/.ai/spec/1685.md).

## Problem

Migration 036 shipped the whole payload codec contract — five metadata columns on `events`,
compatibility counters, guard triggers — and said in its own header what was left:

> "Reader-first payload codec foundation. v0.34 writes identity only; zstd activation and **legacy
> backfill are intentionally deferred to v0.35**."

`infrastructure/sqlite/payload_codec.go` implements a versioned zstd codec that has never compressed a
row. **This issue is that legacy backfill**, and it is the only thing that applies the codec to the
existing corpus. Measured per-row through that same codec at its default level:

| corpus | raw | zstd per row | ratio |
|---|---|---|---|
| prompt | 84.0 MiB | 27.2 MiB | **3.09x** |
| command_executed | 41.1 MiB | 16.0 MiB | **2.57x** |
| transcript | 39.4 MiB | 16.1 MiB | **2.44x** |

(Whole-file compression of the same samples reports 9-12x. That is an artefact of cross-row
redundancy and must not be used; bodies are compressed independently.)

Against the #1620 store gate this issue is worth **-0.584 GiB** (`prompt` 0.739/3.09 plus
post-#1682-dedupe `transcript` 0.143/2.44). The other **-1.087 GiB** of the 1.67 GiB total is
`command_audits`, split into #1742 — which is therefore a dependency of #1679 and #1620 alongside
this issue.

## Why the side table is not here

The original justification was that discarding a body as `UPDATE body=''` does not release pages
while a side-table `DELETE` does. Measured (50,000 rows × 4 KiB, `modernc.org/sqlite`):

| | page_count | freelist |
|---|---|---|
| before | 56,525 | 0 |
| after `UPDATE events SET body=''` | 56,525 | **50,000** (195.3 MiB of 220.8) |
| after `DELETE FROM events` | 56,525 | 56,522 (220.8 MiB) |

[...]

Issue #1680, opened 2026-08-07
v0.34.0-40: execute the tiered search cutover


duck8823:
## Scope

Execute the two-tier design from #1679 on a real corpus, using the existing generation machinery. No new batch runner.

- Build the new generation alongside the old
- Verify the session tier answers old-history queries before abandoning the old generation
- Reclaim the old generation
- Report before/after index-family bytes

## Corrected premise (2026-08-08)

The original text said this moves "the live projection from the full-corpus trigram design", and expected 15.83 GiB -> under 2 GiB. That conflated two different index families:

- The 15.83 GiB on the live store is the **legacy migration-032 family**. It is not a generation of the bounded projection and this issue's generation machinery cannot reclaim it. Removing it is **#1718**
- What this issue cuts over is the **bounded projection**, which #1717 makes the authoritative read path

So the byte expectation belongs to #1718, not here. What this issue owns is that the cutover is **safe and resumable** — build, verify, reclaim, without leaving two generations live.

Also: migrations 035–047 have never shipped (v0.33.1 tops out at 034), so no store in the field has ever run a generation cutover. The first real one will be an upgrade, not a re-run.

## Constraints

- Resumable; interruption must not leave both generations live indefinitely
- Never run against the live store from a repo build (`ref_live_db_migration_gap`)

## Done when

- A fixture store cuts over, old-history search still finds sessions, and the bounded family lands under the configured budget
- An interrupted cutover resumes
- The reported before/after numbers name which family they measure, so they cannot be read as the store-wide reduction (that is #1718)

Depends on #1717 and #1679.

Spec: [`.ai/spec/1606.md`](../blob/main/.ai/spec/1606.md) (landed by #1671)

Parent: #1606

PR #1726, opened 2026-08-08
feat: make the bounded search projection the authoritative search read path

### LINKED ISSUE #1717: v0.34.0-48: make the bounded search projection the authoritative read path
## Problem

The bounded search projection (migrations 038–042, `infrastructure/sqlite/search_projection_rebuild.go`) is **written but never read**.

- The search read path is `infrastructure/sqlite/event_search_query.go:399` — it queries `event_search_fts`, the migration-032 index
- `search_projection_recent_fts` appears in exactly one non-test file: `search_projection_rebuild.go`
- Migration 038 says so in its own first line: `-- Additive, rebuildable projection. Legacy search remains authoritative.`
- No migration up to 047 drops the legacy index

So the whole family — recent trigram tier, session summaries, session keywords, command aggregates, generation lifecycle — is maintenance cost with no reader. #1679 notes this for the session tier; it is true of the entire family.

This blocks Wave 3. #1679 tunes `recent_byte_limit` on `search_projection_recent_documents`, and #1680 cuts over "the live projection" — but the bytes on the live store belong to the legacy index, not to this one. Measured today (`mode=ro`, schema 34):

```
event_search_fts_data      11.04 GiB
event_search_documents      5.11 GiB
                           ------
                           16.15 GiB   (the 15.83 GiB the spec attributes to "the trigram index")
```

Nothing in Wave 3 as written reduces those bytes, because nothing in Wave 3 touches the index that holds them.

None of migrations 035–047 has shipped: v0.33.1 tops out at 034, and the live store is correctly at 034. So this is not a production defect — it is an unfinished path that v0.34 would ship in its current write-only shape.

## Scope

Make the bounded projection the authoritative read path, so that #1679's tiering has something real to tier.

- `event_search_query.go` reads the recent tier (`search_projection_recent_fts`) instead of `event_search_fts`
- A query that falls outside the recent window resolves through the session tier (`search_projection_session_summaries` / `_keywords`) and returns **sessions**
- Result shape: events and sessions are returned as **separately labelled groups**, not interleaved — a session row means "the trail is here, open it", not "this is the matching line". Confirmed with the operator 2026-08-08.
- The legacy index stays in place and keeps being maintained. Removing it is #1718 — a read path that was missed must be recoverable by reverting one commit, not by rebuilding 16 GiB.

No new command, subcommand, flag, or MCP tool. `traceary search` keeps its name and its inputs.

## Alternative considered

Delete the 038–042 family instead and tier the legacy 032 design in place. Rejected: the generation lifecycle, eviction and resumability that #1680 depends on live in the new family, and rebuilding them on the old design is strictly more work than wiring what exists.

## Done when

- `traceary search` returns results from the bounded projection, with old-history hits appearing as labelled session rows
- The legacy index is still written and still queryable through a documented fallback, so the switch is revertible
- A fixture store with both families present returns the same recent-window hits through the new path as through the old one
- The JSON contract change (a result row that identifies a session rather than an event) is documented in the contract snapshot

Parent: #1606

Follow-up filed from this issue: #1725 (two competing tiering mechanisms on `traceary search`).

duck8823:
## What

Makes the bounded search projection the authoritative read path for `traceary search`.

When a projection generation is `complete` with a non-empty `active_generation_id`:

- recent full-text hits come from `search_projection_recent_fts`
- history older than the recent window is answered through `search_projection_session_summaries` / `_keywords` and returned as a separate **SESSIONS** group
- a SESSIONS row means "the trail is here, open it", never "this is the matching line", so the two groups are labelled and never interleaved

Stores that have never rebuilt a generation — every field store today — fall through to the migration-032 index silently. Search never errors solely because the projection is idle, so nobody has to rebuild to keep searching.

## Three things worth reviewing closely

**1. Short queries bypass both tiers.** Neither trigram tokenizer can match a query shorter than three characters. Routing those into the projection returns silently-empty results — measured, not theorised: a probe on this branch returned 1 hit before the projection was activated and 0 after. `selectShortQueryCandidateIDs` now runs ahead of the tier split and resolves through the bounded decoded scan, which reads only `events` and `command_audits` and therefore survives #1718. Pinned by `TestSearchProjectionReadPath_ShortQueryAnswersUnderBothTiers`, which asserts the result is identical before and after the projection becomes authoritative.

**2. `sessions.started_at` goes through `ts_norm` on both sides.** `events` has a persisted `created_at_norm`; `sessions` does not. RFC3339Nano is variable width and `'.'` 0x2E sorts below `'Z'` 0x5A, so a raw `started_at >= ?` comparison drops sub-second timestamps out of range — the #1185 hazard. Pinned by `TestSearchProjectionReadPath_SessionFromFilterHandlesSubSecondStartedAt`.

**3. `--json` keeps its top-level array for the whole of v0.34.** `traceary search` is **Public** in `docs/cli-stability.md`, where the `--json` byte shape is contract and breaking changes need the deprecation flow plus at least one minor of overlap. An array → object change breaks every `traceary search --json | jq '.[0]'`, and there is no reading of "additive only" that covers it. v0.34 also has a hard no-new-flags rule, so `--format=v2` is not available as an escape.

Silently dropping the sessions would be worse than a loud break: a scripted consumer searching old history would get `[]` and could not distinguish it from "no results". So stdout stays byte-identical and the omission is announced on **stderr**, with the count and the v0.35 shape:

```
traceary: 2 matching session(s) are not included in --json output. Run the same
search without --json to see them. From v0.35 `search --json` emits an object
with `events` and `sessions` instead of an array.
```

The deprecation is recorded in `docs/cli-stability.md` / `.ja.md` under the longer-window rule and in both changelogs. This was consulted with gpt-5.6-sol, which rejected shipping the envelope now, rejected omitting sessions silently, and rejected a `result_type` discriminator inside the array.

## Verification

```
go build ./...          # clean
go tool golangci-lint run  # 0 issues
go test ./...           # exit 0
```

## Review notes

- The golden `search with session hits keeps the top-level event array` asserts stdout against `single_event.golden.json` — the same bytes as a search with no session hits. That is the contract being pinned.
- `SearchLegacyPage` deliberately keeps reading the migration-032 index directly rather than routing through the new selector, so rollback and parity comparison stay independent of the cut-over.
- `SearchSessionHits` fails closed when `--kind` is set: kind cannot be applied to session summaries, and returning unfiltered session rows for a filter the user set would be wrong.

## Known gap, filed rather than widened

The MCP `search` tool has no session tier — it returns events only, so an agent searching old history through MCP sees an empty result where the CLI would show SESSIONS. Fixing it changes a frozen MCP response shape, which is outside this issue. Follow-up filed.

Closes #1717


duck8823:
## Review round 1 — gpt-5.6-sol (read-only), reviewed by Claude Opus 5

Four HIGH findings. Three fixed in de3b9c43, two filed as follow-ups. Each was verified against the code before acting; none were taken on assertion.

### ✅ Fixed — projection freshness (CONFIRMED, the serious one)

> `event_search_query.go` switches to the projection on `state='complete'` alone, without checking freshness. After a rebuild completes, a newly saved event is invisible to search.

Confirmed against the schema:

- `schema/sqlite/migrations/000038_add_bounded_search_projections.sql:10` — the events-insert trigger only appends to `search_projection_source_sequence` and bumps `source_revision` **while rebuilding**. It never inserts into `search_projection_recent_documents`.
- `schema/sqlite/migrations/000041_add_search_projection_generation_lifecycle.sql:16-50` — drift triggers fire on event UPDATE/DELETE and on command_audits, but **not** on event INSERT. A complete generation stays complete.
- `schema/sqlite/migrations/000032_add_event_search_fts.sql:124` — the legacy index *does* have an insert trigger, which is why it never had this problem.

So a complete generation is a snapshot, and everything recorded after it fell out of search — silently, for as long as nobody rebuilt. Reachable today via `traceary store search-projection start` + `resume`, not only after #1680.

Fix: `selectProjectionEventSearchCandidateIDs` merges the tail (`search_projection_source_sequence.sequence > high_water`) into the projection result, matched through `decodedEventSearchMatch` so it stays correct once #1685 compresses bodies. The projection side fetches `limit + offset + len(tail)` rows, which is the tightest bound that cannot drop a row from the requested page, then both sets are re-ordered and paged together. When the tail exceeds the bounded candidate limit the generation is too stale to answer cheaply and search falls back to the legacy index rather than return a knowingly incomplete page.

Pinned by `TestSearchProjectionReadPath_EventsAppendedAfterRebuildStayFindable`. The seed helper was also corrected — it hard-coded `high_water = 100`, which is why no existing test caught this.

### ✅ Fixed — session group repeated on every page

> The session tier applies `Limit` but ignores `Offset` and `PageAnchor`.

Confirmed. The session group points at trails the event tier could not reach; it is not a paginated list. It is now returned for the first page only. Pinned by a subtest in `TestSearchProjectionReadPath_FiltersApplyToBothTiers`.

### ✅ Fixed — found while verifying the above, not in the review

`search_projection_recent_fts` is `tokenize='trigram case_sensitive 1'` and `eventSearchFTSPhrase` folds the query to lower case, but `search_projection_rebuild.go` stored `d.Text` raw while the legacy writer stores `lowerSearchASCII(body)`. Every term containing an uppercase letter was therefore unfindable through the projection and findable through the legacy index. The rebuild now folds at write time (length-preserving, so `decoded_bytes` is unaffected). Pinned by `TestSearchProjectionRecentDocumentsAreASCIIFoldedForCaseSensitiveFTS`, which runs a real rebuild rather than seeding the tables.

Worth noting: the seeded read-path tests passed only because the helper folded the body itself, with a comment explaining why. That workaround was hiding the defect.

### 📋 Filed, not fixed here — #1728

> `--kind` empties the session tier entirely; `--from`/`--to` and `--failures-only` are session-granular rather than event-granular.

Both confirmed, both design decisions rather than slips. Session summaries carry no per-kind breakdown, so filtering them by kind is impossible and returning them unfiltered would be wrong — but the resulting emptiness is silent, which is exactly what this PR was careful to avoid for `--json`. Likewise a session hit means "somewhere in this trail", which may be the right meaning but is currently unstated. Deciding either would change user-visible search semantics and belongs in its own issue.

### 📋 Filed, not fixed here — #1727

The MCP `search` tool has no session tier, so an agent searching old history gets an empty result where the CLI shows SESSIONS. Fixing it changes a frozen MCP response shape.

---

`go build ./...` clean, `go tool golangci-lint run` 0 issues, `go test ./...` exit 0.

duck8823:
## Review round 2 — gpt-5.6-sol (read-only), reviewed by Claude Opus 5

Scoped to `de3b9c43`, which was written *after* round 1 and is the most intricate part of this PR (the tail merge). Six specific questions were put to the reviewer rather than a general pass.

**CRITICAL: none. HIGH: one, and it is outside this PR.**

### Answers to the six questions

1. **Merge fetch bound is sufficient.** `offset + limit + len(tail)` (and `limit + len(tail)` under a page anchor) holds even when the two sets overlap: subtracting at most `len(tail)` duplicates still leaves every unfetched projection row ranked below the requested page.
2. **Ordering is consistent.** All three event paths order by `created_at_norm DESC, id DESC` and apply the page anchor as a strict `<` on the same tuple, so the merged page matches what the legacy path would produce.
3. **The `withinBudget=false` fallback is safe.** Migration 000032's insert trigger keeps `event_search_documents` populated for exactly the events in the tail, so falling back cannot silently return nothing.
4. **Short queries apply the full filter set** — workspace, session, client, agent, kind, from/to, page anchor, failures-only — plus offset and limit, through the same shared builder.
5. **The folding fix has no other reader.** Nothing reads `body_text` expecting raw text, and ASCII folding does not change UTF-8 byte length, so `decoded_bytes` stays accurate.
6. **The session tier is fully `ts_norm()`-normalised** and the first-page-only guard is correct.

### 📋 Filed, not fixed here — #1729

The one HIGH is a pre-existing defect in a different read path: `infrastructure/sqlite/sql/list_sessions.sql:18` and `:72` order by raw `s.started_at` while lines 16-17 of the *same query* already filter through `ts_norm()`. Verified, and the sweep found two more sites in `payload_rehearsal.go:685` / `:917`, whose `started_at` is likewise RFC3339Nano. `traceary list sessions --limit 1` returns the older session when the newer one carries a sub-second timestamp.

Not touched by this PR and not in its read path, so it is filed rather than folded in.

---

CI green on `de3b9c43` (Test / Lint / Docs / Integrations all pass). Round-1 findings are all either fixed here or filed as #1727 / #1728; round 2 adds #1729. Merging.

Issue #1730, opened 2026-08-08
v0.40-7: batch command audit payload hydration


duck8823:
`HydrateCommandAudits` (`infrastructure/sqlite/command_audit_payload.go`, added by #1675) decodes one event and one field per query, and each `hydrateAuditPayload` call re-runs a `pragma_table_info('command_audits')` probe before touching the row.

That is up to **2 queries per field per event**: 2 for the command-only display paths, 6 for the full-payload sensitive paths.

### Where it costs

| caller | fields | when |
|---|---|---|
| `presentation/mcpserver/server.go` list / search / context | command | every MCP call an agent makes |
| `presentation/cli/list.go`, `search.go`, `tail.go`, `context_command.go` | command | every listing |
| `presentation/cli/list.go --sensitive`, `doctor_sensitive_access.go` | command + input + output | on demand |

The MCP path is the one that matters: agents hit `list_events` and `search` constantly, and every `command_executed` row in the page now costs a probe plus a decode round trip where the body used to arrive in the row itself. On a local SQLite file each query is cheap, so this is not urgent — but it is per-row work that does not need to be per-row.

### Two independent fixes

1. **Hoist the schema probe.** Whether `command_audits` has the codec columns is a property of the store, not of the row. Resolve it once per store (or per hydration call) instead of once per field per event.
2. **Batch the fetch.** Select the requested payload columns for all event IDs in one statement — `WHERE event_id IN (SELECT CAST(value AS TEXT) FROM json_each(?))`, the same shape `hydrateEventSearchCandidatesQuery` already uses — then decode in Go. Decoding stays per row; the round trips do not.

Both preserve the integrity checks and the `maxDecodedPayloadBytes` limit, which is the reason the payload columns go through this path at all.

### Not in scope

Reverting to a raw JOIN. That is exactly the defect #1675's second commit (`eb1ab214`) fixed: the columns are codec-managed and reading them undecoded breaks under #1618 and fails the sensitive-path classifier open.

### Done when

- A page of N `command_executed` events costs O(1) queries rather than O(N × fields)
- `TestListJoinDoesNotSurfaceEncodedCommandPayloads` still passes unchanged — the non-identity codec, the integrity check and the size limit all still apply
- The measurement is stated for a realistic page size, not a microbenchmark of the query alone

Related: #1675, #1618.

Issue #1620, opened 2026-08-01
v0.34.0-10: prove the two-pillar release gate with measurements


duck8823:
## Scope

The gate is measured, not asserted, and it is stated in **units the software cannot game**. `GiB/month` was withdrawn as a gate unit on 2026-08-11 — it is `bytes-per-unit-of-work x work-per-month`, and only the first factor is a property of Traceary ([issuecomment-5249722984](https://github.com/duck8823/traceary/issues/1620#issuecomment-5249722984)). Per-*event* units were then withdrawn too, because the producer decides what an event is ([issuecomment-5249869562](https://github.com/duck8823/traceary/issues/1620#issuecomment-5249869562), after an adversarial review by gpt-5.6-sol).

Denominators are **canonical operations**: one `command_executed` or one `prompt` — units of the operator's work, which Traceary does not control. 228,595 in the measured 30-day window, 320,091 in the store. Baseline is the live store on 2026-08-07 (24.71 GiB, 534,086 events, 26,728 sessions, 2.5 months); note that gc has since changed the population, so baseline-vs-now is a **difference between snapshots**, not an effect attributable to the Wave 3 changes.

| target | unit | gate | measured 2026-08-11 |
|---|---|---|---|
| undiscardable growth | bytes / canonical operation | **<= 5 KiB** | 3.51 KiB |
| ... of which command record | bytes / command execution | **<= 4 KiB** | 3.06 KiB |
| ... of which prompt record | bytes / user turn | **<= 12 KiB** | 7.86 KiB |
| session-tier coefficient (the O(corpus) term) | bytes / session | **<= 64 KiB** | 41.7 KiB |
| resident store, all levers applied | bytes / canonical operation | **<= 13 KiB** | 9.40 KiB |
| event-emission amplification | events / canonical operation | **<= 2.0** | 1.31 |
| whole-store amplification | resident bytes / **distinct** retained source bytes | **<= 3x** | 2.55x |
| recent search index amplification | index bytes / recent source bytes | **<= 4x** | 3.37x |
[...]

Issue #1618, opened 2026-08-01
v0.34.0-51: activate compressed canonical writes


duck8823:
## Goal

Activate zstd encoding for **new canonical event/audit writes**.

## Pulled forward from v0.35 to v0.34 (2026-08-08)

This issue was deferred to v0.35 under #1693. It is now a v0.34 sub-issue, because **neither #1620 gate closes without it.**

`infrastructure/sqlite/payload_codec.go` already implements the versioned zstd codec (`klauspost/compress/zstd`), landed by #1609. Measured on the live store, it has never been applied to a single row:

```sql
SELECT body_metadata_version, body_availability, COUNT(*) FROM events GROUP BY 1,2;
-- (NULL) | available | 550721
```

Measured per row through that codec at its default level: `prompt` **3.09x**, `command_executed` **2.57x**, `transcript` **2.44x**. (Compressing the samples as one stream reports 9-12x; that is cross-row redundancy and must not be quoted.)

Against the #1620 monthly-growth gate of **<= +0.5 GiB/month**, projected from the last 30 days:

| source | uncompressed | compressed |
|---|---|---|
| `transcript`, post-#1682 dedupe | 0.141 | 0.058 |
| command output via `command_audits` | 1.2 | 0.47 |
| `prompt` | 0.617 | 0.200 |
| `session_workspace_observations` | 0.12 | 0.12 |
| **total** | **2.08** | **0.87** |

0.87 still exceeds the gate; #1676's discard of folded output removes the 0.47 term and brings it to **0.40**. But without activation the figure is 2.08 — four times the gate — and no amount of discard closes it.

Two byte-cap proposals (an 8 KiB cap on command output, a cap on `prompt`) and one exclusion proposal (drop bodies for workspaces with no repository identity) were considered and rejected in favour of this, because compression closes both gates **without discarding content**. The reasoning is recorded on #1620.

### Division of labour

- **This issue: new writes.** Growth-gate half.
- **#1685: existing rows.** That issue moves every body into `event_bodies` and therefore rewrites it; compressing on the way is the same pass, worth ~1.67 GiB against the size gate. This issue does not backfill.

## Preconditions

- Codec-aware readers verified on normal CLI, MCP, hook, archive, backup, retention, and search paths.
- The store format gate can reject readers older than v0.34.0 before the first compressed-only write.
- Operators have an ordinary backup appropriate for local OSS data; no dedicated migration/rollback platform is required.

[...]

PR #1731, opened 2026-08-08
refactor: stop storing command output twice

### LINKED ISSUE #1675: v0.34.0-35: stop storing command output twice
## Problem

Command output is persisted **twice**.

Re-measured on the live store 2026-08-08 (`mode=ro`, schema 34) in **bytes**, not characters — the earlier figures on this issue used `LENGTH()`, which counts characters, and understated a multi-byte corpus by 20-50%:

| | rows | bytes |
|---|---|---|
| `events.body` where `kind='command_executed'` | 262,001 | **1.785 GiB** |
| `command_audits(command_text + input_text + output_text)` | 262,021 | **1.780 GiB** |
| of which `command_audits.output_text` | | 1.514 GiB |

`command_audits` is keyed by `event_id`, one row per `command_executed` event. Joined and summed, the two sides are the same volume at **ratio 1.00**. Only 35% match by exact substring, which is what JSON escaping predicts — `events.body` holds the escaped payload, `command_audits` the decoded fields. It is not a partial overlap; it is the same content in two encodings.

Dropping the raw copy is worth **1.785 GiB**, 45% of the entire #1620 store gate.

## De-duplication is not the lever (premise correction, 2026-08-08)

This issue previously claimed ~2.1 GiB of reclaimable duplication *within* command output, and #1620's gate table credited 2.10 GiB to it. Measured:

| corpus | total | distinct | duplicate share |
|---|---|---|---|
| `events.body` command_executed | 1.785 | 1.545 | **13.4%** |
| `command_audits.output_text` | 1.514 | 1.276 | **15.7%** |

Content de-duplication reclaims **0.24 GiB**, not 2.10. Command output is not repetitive the way transcripts are (94.9% duplicate — that is #1682's corpus, not this one).

## A storage cap was considered and rejected (2026-08-08)

An 8 KiB head+tail cap on `output_text` was briefly adopted, then dropped. Measured per-row zstd through the codec this repository already ships (`infrastructure/sqlite/payload_codec.go`, `klauspost/compress/zstd`) compresses command output **2.57x**, taking `output_text` from 1.514 to 0.589 GiB — more than a 32 KiB cap would save, and **without discarding anything**. Both #1620 gates close with compression and no cap.

So the storage cap is out of scope here. Compression activation is #1618; the body move that applies it to existing rows is #1685. The **display** cap in `application/types/command_truncation.go` is unaffected and stays.

Shorter retention is also not an alternative lever: 87% of body bytes on this store are younger than 30 days, so shrinking the window does not move either gate.

## Framing correction (2026-08-08)

The original title and body called `command_audits` the "attested" record and the outputs "non-attested". **v0.34 has no attestation** — nothing signs, verifies, or otherwise attests an audit row, so no code and no test can evaluate that distinction. It is the same reason `compressed` was dropped from #1676's discard rule: a predicate the store cannot answer is not a rule.

What is actually true is narrower and sufficient: `command_audits` is the **retained execution record** and outputs are **discardable payload**. The discard predicate is the one #1676 already uses — a range is discardable once it has been **folded** into a refinement — not an attestation property.

## Scope

- `command_audits` keeps the execution identity: `command_text`, `input_text`, result codes. It is retained.
- **Stop persisting the raw `command_executed` payload in `events.body`** once `command_audits` holds the parsed form. This is the 1.785 GiB and it is the whole of this issue.
- Output remains discardable under **the same folded predicate as transcript bodies** (#1676). One rule for all discardable payload, not a second parallel one.

Explicitly **not** in scope:

- A storage-side byte cap on `output_text` (see above — compression supersedes it).
- A new `command_outputs(event_id PK, codec, payload)` side table. It adds a table and a join to model something #1685's `event_bodies` already models, and the measured duplicate share does not justify a separate codec-bearing store.

## Open question to settle first

Whether any reader still needs the raw `command_executed` body once `command_audits` holds the parsed form. Known readers to check: `presentation/cli/read_fields.go:205` (colours rows by `exit_code`, sourced from the audit), `application/usecase/replay_usecase_impl.go:269` (extracts a leading token for hotspot clustering), `application/sensitivepath/classify.go:226` (`ClassifyCommandBody`). If any of them parse the raw body, they move to the audit columns before the raw copy is dropped.

## Done when

- The raw `command_executed` body is no longer persisted, and every reader listed above sources its data from `command_audits`
- The audit trail still answers "what ran, when, by whom" with outputs discarded
- Output discard uses the #1676 folded predicate, with no separate attestation-shaped condition anywhere in the code or config
- Measured post-`store compact`: `events.body` for `kind='command_executed'` is 0

Gate arithmetic and attribution: #1620

Parent: #1606

duck8823:
Closes #1675

`command_executed` events persisted their payload twice: once as the composed `events.body` envelope (`command` + `INPUT:` + `OUTPUT:`) and once as the parsed columns of `command_audits`. Measured on the live store the two sides are the same content at ratio 1.00 - **1.785 GiB, 45% of the #1620 store gate**.

`command_audits` is the retained execution record. The write path now stores an empty body for `command_executed`, and migration 048 clears the envelope on rows that already exist.

## The issue's "open question to settle first"

> Whether any reader still needs the raw `command_executed` body once `command_audits` holds the parsed form.

Settled by reading the code, not assumed:

| reader | outcome |
|---|---|
| `read_fields.go` | already sourced the audit via `event.Show`. Unchanged. |
| `show.go` | already composed its display string from the audit columns. Unchanged. |
| `replay_usecase_impl.go` failure hotspots | moved to the audit |
| `list.go --sensitive`, `doctor_sensitive_access.go` | moved to the audit |
| MCP `convertEventsInternal` | moved to the audit |

`ClassifyCommandBody` and `splitCommandBody` are deleted. They existed only to re-split a body that had been composed from the same three fields `Classify` already accepts.

**Search is unaffected**, and this was checked rather than assumed: both index families already index the audit columns independently of `events.body` - the legacy family through the `command_audits` triggers in migration 032, the bounded projection through `hydrateAuditPayload` at `search_projection_rebuild.go:380-398`, and `search_events.sql:31-33` matches `command_text` / `input_text` / `output_text` directly. Dropping the body removes a duplicate copy from the index *input*, not a searchable field. Pinned by `TestDatasource_Search_FindsTermOnlyInCommandAuditOutput`.

## Second commit: the payload columns are codec-managed

The first cut read `ca.command_text` / `input_text` / `output_text` straight off the listing JOIN. Those are **physical, codec-managed columns**. `hydrateAuditPayload` is the only correct reader - it decodes non-identity codecs, verifies the stored SHA-256, and refuses payloads over `maxDecodedPayloadBytes`.

Reading them raw would have broken as soon as **#1618** (an open sibling in this same sprint) activates compressed writes:

- `traceary list` / `search` and the MCP tools would render zstd frames;
- worse, `list --sensitive` and `doctor` would stop matching sensitive command access **with no error anywhere** - a detection path failing open.

`bundle_datasource.go:661` already carries `command_codec` through bundle import, so that is a second route for encoded payloads to reach these columns.

Now the JOIN carries only fixed-size metadata (`command_name`, wrapper, exit code, failure reason, truncation flags, original bytes), and payloads are decoded on demand through `HydrateCommandAudits`, with each caller naming the fields it needs - command-only for the display paths, full payload for the two sensitive-path classifiers. Most listed rows are not `command_executed` at all, so decoding lazily is both correct and cheaper than the eager join it replaced.

Pinned by `TestListJoinDoesNotSurfaceEncodedCommandPayloads`, which stores a genuinely zstd-encoded audit and asserts the join leaks nothing, full hydration restores the sensitive match, and command-only hydration yields the plaintext command line.

## `traceary list` keeps showing the command line

`eventBodyForDisplay` briefly fell back to `command_name`, so a row that read `go test ./...` started reading `go`. That is a plausible-looking wrong value, and nothing downstream - including the public `message` field of `list --json` and `read --json` - can tell it apart from a real command line. The fallback is gone; the display paths hydrate the real command line instead.

Verified end to end on a fixture store:

```
$ traceary audit --command "go test ./... --verbose-flag" --output "PASS ok github.com/example/pkg 1.2s" ...
$ traceary list
02:48:04  command_executed  agent=codex  sess=s-e2e  ws=traceary  go test ./... --verbose-flag

$ traceary list --json     # .[0].message
"go test ./... --verbose-flag"

$ traceary search "github.com/example/pkg"      # a term only in the output
02:48:04  command_executed  agent=codex  sess=s-e2e  ws=traceary  go test ./... --verbose-flag

$ sqlite3 t.db "SELECT kind, length(CAST(body AS BLOB)) FROM events;"
command_executed|0
```

## Migration 048

`data_dependent_offline` - cost scales with historical audit volume, like 000035 and 000045 - not `constant_in_place`.

Stopping the duplicate write only helps new rows; the 1.785 GiB this issue claims is history, so the migration is where the bytes actually come from. `TestMigration048_ClearsHistoricalCommandExecutedBodies` seeds a pre-48 store, reopens it with the full migration set, and asserts the `command_executed` body goes to zero (measured 67 -> 0) while a `transcript` row in the same store is untouched.

Whether the file shrinks or only frees pages is a `VACUUM` question, deliberately left to `store compact` - see the docs change.

## Filed rather than widened

- **#1730** - `HydrateCommandAudits` is per-event, per-field, and re-probes the schema on each call. Batching it and hoisting the probe are worth doing, but not by reverting to the raw join this PR just removed.

## Explicitly out of scope, per the issue

No byte cap on `output_text` (withdrawn in favour of compression, #1618/#1685), no `command_outputs` side table, no content de-duplication (measured at 13.4%, not the lever), and the display cap in `application/types/command_truncation.go` stays.

---

`go build ./...` clean, `go tool golangci-lint run` 0 issues, `go test ./...` exit 0.


duck8823:
## Review round 1 — gpt-5.6-sol (read-only), reviewed by Claude Opus 5

**Five findings: 1 CRITICAL, 4 HIGH. All five reproduced before being acted on — none taken on assertion — and all five fixed in `46f542df`.**

This was a good review. The first two commits were verified against the schema and the search paths, and still shipped a data-loss bug and a blank default MCP surface.

### 🔴 CRITICAL — migration 048 destroyed data

> 000048 clears every `command_executed` body without checking that a `command_audits` row exists. `traceary log --kind command_executed` reaches `Log`, not `SaveWithAudit`.

Reproduced on a fixture store. With one audited command event and one `traceary log --kind command_executed` event side by side, the audited row already has an empty body and its payload in `command_audits`, while the logged row has a 21-byte body and **no audit row at all**.

The migration would have erased that second row permanently on upgrade. It now carries `AND EXISTS (SELECT 1 FROM command_audits a WHERE a.event_id = events.id)`, so it clears only genuinely duplicated bodies. `Log` still writes the body — for that path it is the record, not a duplicate.

`TestMigration048_ClearsHistoricalCommandExecutedBodies` now seeds both shapes and asserts the audited body goes to zero while the log-only body survives.

### 🔴 HIGH — the default MCP projection returned empty bodies

> `list_events` defaults to `EventProjectionBounded`; the bounded path reads `events.body` and nothing else. Only `projection=full` was wired to hydrate.

Confirmed at `server.go:802` and `event_bounded_datasource.go:270` → `sql/select_bounded_event_bodies.sql`. This was the worst of the five in practice: **every agent using the default MCP surface saw a blank body for every command event.**

`hydrateBoundedEvents` now falls back to the decoded command line — in the infrastructure layer, so the codec, checksum and `maxDecodedPayloadBytes` guards still apply — and recomputes the rune budget so the truncation indicator does not lie.

Verified by driving the server over stdio: the `command_executed` entry comes back with `body` set to the full command line and its `exit_code`, where before the fix `body` was empty.

### 🔴 HIGH — top / cockpit / TUI rendered blank rows

Reproduced with `traceary top --snapshot`: under `RECENT COMMANDS` the `traceary log` row still showed its text while the audited command row rendered as an empty string.

My earlier instruction to leave `cockpit.go` alone was wrong — it does render bodies. All three sites hydrate command-only now, and the same command afterwards shows `go test ./... --verbose-flag` on the audited row.

### 🔴 HIGH — replay wrote blank command events

`replay.go:250` renders `ExtractPlainBody(event.Body())` directly; the earlier commit had repaired only failure-hotspot clustering. Fixed, and verified against generated markdown — the command line appears in the event stream.

### 🔴 HIGH — three SQL fallbacks read the physical codec-managed column

`list_sessions.sql`, `list_session_tree.sql` and `session_lineage.sql` fell back to `latest_audit.command_text`, which under a non-identity codec puts **encoded bytes** into the session row and bypasses the checksum and size limit — precisely the defect `eb1ab214` removed from the Go path, reintroduced in SQL.

A per-row decode would be an N+1 on a listing query, so these fall back only when `COALESCE(command_codec, 'identity') = 'identity'` and render empty otherwise. Honest rather than wrong.

---

`go build ./...` clean, `go tool golangci-lint run` 0 issues, `go test ./...` exit 0.

PR #1732, opened 2026-08-08
feat: build the first bounded search projection generation on store open

### LINKED ISSUE #1680: v0.34.0-40: execute the tiered search cutover

duck8823:
Closes #1680

## Why this is not a re-run

Migrations 035–047 have never shipped — v0.33.1 tops out at 034 — so no store in the field has ever built a projection generation. Every store today answers search from the legacy migration-032 index, and the authoritative-projection switch #1717 added has never once been reachable. #1718 deletes that legacy index, so a generation has to exist before it lands. This PR is where the first cutover actually happens, and it happens on upgrade rather than on a command nobody will run.

## How the first generation gets built

The same way event search backfill is driven: `Initialize` performs one bounded, durable unit of work per store open, resumable across opens, no operator command and no new subcommand. The batch size is the existing `eventSearchBackfillBatchSize` value (128), so the precedent this repository already accepted for the same corpus carries over unchanged. Legacy search stays authoritative until the generation reaches `complete`, so the read path needs no change — `searchProjectionReadReady` already makes the switch automatic.

An empty store stays idle rather than starting a generation, and an operator-started rebuild with a different budget is not hijacked.

## What the verification gate actually asks

Reclaiming the previous generation is gated on a real session-tier query against the generation under construction, using the same match contract production search uses — keyword equality on the keyword table, or summary `LIKE` — with the same `ts_norm` ordering. Not a row count.

The gate asks whether the tier can answer *for a session it holds a summary for*, scoped to that session. Two things drove that shape:

- Sessions in one workspace share vocabulary, so a term drawn from one summary routinely matches others. Requiring the probed session to come back as the top-ranked hit fails on any store whose summaries share a word, which is essentially all of them.
- A failed verification marks the generation failed, and a failed generation restarts on the next open. So a spurious failure is not a warning — it is an unbounded rebuild loop that never reaches `complete`, on a code path whose entire purpose is reaching `complete` before #1718 removes the alternative.

`TestSearchProjectionVerifySessionTier_PassesWhenSeveralSessionsShareATerm` pins this. It fails on the pre-fix gate with `session tier verification returned unexpected session sess-zzz`.

## Schema completeness is decided from the schema

The implementation originally decided "the projection schema is incomplete, skip" by matching `no such table` / `no such column` / `no such module` against the text of a failure, in the catch-up path and in three fallbacks inside the rebuild path.

That test also matches a genuine SQL defect in the projection. A typo in the primary statement would fall through to a secondary statement or to a silent skip, the generation would never build, and after #1718 search would return nothing with no error anywhere — the same fail-silent class this issue exists to close.

Replaced with an explicit probe of the objects migrations 038–042 create, plus migration 049's column, naming the first missing object so a skip is diagnosable. The three rebuild-path fallbacks turned out to be dead — removing them leaves the full suite green.

## Interruption and reclaim

An interrupted rebuild resumes on the next open from its checkpoint, and only one generation is live while it does. `TestSearchProjectionCatchUp_InterruptedMidRebuildResumesWithoutTwoLiveGenerations` drives a rebuild that cannot finish in one Resume, asserts exactly one generation in `rebuilding`/`complete` mid-flight, then resumes to completion with one complete and zero rebuilding.

## Bytes

Migration 049 records `cutover_index_family` alongside before/after bytes, always naming `bounded_search_projection`. The measurement covers `search_projection_*` + `literal_search_*` only. The legacy family's 16.15 GiB is #1718's number and is deliberately not reported here, so nothing in `status` output can be read as a store-wide reduction. On a fixture store the two figures are close, since the fixture's corpus is tiny and the schema is already present; the value of the field is that the number is attributed to a family rather than to the store.

## Carried to #1718, not fixed here

After #1718 removes the legacy family, both remaining fallback routes in `selectEventSearchCandidateIDs` — "no complete generation" and "the tail is too long to answer within budget" — currently land on `selectLegacyEventSearchCandidateIDs`. This PR makes the first case rare rather than universal, but it does not make it impossible: a store that upgrades and is then queried before the generation finishes is in exactly that state. #1718 has to degrade rather than break there. Noted on that issue rather than pre-empted here.

## Verification

```
go build ./...              clean
go tool golangci-lint run   0 issues
go test ./...               exit 0
```

### DIFF
--- a/application/types/search_projection.go
+++ b/application/types/search_projection.go
@@ -222,4 +222,30 @@ type SearchProjectionStatus struct {
 	FingerprintLogicalBytes int64            `json:"fingerprint_logical_bytes"`
 	LifecycleState          string           `json:"lifecycle_state"`
 	AbandonedAt             string           `json:"abandoned_at,omitempty"`
+	// CutoverIndexFamily names which physical family CutoverFamilyBytes*
+	// measure. Always "bounded_search_projection" when set — never the
+	// legacy migration-032 event_search_* family (that is #1718).
+	CutoverIndexFamily       string `json:"cutover_index_family,omitempty"`
+	CutoverFamilyBytesBefore int64  `json:"cutover_family_bytes_before,omitempty"`
+	CutoverFamilyBytesAfter  int64  `json:"cutover_family_bytes_after,omitempty"`
+}
+
+// SearchProjectionCatchUpResult is one bounded unit of automatic generation
+// work performed during store initialization. It mirrors the event-search
+// backfill shape: a single open does a bounded amount of work and resumes
+// later without operator action.
+type SearchProjectionCatchUpResult struct {
+	Action                   string `json:"action"`
+	State                    string `json:"state"`
+	Phase                    string `json:"phase"`
+	GenerationID             string `json:"generation_id,omitempty"`
+	Completed                bool   `json:"completed"`
+	Batches                  int    `json:"batches"`
+	Selected                 int    `json:"selected"`
+	Written                  int    `json:"written"`
+	SkippedReason            string `json:"skipped_reason,omitempty"`
+	CutoverIndexFamily       string `json:"cutover_index_family,omitempty"`
+	CutoverFamilyBytesBefore int64  `json:"cutover_family_bytes_before,omitempty"`
+	CutoverFamilyBytesAfter  int64  `json:"cutover_family_bytes_after,omitempty"`
+	SessionTierVerified      bool   `json:"session_tier_verified,omitempty"`
 }
--- a/application/types/search_projection.go
+++ b/application/types/search_projection.go
@@ -232,10 +232,15 @@ type SearchProjectionStatus struct {
 	CutoverIndexFamily       string `json:"cutover_index_family,omitempty"`
 	CutoverFamilyBytesBefore int64  `json:"cutover_family_bytes_before,omitempty"`
 	CutoverFamilyBytesAfter  int64  `json:"cutover_family_bytes_after,omitempty"`
-	// CutoverFamilyEvidence states whether the byte figures above were actually
-	// measured. Without it a family that could not be measured is reported as
-	// zero bytes, which reads identically to a genuinely empty family.
-	CutoverFamilyEvidence CapacityEvidence `json:"cutover_family_evidence"`
+	// CutoverBeforeEvidence and CutoverAfterEvidence state whether each byte
+	// figure above was actually measured. Without them a family that could not
+	// be measured reports zero bytes, which reads identically to a genuinely
+	// empty family. They are separate because the two walks happen at different
+	// times against families of different sizes: one can succeed while the
+	// other times out. An empty Status means no measurement has been attempted
+	// yet.
+	CutoverBeforeEvidence CapacityEvidence `json:"cutover_before_evidence"`
+	CutoverAfterEvidence  CapacityEvidence `json:"cutover_after_evidence"`
 }
 
 // SearchProjectionCatchUpResult is one bounded unit of automatic generation
@@ -255,9 +260,11 @@ type SearchProjectionCatchUpResult struct {
 	CutoverIndexFamily       string `json:"cutover_index_family,omitempty"`
 	CutoverFamilyBytesBefore int64  `json:"cutover_family_bytes_before,omitempty"`
 	CutoverFamilyBytesAfter  int64  `json:"cutover_family_bytes_after,omitempty"`
-	// CutoverFamilyEvidence carries the same measured/unavailable distinction as
-	// SearchProjectionStatus so a zero in the byte fields above is never read as
-	// an empty family when it only means the walk did not run.
-	CutoverFamilyEvidence CapacityEvidence `json:"cutover_family_evidence"`
+	// CutoverBeforeEvidence and CutoverAfterEvidence carry the same
+	// measured/unavailable distinction as SearchProjectionStatus so a zero in
+	// the byte fields above is never read as an empty family when it only means
+	// the walk did not run.
+	CutoverBeforeEvidence CapacityEvidence `json:"cutover_before_evidence"`
+	CutoverAfterEvidence  CapacityEvidence `json:"cutover_after_evidence"`
 	SessionTierVerified   bool             `json:"session_tier_verified,omitempty"`
 }
--- a/application/types/search_projection.go
+++ b/application/types/search_projection.go
@@ -222,12 +222,20 @@ type SearchProjectionStatus struct {
 	FingerprintLogicalBytes int64            `json:"fingerprint_logical_bytes"`
 	LifecycleState          string           `json:"lifecycle_state"`
 	AbandonedAt             string           `json:"abandoned_at,omitempty"`
+	// FailureClass names why the last generation failed. It is what decides
+	// whether automatic catch-up may start a replacement: a deterministic class
+	// would fail the same way on every open.
+	FailureClass string `json:"failure_class,omitempty"`
 	// CutoverIndexFamily names which physical family CutoverFamilyBytes*
 	// measure. Always "bounded_search_projection" when set — never the
 	// legacy migration-032 event_search_* family (that is #1718).
 	CutoverIndexFamily       string `json:"cutover_index_family,omitempty"`
 	CutoverFamilyBytesBefore int64  `json:"cutover_family_bytes_before,omitempty"`
 	CutoverFamilyBytesAfter  int64  `json:"cutover_family_bytes_after,omitempty"`
+	// CutoverFamilyEvidence states whether the byte figures above were actually
+	// measured. Without it a family that could not be measured is reported as
+	// zero bytes, which reads identically to a genuinely empty family.
+	CutoverFamilyEvidence CapacityEvidence `json:"cutover_family_evidence"`
 }
 
 // SearchProjectionCatchUpResult is one bounded unit of automatic generation
@@ -247,5 +255,9 @@ type SearchProjectionCatchUpResult struct {
 	CutoverIndexFamily       string `json:"cutover_index_family,omitempty"`
 	CutoverFamilyBytesBefore int64  `json:"cutover_family_bytes_before,omitempty"`
 	CutoverFamilyBytesAfter  int64  `json:"cutover_family_bytes_after,omitempty"`
-	SessionTierVerified      bool   `json:"session_tier_verified,omitempty"`
+	// CutoverFamilyEvidence carries the same measured/unavailable distinction as
+	// SearchProjectionStatus so a zero in the byte fields above is never read as
+	// an empty family when it only means the walk did not run.
+	CutoverFamilyEvidence CapacityEvidence `json:"cutover_family_evidence"`
+	SessionTierVerified   bool             `json:"session_tier_verified,omitempty"`
 }
--- a/application/usecase/search_projection_usecase.go
+++ b/application/usecase/search_projection_usecase.go
@@ -260,7 +260,8 @@ func (u *SearchProjectionUsecase) CatchUp(ctx context.Context, b apptypes.Search
 	out.CutoverIndexFamily = status.CutoverIndexFamily
 	out.CutoverFamilyBytesBefore = status.CutoverFamilyBytesBefore
 	out.CutoverFamilyBytesAfter = status.CutoverFamilyBytesAfter
-	out.CutoverFamilyEvidence = status.CutoverFamilyEvidence
+	out.CutoverBeforeEvidence = status.CutoverBeforeEvidence
+	out.CutoverAfterEvidence = status.CutoverAfterEvidence
 	if status.State == "complete" {
 		out.Action = "already_complete"
 		out.Completed = true
@@ -281,8 +282,11 @@ func (u *SearchProjectionUsecase) CatchUp(ctx context.Context, b apptypes.Search
 	// is ever introduced, this is where it gets its exception.
 	if status.State == "failed" {
 		out.Action = "skipped"
+		// Naming the recovery command matters: neither resume nor abort clears
+		// this state. Resume rejects a failed generation, and abort leaves the
+		// row failed with class abandoned. Only an explicit start replaces it.
 		out.SkippedReason = "parked after generation failure " + failureClassOrUnknown(status.FailureClass) +
-			"; resume or abort explicitly to unblock automatic progress"
+			"; run 'traceary store search-projection start' to replace the generation"
 		return out, nil
 	}
 	switch {
@@ -335,7 +339,8 @@ func (u *SearchProjectionUsecase) CatchUp(ctx context.Context, b apptypes.Search
 	out.CutoverIndexFamily = statusAfter.CutoverIndexFamily
 	out.CutoverFamilyBytesBefore = statusAfter.CutoverFamilyBytesBefore
 	out.CutoverFamilyBytesAfter = statusAfter.CutoverFamilyBytesAfter
-	out.CutoverFamilyEvidence = statusAfter.CutoverFamilyEvidence
+	out.CutoverBeforeEvidence = statusAfter.CutoverBeforeEvidence
+	out.CutoverAfterEvidence = statusAfter.CutoverAfterEvidence
 	if progress.Completed {
 		out.SessionTierVerified = true
 		out.Completed = statusAfter.Completed
--- a/application/usecase/search_projection_usecase.go
+++ b/application/usecase/search_projection_usecase.go
@@ -32,6 +32,12 @@ type SearchProjectionAbandonStore interface {
 	AbandonSearchProjection(context.Context, time.Time) (apptypes.SearchProjectionAbandonResult, error)
 }
 
+// SearchProjectionVerifyStore gates old-generation reclaim on a real
+// session-tier query against the generation under construction.
+type SearchProjectionVerifyStore interface {
+	VerifySearchProjectionSessionTier(context.Context, string) error
+}
+
 // NewSearchProjectionUsecase constructs the projection workflow.
 func NewSearchProjectionUsecase(store SearchProjectionStore) *SearchProjectionUsecase {
 	return &SearchProjectionUsecase{store: store}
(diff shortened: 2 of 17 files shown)


duck8823:
## Claude Opus 5 review — changes required

Two defects found and fixed before this PR was opened, and one found after by Sol that I have now reproduced and which blocks merge.

### Fixed before opening

**Session-tier verification failed whenever two sessions shared a word.** The gate picked its candidate with `ORDER BY sum.session_id LIMIT 1` and then asked whether that session came back from a query ordered `ORDER BY ts_norm(s.started_at) DESC, s.session_id DESC LIMIT 1`. Two different orderings, so the two agree only when exactly one session matches. The term is drawn from the candidate's own summary and is the first token of three runes or more — sessions in one workspace share vocabulary, so multiple matches is the normal case.

The consequence is not a warning. A failed verification calls `markFailed`, a failed generation restarts on the next open, and the restarted generation fails verification again — an unbounded rebuild loop on the one code path whose entire purpose is reaching `complete`. Now scoped with `sum.session_id = ?`, with `TestSearchProjectionVerifySessionTier_PassesWhenSeveralSessionsShareATerm` pinning it. That test fails on the pre-fix gate with `session tier verification returned unexpected session sess-zzz`.

**Schema-incompleteness was decided by matching error text** — `no such table` / `no such column` / `no such module` — in the catch-up path and in three fallbacks inside the rebuild path. That test also matches a genuine SQL defect: a typo in the primary statement would fall through to a secondary statement or to a silent skip, the generation would never build, and after #1718 search would return nothing with no error anywhere. Replaced with an explicit probe of the objects migrations 038–042 create plus migration 049's column. The three rebuild-path fallbacks were dead — the suite is green without them.

### Blocking: the automatic generation makes zero progress on the store it exists for

Sol raised this; I reproduced it. Fixture: a pre-038 store with 5,000 historical events, migrated to head, then alternating {insert one event, open the store} — which is exactly how a hook behaves, since it writes an event immediately after `Initialize` returns.

Over 20 opens, `search_projection_source_sequence` grew only by the live inserts. The inventory cursor froze at the first batch boundary and never moved. A `search_projection_generation_lifecycle` row accumulated every two opens.

Instrumented, the cycle is:

- **Open N** — `SelectInventory` sees the live revision one ahead of the generation's anchor, marks the generation `drifted`, returns. No work done.
- **Open N+1** — `CatchUp` starts a new generation, re-anchors, falls through to one `Resume`. That Resume does zero inventory rows.
- **Open N+2** — inventory advances by 128, but only if nothing wrote in between.

So inventory advances only when two consecutive store opens see no intervening event write. Removing the writes confirms recovery: one write followed by six quiet opens resumes advancing 128 per open from the third open onward, and the cursor is preserved across generations — nothing is lost, it simply never advances. On a store where a hook writes on essentially every open, progress is exactly zero.

The mechanism is the insert trigger in migration 038. It bumps `source_revision` unconditionally while `requires_inventory=1`, and migration 038 sets that flag for any store that already had events — that is, every store this PR targets. The conservatism looks deliberate: the inventory walk is `ORDER BY e.id` with an `e.id > cursor` cursor, and event ids are random hex, so a new event can sort below the cursor and be skipped.

But the trigger's own first statement writes the new event's `search_projection_source_sequence` row unconditionally, before the revision bump. That row is the only thing inventory would have written for it. So an insert during inventory cannot invalidate inventory work already done and cannot create work inventory needs to do — the drift signal for inserts is guarding an invariant the trigger already guarantees. Update and delete are genuinely different and keep the conservative behaviour.

Sol is reviewing that reasoning now; the fix will follow.

### Also to fix

**A config-hash skip is completely silent.** `logSearchProjectionCatchUp` returns early for `skipped`, so if an operator starts a rebuild with a non-default budget and abandons it, every later open reports success while doing nothing, and the store never reaches `complete` with no operator-visible signal anywhere. Given this PR's whole premise is that the cutover happens without anyone watching, a silent permanent skip is the wrong default. It needs a level an operator would actually see.

### Verified

Sol confirmed the connection and locking behaviour is sound: the outer connection holds no active transaction, a second `*sql.DB` in WAL mode is fine, selection runs outside a write transaction, the write transaction is capped at 250ms, and a catch-up error does not fail `Initialize`.

Steady-state cost measured on a 20,000-event fixture: about 90ms per open, migrating open about 120ms. That is acceptable. Inventory consumes exactly 128 rows per open as intended.

CI green on `dec40cb2`.

duck8823:
## Sol review of the proposed fix — one correction to my diagnosis, and one thing my fix missed

### Correction

I wrote above that "the cursor is preserved across generations — nothing is lost, it simply never advances." That is wrong about the mechanism. `StartGeneration` resets the inventory cursor to empty (`infrastructure/sqlite/search_projection_rebuild.go:79`), so each restart re-walks the same already-registered prefix from the head. `INSERT OR IGNORE` makes that idempotent, which is why the cursor appears to sit still at the same value. The observable outcome — never getting past the first batch — is unchanged.

### What my proposed fix missed

Fixing only `search_projection_events_insert` would not have worked. An ordinary command event writes an `events` row and a `command_audits` row in the same transaction (`infrastructure/sqlite/event_delivery_store.go:176-207`), and `search_projection_audits_insert` carries the identical `requires_inventory=1` disjunct (migration 038 line 13). Command hooks are the dominant writer on this store, so the stall would have survived the fix entirely. Both insert triggers need the change.

### What held up

The safety argument checks out on both sides of the watermark handoff:

- An insert committing before the inventory handoff lands under the final `high_water` — the handoff sets it from the then-current `MAX(sequence)` across main state, lifecycle and literal state (`search_projection_rebuild.go:279-303`), and the source phase processes everything at or below it (`:363-421`). The generation picks it up.
- An insert after the handoff gets `sequence > high_water`, which is exactly what #1717's tail merge keys on (`event_search_query.go:186-217`).

Update and delete triggers keep the conservative behaviour — they can touch historical rows the walk has not reached, with no membership information available while `requires_inventory=1`.

The alternatives were rejected on their merits: ordering the inventory walk by `events.rowid` widens the design surface for no gain here; doing inventory work in the same open as a restart only reduces the requirement from two quiet opens to one, and one is still never; and re-anchoring on insert-only drift is not implementable against a single revision counter that cannot distinguish inserts from mutations.

Lifecycle row accumulation stops on its own once a generation completes, since `CatchUp` does not start a new one from `complete`. Rows already accumulated are not cleaned up, and a bulk DELETE inside a migration would be the wrong instrument — left alone deliberately.

### Filed separately

**#1733** — the `tiered` search authority errors permanently after the first event recorded post-completion. `searchTieredMetadataTx` gates on `literalHigh != sourceHigh` (`event_search_authority.go:143`), `sourceHigh` is `MAX(sequence)` and grows on every insert, `literalHigh` is frozen at generation completion, and `CatchUp` returns `already_complete` so nothing re-syncs them. Pre-existing, not introduced here, and it does not fire while the authority is `legacy` because the tail merge covers that case. It blocks #1718, which is what switches the authority.

Verified against the code before filing.

### Confirmed sound

Connection and locking behaviour: the outer connection holds no active transaction, a second `*sql.DB` in WAL mode is fine, selection runs outside a write transaction, the write transaction is capped at 250ms, and a catch-up error does not fail `Initialize`.

Fix in progress on this branch.

duck8823:
## Fixed in `efcf688f` — verified in both directions

Migration 050 drops the `requires_inventory=1` disjunct from both insert triggers, keeping `sequence <= high_water` as the only insert-time drift signal. Update and delete triggers are untouched.

### The regression tests fail without it

`TestSearchProjectionCatchUp_AlternatingWritesConvergeInventory` and `TestSearchProjectionCatchUp_AlternatingCommandEventsConverge` build a pre-038 store with historical events, migrate to head, then alternate {write an event, open the store}. With migration 050 removed, both fail with exactly the signature I measured by hand:

```
projection did not complete under alternating writes:
  inventory(cursor="historical-00000127" started=1 seq=208 requires=1 lifecycle=41 state=rebuilding phase=source)
```

Cursor pinned at the first batch boundary after 40 opens, 41 lifecycle rows — one per restart. With the migration, both converge to `complete` and lifecycle stays bounded.

### The audits trigger is load-bearing, not defensive

I applied each half of the migration alone to check. With only `search_projection_events_insert` fixed:

```
--- FAIL: TestSearchProjectionCatchUp_AlternatingCommandEventsConverge
ok      TestSearchProjectionCatchUp_AlternatingWritesConvergeInventory
```

Plain writes converge, command events still stall. Since a command event writes `events` and `command_audits` in one transaction and command hooks are the dominant writer on this store, the fix I originally proposed would have shipped and left the live store exactly as stalled as before. That correction came from Sol's review.

### Other coverage added

- Update and delete during inventory still drift — pinning the behaviour deliberately left alone.
- `search_projection_generation_lifecycle` stays bounded rather than growing per open.
- The config-hash skip now logs at warning level with its reason and says what unblocks it, instead of returning early and discarding the reason.
- Docs updated in both languages: inserts no longer invalidate a generation, updates and deletes still do, and skips are visible.

### Verification

```
go build ./...              clean
go tool golangci-lint run   0 issues
go test ./...               exit 0
```

Migration 050 SHA in the prepared-migration catalog matches the file, classified `MigrationConstantInPlace` — it drops and recreates two triggers with no data scan or rewrite.

Ready for a second Sol pass.

duck8823:
## Second Sol pass — two more HIGHs, both confirmed, both the same shape

Sol cleared migration 050 itself: no difference from the migration-038 originals beyond the intended disjunct, statement order and the unconditional `INSERT OR IGNORE` preserved, `WHEN` versus `WHERE` placement preserved, `new.event_id` preserved. No clobbering or shadowing of the triggers created by 038, 039 or 041, and no dependence on trigger creation order — the events-before-audit ordering that matters is guaranteed by the statement order in `SaveWithAudit`, not by trigger order. `MigrationConstantInPlace` confirmed correct: `DROP TRIGGER` / `CREATE TRIGGER` touch the schema table only.

The two new findings are both cases where the generation can never reach `complete` and nothing says so — the same class as the inventory starvation.

### The completion transaction measures the projection family inside a 250ms lock

`applyProjectionBatch` calls `measureSearchProjectionFamilyBytes` inside the completion transaction, before publishing `active_generation_id` (`search_projection_rebuild.go:685-690`), under a 250ms context.

Sol's stated mechanism was that the `dbstat` query scans the whole database. That part is wrong, and I checked it rather than passing it on: SQLite materialises the `name IN (subquery)` and probes per name, so on a 67,521-page store with a tiny projection family the query takes 216µs. Whole-database scanning is not what happens.

The finding survives anyway, in the family's own size:

| projection family | shipped query, cold |
|---|---|
| a few pages, 67,521-page store | 216µs |
| ~880MiB, 247,571-page store | **1.44s** |

1.44s against a 250ms budget — 5.8× over. A full generation over ~550,000 events lands the family well into that range. The deadline fires, the query errors, the transaction rolls back, `active_generation_id` is never published, and the next open repeats the same final batch forever.

`StartGeneration` has the same exposure at `:66-69`. It is cheap today only because the family is empty before the first generation.

The figure is diagnostic — nothing reads it to decide anything. It should not be able to prevent a generation completing, and when it cannot be produced that needs to be distinguishable from a genuinely zero-byte family rather than recorded as `0`.

### A deterministic failure restarts the generation forever

`CatchUp` treats `failed` as auto-startable (`search_projection_usecase.go:273-294`). Not all failures are transient. The automatic budget allows 8MiB stored/decoded/write while the payload contract allows up to 16MiB decoded, so a perfectly valid row above the budget yields `SearchProjectionOversizeError`, `Resume` marks the generation failed, the next open starts a fresh one, and it fails on the same row again — with a lifecycle row per cycle.

`session_tier_unverified` is deterministic in the same way. Notably that is the same loop the verification fix earlier on this branch removed the *trigger* for; the loop structure itself is still there for genuine failures.

The split should be deterministic-by-default: auto-restart only classes justifiable as transient, park everything else with a warning naming the class. A wrongly-retried deterministic failure is an infinite loop; a wrongly-parked transient one is a warning and a stalled cutover an operator can see.

### On the 128-row batch and the long incomplete window

Sol put the live store at roughly 8,600 opens minimum to complete — about 4,300 for inventory and as many again for the source phase. It did not argue the batch size is wrong, and neither do I: it matches the `eventSearchBackfillBatchSize` precedent this repository already accepted for the same corpus. Its point is narrower and fair — that precedent justifies a bounded amount of *body* work per open, and does not extend to an unbounded page measurement or to restarting a deterministic failure. Both of those are being fixed rather than budgeted around.

Throughout that window the legacy family stays authoritative, so ordinary search correctness is unaffected. #1718 already carries the explicit requirement to degrade rather than break when no generation is complete.

Fixes in progress.

duck8823:
## Third Sol pass — four HIGHs, all confirmed, all fixed in `480c425a`

I reproduced each before acting rather than taking the review at face value. All four were real.

### 1. The evidence write could land on the wrong generation

`recordSearchProjectionCutoverEvidence` scoped its UPDATE `WHERE singleton=1 AND active_generation_id=?`. Between the completion commit and that write, another process can `Start` a replacement: `generation_id` moves on, `active_generation_id` still points at the generation that just completed. The stale write then matched and stamped the completed generation's after-evidence over the replacement's freshly-recorded before-evidence.

Now fenced on `generation_id` as well. `TestSearchProjectionCutoverEvidence_DoesNotOverwriteAReplacementGeneration` fails against the old scope with `status="measured" bytes=102400` on the replacement row.

### 2. The walk was detached from cancellation; the write that stored it was not

`measureSearchProjectionFamilyBytes` used `context.WithoutCancel`, but the `UPDATE` carrying its result still used the batch context. `Resume` gives that context `b.WallTime` — one second on the automatic path — so a walk that took anywhere near the 1.44s I measured for an ~880MiB family would complete and then fail to record. The after figure stays at 0 and is never re-measured.

That is precisely the ambiguity this branch set out to remove, reintroduced one line later. Both the walk and the write now share a single detached context bounded at `measureTimeout + 1s`. `TestSearchProjectionCutoverEvidence_RecordsOnACancelledBatchContext` calls the recorder with an already-cancelled context; against the old code it reports `status=""` and `bytes=0`.

Sol also correctly noted the measurement is synchronous and cannot outlive the caller's `db.Close` — `WithoutCancel` does not leak here.

### 3. One evidence field for two independent measurements

Before is measured at `Start` against a family that is often empty; after is measured post-cutover against a much larger one. They shared `cutover_evidence_status`, so the after-write overwrote the before-status. A before-walk that timed out followed by a successful after-walk reported the whole row as `measured` while `cutover_family_bytes_before` sat at an unmeasured 0.

Migration 051 was rewritten (not yet released, so no 052 needed) to record four columns: `cutover_before_evidence_status`/`_reason` and `cutover_after_evidence_status`/`_reason`. `SearchProjectionStatus` and the catch-up result carry `CutoverBeforeEvidence` and `CutoverAfterEvidence`. Empty status now means "not attempted", which makes the distinction three-way and legible.

### 4. The parked-failure message named commands that do not recover

The skip reason said "resume or abort explicitly to unblock". Neither does: `Resume` rejects anything that is not `rebuilding` or `drifted`/`cleanup` (`search_projection_usecase.go:71-73`), and `Abandon` sets `state='failed', failure_class='abandoned'` (`search_projection_rebuild.go:743-750`), so the next open parks again. Only `start` replaces the generation. The message and both doc pages now say so.

Worth stating plainly: parking makes `abort` a real stop rather than something the next store open silently undoes. That is a behaviour change and it is the intended one — before this branch, `abort` was effectively a no-op against automatic catch-up.

### Verification

`go build ./...`, `go tool golangci-lint run` (0 issues), `go test ./...` all green. Each of the four new tests was run against the pre-fix code and observed to fail.

duck8823:
## Fourth Sol pass — clean

No CRITICAL or HIGH surviving on `480c425a`. Sol independently ran `go test ./...` (green) and pressed on the specific risks I asked about: the detached write racing an abandon or replacement, rewriting migration 051 in place against the prepared-migration catalog / plan digest / publication path, `Start` clearing the after pair on the drifted-cleanup path, NULL scans on the four new columns, and the read-time `Method` derivation.

Four review rounds on this branch, nine findings, all reproduced before fixing:

| Round | Finding | Fixed in |
|---|---|---|
| 1 | Session-tier verification compared two different `ORDER BY`s | `dec40cb2` |
| 1 | Error-text matching as control flow, four sites | `dec40cb2` |
| 2 | Inventory starvation — `requires_inventory=1` disjunct in **both** insert triggers | `efcf688f` (migration 050) |
| 3 | dbstat walk inside the 250ms completion lock rolled back finished generations | `56eccff6` |
| 3 | Deterministic failures auto-restarted forever, one lifecycle row per open | `56eccff6` |
| 4 | Evidence write scoped only on `active_generation_id` clobbered a replacement | `480c425a` |
| 4 | Evidence write inherited the 1s batch context the walk was detached from | `480c425a` |
| 4 | Before/after sharing one evidence field reported unmeasured bytes as measured | `480c425a` |
| 4 | Parked-failure message named `resume`/`abort`, neither of which recovers | `480c425a` |

CI green on `480c425a` (Test, Lint, Docs, Integrations). Marking ready.

PR #1734, opened 2026-08-09
fix: answer tiered search across the post-cutover tail

### LINKED ISSUE #1733: v0.34-52: tiered search authority errors permanently after the first event recorded post-cutover
Found by Sol during review of #1732 (#1680) and verified against the code. Pre-existing — not introduced by #1680 — but it blocks #1718, which is what switches the authority to `tiered`.

## What breaks

Once `search_maintenance_control.authority = 'tiered'`, non-structural search goes straight to `searchTieredMetadataTx` and never reaches #1717's tail merge. That function gates on:

```go
if literalState != "complete" || boundedState != "complete" || literalGeneration == "" ||
   literalGeneration != boundedGeneration || literalHigh != sourceHigh {
    return nil, xerrors.New("tiered search projection is incomplete or stale")
}
```

`infrastructure/sqlite/event_search_authority.go:143`

`sourceHigh` is `SELECT COALESCE(MAX(sequence),0) FROM search_projection_source_sequence`. Migration 038's insert trigger registers a sequence row for **every** new event, unconditionally, outside any WHERE clause:

```sql
INSERT OR IGNORE INTO search_projection_source_sequence(event_id) VALUES(new.id);
```

`schema/sqlite/migrations/000038_add_bounded_search_projections.sql:10`

`literalHigh` is frozen at the value the generation completed with. Nothing updates it afterwards.

So one ordinary event — one hook fire — makes `sourceHigh = literalHigh + 1`, and every subsequent non-structural search returns an error. Not an empty result: an error.

It does not recover. `CatchUp` sees `state = 'complete'` and returns `already_complete` without starting a new generation (`application/usecase/search_projection_usecase.go:261-264`), so nothing ever re-syncs the two watermarks.

## Why this matters for #1718

Under `authority = 'legacy'` the equivalent situation is handled: #1717's tail merge picks up everything with `sequence > high_water` (`infrastructure/sqlite/event_search_query.go:186-217`), falling back to the legacy family past 10,000 candidates. The tiered path has no equivalent — it errors instead of merging.

#1718 removes the legacy writer and switches the authority to tiered (`infrastructure/sqlite/search_maintenance.go:163-173`). The moment it does, the first hook write after cutover breaks search permanently on that store.

## What needs to happen

Either the tiered authority grows post-`high_water` handling equivalent to the #1717 tail merge, or completing a generation stops being a terminal state and a new generation starts once the watermark gap crosses some threshold. The first is closer to what already exists and does not reintroduce continuous rebuilding.

Whichever way it goes, a strict watermark-equality gate is the wrong shape for a store that is written to continuously — that condition is false almost all of the time by construction.

## Acceptance

- Search under `authority = 'tiered'` returns correct results after events are recorded post-completion, not an error.
- A test that completes a generation, records an ordinary event, and searches — asserting results rather than a "stale" error.
- The behaviour is covered for both the metadata path and the full-event path (`searchByPersistedAuthority` and `searchFullByPersistedAuthority` both route to `searchTieredMetadataTx`).

Blocks #1718.

duck8823:
Closes #1733

## Problem

Once `search_maintenance_control.authority='tiered'`, every non-structural search went through `searchTieredMetadataTx` and returned `tiered search projection is incomplete or stale` after the first ordinary write, permanently. `CatchUp` sees `state='complete'` and returns `already_complete`, so it never recovers on its own. This blocks #1718, which flips the authority to tiered and deletes the legacy fallback.

Three independent mechanisms produce that same refusal. Fixing only the one the issue names leaves the bug intact — verified against the working tree.

| Mechanism | Where | Fires on |
|---|---|---|
| `literalHigh != sourceHigh` | `event_search_authority.go` gate | Migration 038 registers a source-sequence row for **every** event; literal `high_water` is frozen at completion. First append. |
| `literalState != "complete"` | same gate | Migration 039's six triggers set the literal row to `'stale'` on every `events` / `command_audits` insert, update and delete. First write. |
| Completion never recorded | `search_projection_rebuild.go:673` | The transition matched only `state='rebuilding'` and did not check `RowsAffected`. An append **during** a rebuild flips the row to `'stale'`, so on a live store the UPDATE routinely matches zero rows and `'complete'` is never reached at all. |

## Fix

- **Widen the candidate walk.** `searchTieredTopKMetadataTx` drops the `WHERE q.sequence<=high_water` bound. It already visits `search_projection_source_sequence JOIN events` in the public order (`created_at_norm DESC, id DESC`) and decodes each candidate, and its fingerprint pre-filter is fail-open for rows with no fingerprints for the generation. Post-cutover rows have none, so they are admitted and decided on their live content — same order, same offset/limit semantics, same `DeepLiteralSearchBudget`. They are newest, so they are visited first and consume budget first.
- **Drop the watermark equality.** A watermark gap is the normal steady state of a written-to store, not a "cannot answer" condition.
- **Accept `literalState='stale'`.** Kept unchanged: `boundedState=='complete'`, `literalGeneration != ""`, `literalGeneration == boundedGeneration`.
- **Fix the completion transition** to `state IN ('rebuilding','stale')` and treat zero affected rows as drift (`SearchProjectionDriftError`), matching the bounded lifecycle update immediately below it.

No migration, no trigger change.

## Why accepting `'stale'` is not a wrong-answer hole

`'stale'` conflates appends with mutations of already-projected rows. Appends are safe — no fingerprints, fail-open, decoded live. Mutations would be unsafe if the literal gate were the only guard, because the row's fingerprints are now wrong and the pre-filter would exclude it silently.

It is not the only guard. The `search_projection_complete_*` triggers in `schema/sqlite/migrations/000041_add_search_projection_generation_lifecycle.sql:16-40` fire on mutation of an already-projected event or audit and set the **bounded** row to `state='drifted'` with `active_generation_id=NULL`. So `boundedState=='complete'` is what rejects a mutated projection. The tests assert that `drifted` state directly, so if that trigger ever stops firing the suite fails loudly instead of degrading into silent omissions.

An earlier design that invalidated fingerprint rows from the mutation triggers was proposed and withdrawn — migration 041 already covers it, and it would have been a migration the issue does not need.

## Tests

`infrastructure/sqlite/event_search_authority_internal_test.go` (new, `package sqlite`). Every case was run against the pre-fix tree first:

| Test | Pre-fix | Post-fix |
|---|---|---|
| `...IncludesPostCutoverTail` (metadata / full / ordering across the cutover) | FAIL — stale error | pass |
| `...SearchWatermarkGapIsNotStale` | FAIL — stale error | pass |
| `...AppendLeavesLiteralStaleAndStillAnswers` | FAIL — stale error | pass |
| `...AuditInsertOnTailEventStaysCompleteAndMatches` | FAIL — stale error | pass |
| `...LiteralCompletionAcceptsStaleFromRebuildAppend` | FAIL — literal stuck at `stale` | pass |
| `...LiteralCompletionZeroRowsIsDrift` | FAIL — reported success | pass |
| `...SearchBudgetExhaustionOnLargeTail` | FAIL — stale error before budget | pass, `EventSearchUnavailableError` / `index_incomplete` |
| `...BodyUpdateOfProjectedEventDriftsAndRefuses` | pass (refused for the wrong reason) | pass, now asserts bounded `drifted` |
| `...AuditInsertOnProjectedEventDriftsAndRefuses` | pass | pass, asserts bounded `drifted` |
| `...FingerprintPreFilter` | pass | pass — real generation fingerprints seeded so the pre-filter is actually exercised |
| `...StillFailsWhenProjectionCannotAnswer` (4 cases: literal rebuilding, bounded rebuilding, generation mismatch, empty generation) | pass | pass |

Fingerprint fixtures use the real writer shape: `CharacterizeLiteralQuery(body).Fingerprints()`, 16-byte blobs at `fingerprint_version=1`, keyed by `generation_id + source_sequence + event_id`.

`TestPersistedTieredAuthorityPreservesDescendingContinuationAndFailsClosed` asserted the old contract (`'stale'` fails closed). Updated to assert the new one: `'stale'` answers, `'rebuilding'` still fails closed.

`go build ./...`, `go tool golangci-lint run` (0 issues) and `go test ./...` all pass.

## Deliberately out of scope

- Making generation completion non-terminal. `CatchUp`'s `already_complete` branch is untouched.
- Tiered availability while a generation is rebuilding, or parked in `failed` after #1732. Both still refuse — separate follow-up, and it blocks #1718.
- The empty-query structural branch sitting behind the literal gate despite its comment claiming it does not need literal traversal — separate follow-up.
- #1718 itself. No authority default is flipped here.

### DIFF
--- a/infrastructure/sqlite/event_search_authority.go
+++ b/infrastructure/sqlite/event_search_authority.go
@@ -131,19 +131,26 @@ func (d *EventDatasource) searchTieredMetadataTx(ctx context.Context, tx *sql.Tx
 		return nil, xerrors.New("offset must be greater than or equal to 0")
 	}
 	var literalState, literalGeneration, boundedState, boundedGeneration string
-	var literalHigh, sourceHigh int64
-	if err := tx.QueryRowContext(ctx, `SELECT generation_id,high_water,state FROM literal_search_projection_state WHERE singleton=1`).Scan(&literalGeneration, &literalHigh, &literalState); err == nil {
-		err = tx.QueryRowContext(ctx, `SELECT COALESCE(active_generation_id,''),state FROM search_projection_state WHERE singleton=1`).Scan(&boundedGeneration, &boundedState)
-		if err == nil {
-			err = tx.QueryRowContext(ctx, `SELECT COALESCE(MAX(sequence),0) FROM search_projection_source_sequence`).Scan(&sourceHigh)
-		}
-		if err != nil {
-			return nil, xerrors.Errorf("read tiered authority projection state: %w", err)
-		}
-	} else {
+	// Generation consistency is a real "cannot answer" condition: mismatched
+	// or non-authoritative projections have no coherent candidate set.
+	// Literal state 'stale' is not one of those conditions — migration 039
+	// flips literal to stale on every events/command_audits write, including
+	// ordinary post-cutover appends whose fingerprint absence already fail-
+	// opens in the walk. Mutated already-projected rows are rejected by the
+	// bounded gate instead: search_projection_complete_* triggers set
+	// bounded state='drifted' and clear active_generation_id. A watermark
+	// gap after completion is also not "cannot answer" — sourceHigh always
+	// advances while literal high_water stays frozen. Post-cutover rows are
+	// visited by the ordered walk below and admitted by the fail-open
+	// fingerprint pre-filter when they lack generation fingerprint rows.
+	if err := tx.QueryRowContext(ctx, `SELECT generation_id,state FROM literal_search_projection_state WHERE singleton=1`).Scan(&literalGeneration, &literalState); err != nil {
 		return nil, xerrors.Errorf("read tiered authority projection state: %w", err)
 	}
-	if literalState != "complete" || boundedState != "complete" || literalGeneration == "" || literalGeneration != boundedGeneration || literalHigh != sourceHigh {
+	if err := tx.QueryRowContext(ctx, `SELECT COALESCE(active_generation_id,''),state FROM search_projection_state WHERE singleton=1`).Scan(&boundedGeneration, &boundedState); err != nil {
+		return nil, xerrors.Errorf("read tiered authority projection state: %w", err)
+	}
+	literalReady := literalState == "complete" || literalState == "stale"
+	if !literalReady || boundedState != "complete" || literalGeneration == "" || literalGeneration != boundedGeneration {
 		return nil, xerrors.New("tiered search projection is incomplete or stale")
 	}
 	// Empty queries are structural-only searches. They do not require literal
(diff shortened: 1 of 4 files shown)


duck8823:
Sol review (gpt-5.6-sol, read-only): one HIGH, no CRITICAL. Verdict was "do not merge". I checked it and disagree with the verdict, while agreeing the finding is real. Reasoning below.

## Sol HIGH — invalidators do not watch every column the decoder depends on

`search_projection_complete_event_update` (`000041:17`) watches `body, body_availability, session_id, kind, created_at`. `literal_search_event_update` (`000039:33`) watches those plus `workspace, client, agent`. Neither watches `body_codec`, `body_format_version`, `body_plaintext_bytes`, `body_encoded_bytes`, `body_sha256` — all of which the decoder in `payload_hydration.go` reads. So a codec-metadata-only mutation of a projected event changes its decoded text while leaving both projection states `complete`, and the event is then excluded by the fingerprint `COUNT(DISTINCT ...)` branch rather than fail-open. Silent false negative.

Confirmed by reading the migrations, and confirmed empirically with a throwaway probe:

```
before: literal="complete" bounded="complete"
UPDATE events SET body_codec='zstd' WHERE id='projected'
after:  literal="complete" bounded="complete"
```

**But this is not introduced by this PR.** The probe is the proof: a codec-only mutation leaves both states `complete` *and* leaves the watermarks equal, because no event was inserted. So the pre-#1733 gate — `literalState=='complete' && boundedState=='complete' && literalHigh==sourceHigh` — admits the search identically and produces the identical false negative. This PR neither creates nor widens the gap.

It is also not reachable from production code today. Both writers that touch codec columns (`raw_body_retention_datasource.go:249-256` prune, `:477-484` restore) include `body=?` in the same SET list, and SQLite's `UPDATE OF` fires on column *mention*, so both invalidators do fire. The only codec-only mutation in the tree is a deliberate drift fixture at `payload_rehearsal_swap_internal_test.go:70`.

So: a real latent contract gap, pre-existing, not reachable in production, and one that must be closed before #1718 removes the fallback — not a blocker for a change that leaves it exactly as it found it. Filed as #1737, marked as blocking #1718.

Sol's secondary note that `events.id` is unwatched is folded into the same issue. No production path updates the primary key; the question there is a guard, not an invalidation.

## Sol found no CRITICAL/HIGH in

`command_audits` insert/update/delete; event delete and audit cascade; retention prune/restore; body-rewriting redaction; cross-generation source-sequence reuse and delete/re-insert with the same id; fingerprint writer atomicity and fixture shape; partial fingerprints from an interrupted rebuild; budget exhaustion returning `EventSearchUnavailableIndexIncomplete`; offset, page anchor and ordering across the widened walk; the zero-row drift verdict and its interaction with #1732's parking.

Sol independently ran `go test ./infrastructure/sqlite -run 'TestTieredAuthority' -count=1` and `go test ./...` — both pass.

## Follow-ups filed

- #1735 — tiered search availability while rebuilding / `failed` / `drifted`. Blocks #1718.
- #1736 — structural (empty-query) searches gated on the literal projection they do not use.
- #1737 — this finding. Blocks #1718.

CI green. Marking ready for review.

Issue #1736, opened 2026-08-09
v0.37-1: structural searches should not be gated on the literal projection


duck8823:
## Problem

In `searchTieredMetadataTx` (`infrastructure/sqlite/event_search_authority.go`), the empty-query structural branch sits **after** the projection gate:

```go
if !literalReady || boundedState != "complete" || literalGeneration == "" || literalGeneration != boundedGeneration {
    return nil, xerrors.New("tiered search projection is incomplete or stale")
}
// Empty queries are structural-only searches. They do not require literal
// traversal ...
if criteria.Query() == "" {
    ...
}
```

The comment states the branch does not require literal traversal, and that is true — a structural search filters on workspace, session, kind, time and never touches `literal_search_fingerprints`. But it is only reached once the literal and bounded projections have both been judged answerable. So a structural search is refused for a reason that does not apply to it.

Combined with the availability states listed in the sibling follow-up (rebuilding, `failed`, `drifted`), this means `traceary list --workspace X` style structural queries stop working whenever the *literal* projection is mid-rebuild, even though nothing about the structural query depends on it.

## Scope

Move the empty-query structural branch above the gate, or narrow the gate so it only applies to queries that actually traverse the literal projection. Confirm by reading the branch which projection tables it touches — if it reads `search_projection_source_sequence` for ordering it may still need the bounded generation to be coherent, in which case the gate should be split rather than simply reordered.

## Acceptance criteria

- A structural (empty-query) search under `authority='tiered'` succeeds while the literal projection is in a state that refuses literal searches.
- Whatever coherence the structural branch genuinely needs is still enforced, and a test covers a state where the structural branch *must* refuse.
- Non-structural searches are unaffected — the existing refusal cases still fail closed.

Follow-up from #1733 / PR #1734.

Issue #1737, opened 2026-08-09
v0.37-2: projection invalidation triggers do not watch every column the decoder depends on


duck8823:
## Problem

The search projection's invalidation triggers watch a narrower set of columns than the decoder actually depends on. A mutation of an unwatched column changes an event's decoded searchable text without invalidating the fingerprints that describe it, so the fingerprint pre-filter excludes the event from results — a **silent false negative**, not an error.

Watched columns today:

| Trigger | Migration | `UPDATE OF` columns |
|---|---|---|
| `search_projection_complete_event_update` | `000041:17` | `body, body_availability, session_id, kind, created_at` |
| `literal_search_event_update` | `000039:33` | `workspace, client, agent, session_id, kind, created_at, body, body_availability` |
| bounded source-revision event update | `000038` / `000050` | same narrow set |

Columns the decoded body depends on but which no invalidator watches (`infrastructure/sqlite/payload_hydration.go`):

```
body_codec
body_format_version
body_plaintext_bytes
body_encoded_bytes
body_sha256
```

`events.id` is also unwatched. Updating a projected event's id leaves `search_projection_source_sequence` pointing at the old id, and the walk's inner join on `events` silently drops the row.

The codec triggers in `000036_add_payload_codec_foundation.sql:46-57` are compatibility **counters**, not invalidators — they do not touch either projection state.

### Failure scenario

1. Event P is projected in generation G and has G fingerprints.
2. P's codec metadata is changed without touching `body` — same stored bytes, different decode.
3. The new decoded text contains query term Q; the fingerprints, built from the old decode, do not.
4. Neither migration 039 nor 041 fires. Both projection states stay `complete`.
5. The gate admits the search. P has *some* fingerprints for G, so it does not take the fail-open `NOT EXISTS` branch — it is excluded by `COUNT(DISTINCT ...)=?` in `event_search_authority.go`.
6. `decodedEventSearchMatch` is never reached. The search returns `[]` where `[P]` is correct.

The same gap applies during a rebuild: a codec-only update does not bump `search_projection_source_revision`, so a generation can be marked `complete` while carrying fingerprints built from a superseded decode.

## Reachability today

[...]

PR #1738, opened 2026-08-09
fix: keep tiered search available when the projection cannot be trusted

### LINKED ISSUE #1735: v0.34-56: make tiered search available or explicit while rebuilding, failed, or drifted
## Problem

`searchTieredMetadataTx` (`infrastructure/sqlite/event_search_authority.go`) requires `boundedState == "complete"` and `literalGeneration == boundedGeneration`. Two ordinary lifecycle states violate that and leave tiered search returning a bare error rather than a degraded but honest answer:

**1. While a generation is rebuilding.** `Start` sets the literal singleton to the new generation with `state='rebuilding'`, while the bounded row's `active_generation_id` still points at the previously completed generation. `literalGeneration != boundedGeneration`, so every non-structural search is refused for the whole duration of the rebuild — which on a large store is not short.

**2. When a generation is parked in `failed`.** #1732 made a failed generation parked rather than retried: `CatchUp` returns `skipped` with `parked after generation failure <class>` and does not start a replacement. `boundedState` stays `failed`, so tiered search is refused until an operator notices and runs `traceary store search-projection start`. Nothing surfaces that the refusal and the parking are the same fact.

**3. When a projection has drifted.** The `search_projection_complete_*` triggers (migration 041) set bounded `state='drifted'` on mutation of an already-projected row. Correct as a guard — #1733 relies on it — but the resulting user experience is again a bare error with no indication that a rebuild is what fixes it.

In all three the caller gets `xerrors.New("tiered search projection is incomplete or stale")`, which is neither a `queryservice.EventSearchUnavailableError` nor actionable.

## Why this blocks #1718

#1718 flips the authority default to tiered and deletes the legacy `event_search_*` family. After that there is no fallback: any state that refuses is a total loss of search until an operator intervenes. Before the fallback goes away, refusal has to be either avoidable or self-explanatory.

## Scope

Decide and implement, for each of the three states, whether tiered search should:

- **answer from the previous complete generation** where one exists (rebuilding is the obvious candidate — the old generation is still coherent and `active_generation_id` still names it), or
- **degrade explicitly** via `queryservice.EventSearchUnavailableError` with a reason that names the recovery command, rather than a bare error.

Also decide whether `failed`/`drifted` should surface in `traceary store search-projection status` prominently enough that an operator sees it without being told to look.

Out of scope: making generation completion non-terminal; changing the parking behaviour from #1732.

## Acceptance criteria

- A search issued while a generation is rebuilding either returns results from the still-coherent previous generation, or returns an `EventSearchUnavailableError` whose reason names the state and the recovery step — not a bare error.
- Same for `failed` and `drifted`.
- Tests cover each of the three states on both `SearchMetadata` and `SearchPage`.
- No regression in the refusal cases #1733 kept: genuinely mismatched or empty generations still fail closed.

Follow-up from #1733 / PR #1734. Blocks #1718.

duck8823:
Closes #1735

## Problem

`searchTieredMetadataTx` refused with a bare `tiered search projection is incomplete or stale` whenever the literal and bounded projections were not a coherent completed pair. Four ordinary states hit that:

| State | Why the gate fails |
|---|---|
| Rebuilding | `Start` points the literal singleton at the new generation while bounded's `active_generation_id` still names the previous complete one → `literalGeneration != boundedGeneration` for the whole rebuild |
| `failed` | #1732 parks a failed generation instead of retrying, so `boundedState` stays `failed` until an operator runs `search-projection start` |
| `drifted` | Migration 041's `search_projection_complete_*` triggers set bounded `drifted` with `active_generation_id=NULL` on mutation of a projected row |
| No generation | Every store's initial state |

#1718 deletes the legacy `event_search_*` family. After that there is no fallback, so any of these means search is simply gone.

## Fix

The gate conflated two questions, and separating them removes the error mode entirely.

**The fingerprint index is an optimisation, not a correctness requirement.** `searchTieredTopKMetadataTx` decides every candidate by decoding it through `decodedEventSearchMatch`; the fingerprint clause is a pre-filter that skips decoding for rows it *proves* cannot match, and it is fail-open for rows with no fingerprints for the generation. So "can we answer" and "can we trust the fingerprints" are separate questions, and only the second has a false case.

- `usableLiteralFingerprintGeneration` applies the old gate's conditions as a **usability** test and returns `""` when they do not hold.
- An empty generation **skips the fingerprint clause outright** rather than relying on it matching nothing. Explicit, avoids three correlated subqueries per candidate row, and does not depend on an invariant about which generation ids can appear in `literal_search_fingerprints`.
- The refusal is deleted. There is no projection state in which the tiered path cannot answer.

A rebuild deliberately does **not** fall back to the previous generation's fingerprints. Whether the rebuild has already overwritten or deleted those rows is not observable from the read side, and trusting them wrongly is a silent false negative. Searches decode every candidate until the rebuild completes — a real cost, named in the code comment and in the docs.

Budget exhaustion is unchanged and is now the only degradation path: a decode-bound walk that runs out still reports `EventSearchUnavailableIndexIncomplete` rather than truncating.

The empty-query structural branch moves above the state read. It never traversed the literal projection, and with the gate gone it no longer waits on one — which is what #1736 describes. Nothing else about #1736 is addressed here.

No migration. Read-path only. `authority='legacy'` is untouched and no default is flipped.

## Tests

`infrastructure/sqlite/event_search_authority_internal_test.go`. Every new or rewritten case was run against the pre-fix tree first:

| Test | Pre-fix | Post-fix |
|---|---|---|
| `...SearchWithoutProjectionGeneration` — fresh store, no projection ever built | FAIL — refused | pass, both paths, correct order |
| `...SearchDuringRebuildSkipsOldPreFilter` | FAIL — refused | pass |
| `...SearchAnswersWhenGenerationFailed` | FAIL — refused | pass |
| `...BodyUpdateOfProjectedEventDriftsAndAnswersMutated` | FAIL — refused | pass, returns the **mutated** body |
| `...AuditInsertOnProjectedEventDriftsAndAnswers` | FAIL — refused | pass |
| `...SearchBudgetExhaustionWithoutPreFilter` | FAIL — refused before budget | pass, `index_incomplete` |
| `...SearchAnswersWhenProjectionUnusable` (4 cases: literal rebuilding, bounded rebuilding, generation mismatch, empty generation) | FAIL — refused | pass |
| `TestPersistedTieredAuthorityPreservesDescendingContinuationAndAnswersWhenUnusable` | FAIL — refused | pass |
| `...FingerprintPreFilter` — usable generation still pre-filters | pass | pass, unchanged |

`...SearchDuringRebuildSkipsOldPreFilter` is the one that proves the pre-filter is really skipped rather than silently applied: it seeds `would-exclude` with a body that matches the query but fingerprints from a non-overlapping string, asserts that a **usable** generation excludes it, then calls `Start` and asserts it now surfaces.

`...BodyUpdateOfProjectedEventDriftsAndAnswersMutated` queries a term unique to the post-mutation body, so an answer can only have come from decoding, not from the pre-mutation index. It also asserts `active_generation_id` is NULL and bounded is `drifted`, so a trigger regression fails loudly.

`go build ./...`, `go tool golangci-lint run` (0 issues) and `go test ./...` all pass.

## Docs

`docs/operations/search-maintenance.md` and its `.ja.md` pair said tiered search "fail[s] closed if its projections become incomplete or stale". Both updated to state the new behaviour and its cost.

### DIFF
--- a/application/queryservice/event_search_errors.go
+++ b/application/queryservice/event_search_errors.go
@@ -21,8 +21,10 @@ const (
 	// EventSearchUnavailableScopeTooBroad rejects a legacy fallback whose
 	// structural/time candidate set cannot be proven below the hard cap.
 	EventSearchUnavailableScopeTooBroad EventSearchUnavailableReason = "scope_too_broad"
-	// EventSearchUnavailableIndexIncomplete rejects an unbounded indexed query
-	// while historical search documents are still being backfilled.
+	// EventSearchUnavailableIndexIncomplete rejects a query the index could not
+	// finish answering: a legacy backfill still in progress, or a tiered walk
+	// that ran out of candidate budget because no trustworthy fingerprint
+	// generation was available to pre-filter with.
 	EventSearchUnavailableIndexIncomplete EventSearchUnavailableReason = "index_incomplete"
 )
 
@@ -38,8 +40,16 @@ type EventSearchUnavailableError struct {
 func (e *EventSearchUnavailableError) Error() string {
 	switch e.Reason {
 	case EventSearchUnavailableIndexIncomplete:
+		// Two very different situations reach this reason, and the message
+		// must not name only one of them. A legacy store is mid-backfill; a
+		// tiered store has no usable fingerprint generation, so every
+		// candidate is decoded and the budget runs out on a wide query. Both
+		// are fixed either by narrowing the query or by finishing the index,
+		// so the message names both remedies rather than guessing which.
 		return fmt.Sprintf(
-			"event search index backfill is incomplete; add workspace/session and from/to bounds (legacy candidate limit %d)",
+			"event search could not finish within the candidate budget of %d rows; "+
+				"add workspace/session and from/to bounds, "+
+				"or complete the search projection with 'traceary store search-projection start'",
 			e.CandidateLimit,
 		)
 	default:
--- a/docs/operations/search-maintenance.ja.md
+++ b/docs/operations/search-maintenance.ja.md
@@ -31,7 +31,12 @@ DROP、VACUUM しません。
    legacy plaintext document は指定行数以内です。進捗と前後の logical /
    physical bytes を永続化するため、中断後も再開できます。
 4. `status` を確認します。通常の CLI / MCP 検索は永続化された tiered
-   authority を使用し、projection が incomplete / stale なら fail closed します。
+   authority を使用します。projection が incomplete / rebuilding / failed /
+   drifted でも検索は利用不能になりません。fingerprint index は pre-filter に
+   すぎないため、候補を復号して判定する経路へ fail open し、正しい結果を返します。
+   代償は正しさではなく処理量です。この状態の検索は decode-bound になり、deep
+   literal search budget を使い切った場合は結果を切り詰めずに `index_incomplete`
+   を報告します。
 
 rollback は codec decode を含む canonical history から再構築します。
 
--- a/docs/operations/search-maintenance.md
+++ b/docs/operations/search-maintenance.md
@@ -29,8 +29,13 @@ Retirement is an operator-only workflow:
 3. Repeatedly run `resume-retire --rows 128`. Each transaction removes at most
    the requested number of legacy plaintext documents and records progress and
    before/after logical and physical bytes. Interruption is safe to resume.
-4. Check `status`. Normal CLI and MCP search use the persisted tiered authority
-   and fail closed if its projections become incomplete or stale.
+4. Check `status`. Normal CLI and MCP search use the persisted tiered
+   authority. An incomplete, rebuilding, failed, or drifted projection does not
+   make search unavailable: the fingerprint index is only a pre-filter, so
+   searches fall back to decoding each candidate and still return correct
+   results. The cost is work, not correctness — a search in that state is
+   decode-bound, and if it exhausts the deep literal search budget it reports
+   `index_incomplete` rather than truncating the answer.
 
 Rollback uses canonical history, including codec decoding:
 
--- a/infrastructure/sqlite/event_search_authority.go
+++ b/infrastructure/sqlite/event_search_authority.go
@@ -162,11 +162,22 @@ func (d *EventDatasource) searchTieredMetadataTx(ctx context.Context, tx *sql.Tx
 // triggers set bounded state='drifted' and clear active_generation_id, so
 // such a generation stops being usable here and the walk decodes live content.
 //
-// A rebuild in progress is deliberately not answered from the previous
-// generation's fingerprints. Whether the rebuild has already overwritten or
-// deleted those rows is not observable from the read side, and trusting them
-// wrongly is a silent false negative. The cost is that searches decode every
-// candidate until the rebuild completes.
+// A rebuild in progress is not answered from the previous generation's
+// fingerprints, and the cost of that is real: searches decode every candidate
+// until the rebuild completes, so a wide query on a large store exhausts the
+// budget and reports index_incomplete.
+//
+// Those rows do survive. Start only repoints the literal singleton
+// (search_projection_rebuild.go:82), fingerprint writes are additive inserts
+// keyed by generation (:624), and old-generation rows are removed only in the
+// new generation's cleanup phase (:445, :633). So the previous generation is
+// still identifiable through active_generation_id and still readable.
+//
+// Using it safely needs more than identifying it: the rebuild must not have
+// entered cleanup, and no canonical mutation may have landed since the old
+// generation completed, or its fingerprints describe rows that have changed.
+// That is a distinct piece of work with its own failure modes, tracked
+// separately, so this deliberately takes the slow, always-correct path.
 func usableLiteralFingerprintGeneration(ctx context.Context, tx *sql.Tx) (string, error) {
 	var literalState, literalGeneration, boundedState, boundedGeneration string
 	if err := tx.QueryRowContext(ctx, `SELECT generation_id,state FROM literal_search_projection_state WHERE singleton=1`).Scan(&literalGeneration, &literalState); err != nil {
--- a/infrastructure/sqlite/event_search_authority.go
+++ b/infrastructure/sqlite/event_search_authority.go
@@ -130,31 +130,8 @@ func (d *EventDatasource) searchTieredMetadataTx(ctx context.Context, tx *sql.Tx
 	if criteria.Offset() < 0 {
 		return nil, xerrors.New("offset must be greater than or equal to 0")
 	}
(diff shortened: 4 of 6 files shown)


duck8823:
Self-review found a hole in the first push, fixed in `0773e9b0` (force-pushed over `e136d3c4`).

## The candidate set was anchored on the wrong table

`searchTieredTopKMetadataTx` walked `search_projection_source_sequence JOIN events`. That table is the **projection's own checkpoint ledger**, not an index of events. Rows arrive two ways:

- the migration-038 `search_projection_events_insert` trigger, for events recorded after the migration ran;
- `search_projection_rebuild.go:271` — `INSERT OR IGNORE INTO search_projection_source_sequence(event_id) SELECT id FROM events WHERE id=?` — the rebuild's **inventory phase**, one event at a time, for everything that predates the trigger. That is what `search_projection_inventory_compat.requires_inventory` exists to schedule.

So on an upgraded store that has never completed a generation, the table contains **nothing from before the upgrade**. Under the old gate that was invisible: no complete generation meant the search was refused outright. Removing the refusal without changing the candidate set converts that loud refusal into a **silent omission of the entire pre-upgrade history** — the exact inversion of this PR's purpose.

Reproduced before fixing:

```
--- FAIL: TestTieredAuthoritySearchFindsEventsNeverInventoried
    SearchMetadata() IDs mismatch (-want +got):
      []string{
      	"inventoried",
    - 	"never-inventoried",
      }
```

The fix is to walk `events` directly. Nothing else in the query needed the join — the SELECT list is entirely `e.*` / `a.*`, `appendEventSearchFilters` only emits `e.` and `a.` predicates, and the fingerprint subqueries key on `known.event_id=e.id`, not on sequence. Once #1733 removed the `q.sequence<=high_water` bound, the join's only remaining effect was to restrict candidates to inventoried events.

It is also faster. `EXPLAIN QUERY PLAN` on both shapes:

```
new: SCAN e USING COVERING INDEX idx_events_created_at_norm_id_desc
old: SCAN e USING COVERING INDEX idx_events_created_at_norm_id_desc
     SEARCH q USING COVERING INDEX sqlite_autoindex_search_projection_source_sequence_1 (event_id=?)
```

Same covering-index scan, minus one index probe per candidate row.

## Two tests added

- `TestTieredAuthoritySearchFindsEventsNeverInventoried` — an event with no source-sequence row is found. Fails on the previous push, passes now.
- `TestTieredAuthoritySearchIgnoresOrphanedSourceSequenceRows` — the other direction. Nothing deletes a source-sequence row when its event is deleted, so the table outlives its events; the fixture asserts the orphan really exists before asserting the deleted event does not appear in results.

`go build ./...`, `go tool golangci-lint run` (0 issues) and `go test ./...` all pass.

duck8823:
Sol review (gpt-5.6-sol, read-only) of `e136d3c4`: two HIGH, no CRITICAL. Both accepted. One was already fixed before the review landed; the other is split between a fix here and a follow-up.

## HIGH 1 — source-sequence omission → already fixed in `0773e9b0`

Sol independently reproduced the same failure I found while waiting for the review, with the same test shape:

```
SearchMetadata() IDs mismatch:
  ["inventoried",
-  "never-inventoried"]
```

It also correctly identified why `TestTieredAuthoritySearchWithoutProjectionGeneration` passed for an insufficient reason — it inserts its events after the migration, so the trigger gives every one a source-sequence row and the upgraded-store shape is never exercised. `TestTieredAuthoritySearchFindsEventsNeverInventoried` covers that shape now. Fixed by walking `events` directly; see the previous comment for the plan comparison.

## HIGH 2 — decode-bound searches during a rebuild, and a message that misattributes them

Two separate claims, and both check out.

**The message was wrong.** `EventSearchUnavailableIndexIncomplete` rendered as `event search index backfill is incomplete; add workspace/session and from/to bounds (legacy candidate limit N)`. Under the tiered authority nothing is backfilling and nothing is legacy — the walk simply ran out of budget because no trustworthy fingerprint generation was available. Fixed in `76cdab21`: the text now names the budget, both remedies, and the recovery command. The stable reason string `index_incomplete` is unchanged, so the CLI/MCP classification and the operations docs that pin it are untouched.

**My comment justifying the slow path was factually false.** I wrote that the previous generation's fingerprints may have been overwritten or deleted and were "not observable from the read side". Verified against the code Sol cites, and Sol is right:

- `search_projection_rebuild.go:82` — `Start` only repoints the literal singleton. It deletes nothing.
- `:624` — fingerprint writes are additive `INSERT OR IGNORE` keyed by `generation_id`.
- `:445` / `:633` — old-generation rows are removed only in the **new generation's cleanup phase**, via the `WHERE generation_id<>?` branch.
- `active_generation_id` still names the old generation throughout.

So during `source` and `eviction` the previous generation is intact and addressable. The comment is corrected in `76cdab21` to say that, rather than defending the slow path with something untrue.

**Actually using it is a follow-up, not this PR.** Two conditions have to be proven from the read side, and neither is a one-liner: cleanup must not have started (deletion is batched, so partway through counts as gone), and no canonical mutation may have landed since the old generation completed. The migration-041 drift triggers that #1733 leans on do **not** fire here, because they require bounded `state='complete'` and during a rebuild it is `rebuilding` — so the signal would have to be `search_projection_source_revision`, whose watched column set #1737 says is already incomplete. Filed as **#1739**, to land before #1718.

I disagree with one framing in the finding. Sol reads the outcome as "turns many non-trivial searches into unavailable-with-a-reason" rather than making search available. On a large store mid-rebuild that is a fair description of the effect — but the comparison is against the previous behaviour, which was a bare `tiered search projection is incomplete or stale` for *every* query in that state, wide or narrow, with no reason and no recovery hint. Narrow and time-bounded queries now succeed, and the ones that do not now say what to do. That is an improvement, and #1739 closes the remaining gap.

## Sol found no CRITICAL/HIGH in

SQL argument ordering with the fingerprint clause conditionally absent; `decodedEventSearchMatch` and hydration having no generation or high-water dependency; the structural branch's move above the state read; the rebuild test's before-assertion genuinely exercising the pre-filter; read-transaction interference with the rebuild's write lock (WAL — it notes possible checkpoint delay and WAL growth, not blocking).

## Note on the review run

Sol reports that a sub-reviewer wrote uncommitted changes into the working tree against its read-only instruction, and that it worked from a `/private/tmp` copy of `e136d3c4` to avoid contamination. `git status` here is clean and the branch contains only my three commits, so nothing leaked into what is being reviewed — but worth recording.

## Follow-ups now open from this line of work

- #1737 — invalidation triggers do not watch every column the decoder depends on. Blocks #1718.
- #1739 — use the previous complete generation's fingerprints during a rebuild. Should land before #1718.
- #1736 — structural searches gated on the literal projection. The symptom is resolved here as a side effect; the issue stays open for the explicit narrowing.

CI pending. Will mark ready once green.

Issue #1740, opened 2026-08-09
v0.34-60: bounded search projection's FTS index accumulates tombstones with no deterministic reclaim


duck8823:
> **Rewritten 2026-08-10.** The original text said the FTS index was outside the byte budget. That was wrong — it enters the budget through measured amplification. The tombstone growth is real; the consequence is not what was written. Corrected below after an independent analysis by gpt-5.6-sol and my own measurement. The original framing is preserved at the bottom.

## Problem

`search_projection_recent_fts` accumulates FTS5 delete markers and nothing in this repository reclaims them deterministically. The live corpus can stay constant while the index behind it grows.

FTS5 records a deletion as an *appended* marker, not an in-place removal. Rows are deleted row-by-row (`infrastructure/sqlite/search_projection_rebuild.go:783-807`) in two routine paths — steady-state eviction (`:576-589`) and old-generation cleanup — and each firing of `search_projection_recent_ad` (`schema/sqlite/migrations/000038_add_bounded_search_projections.sql:20-21`) appends a marker.

## What the budget actually does

The index is **not** outside the budget. `measure_search_projection_family_split.sql` classifies FTS shadow tables onto the recent side deliberately, and the chain is:

```
SourceCeiling = floor((budget - nonRecentReserve) × 1e6 / AmplificationPPM)
```

with `AmplificationPPM` derived from recent physical dbstat bytes over source bytes (`search_projection_rebuild.go:1091-1155`), persisted (`:208-217`), and enforced by eviction against `SUM(decoded_bytes)` (`:573-589`). Tombstones inflate the measured physical side, so they *do* register.

**But the budget is not a hard bound.** Post-completion family bytes are measured and recorded into `index_family_within_budget` (`:1327-1346`, `application/usecase/search_projection_usecase.go:301-304`) and nothing corrects an overrun. There is no mechanism that shrinks FTS physical allocation to get back inside the budget. So the store can exceed the budget and stay there, reporting the fact.

## The actual failure mode

Three behaviours interact, and none of them alone is the story:

1. **While the amplification sample is large enough**, rising physical bytes raise `AmplificationPPM`, which lowers `SourceCeiling`, which shrinks the retained window — roughly `1/n` in the number of churn generations. Search coverage degrades with no change in retained content.
[...]

Issue #1739, opened 2026-08-09
v0.37-3: use the previous complete generation's fingerprints while a rebuild is in progress


duck8823:
## Problem

After #1735, a tiered search whose fingerprint generation is unusable falls back to decoding every candidate. That is always correct, but on a large store it means a wide query exhausts `DeepLiteralSearchBudget` (4,096 rows / 64 MiB stored / 128 MiB decoded) before reaching an older match, and returns `index_incomplete` instead of results.

The state where this bites hardest is an ordinary rebuild. `Start` points the literal singleton at the new generation while bounded's `active_generation_id` still names the previous complete one, so `usableLiteralFingerprintGeneration` (`infrastructure/sqlite/event_search_authority.go`) returns `""` for the entire duration — which on a store the size of the one #1718 targets is not short.

### The previous generation is still there

The reason #1735 gives for not using it — that the rebuild may have overwritten or deleted those rows — is wrong, and the code says so:

- `Start` only repoints the literal singleton; it deletes nothing (`infrastructure/sqlite/search_projection_rebuild.go:82`).
- Fingerprint writes are additive `INSERT OR IGNORE ... (generation_id, source_sequence, event_id, fingerprint, 1)` keyed by generation (`:624`).
- Old-generation rows are removed only in the **new generation's cleanup phase**, via the `WHERE generation_id<>?` branch (`:445`) executed at `:633`.
- The old generation stays identifiable through `search_projection_state.active_generation_id` until the new one completes.

So during the `source` and `eviction` phases of a rebuild, the previous generation's fingerprints are intact, complete for everything at or below its `high_water`, and addressable.

## Why it is not simply a one-line change

Trusting them requires proving two things from the read side:

1. **Cleanup has not started.** `search_projection_state.phase` distinguishes `source` / `eviction` / `cleanup` / `complete`. Cleanup with `cleanup_scope='old'` is exactly the deletion of the previous generation's rows, so the fingerprints must be considered gone from the moment that phase begins — including partway through, since deletion is batched.
[...]

PR #1741, opened 2026-08-09
feat: retire the legacy migration-032 search index

### LINKED ISSUE #1718: v0.34.0-49: retire the legacy full-corpus search index

duck8823:
Closes #1718

## What

The full-corpus search index family from migration 032 (`event_search_documents`, `event_search_fts`, `event_search_backfill_state`) has not been read since the bounded tiered projection replaced it, but it was still maintained on every write and still resident. On the maintainer's store it holds **16.15 GiB of 24.7 GiB — about 65% of the database**.

This stops maintaining it and adds one command to remove it.

## Upgrade experience

Per the decision recorded on the issue — *startup succeeds, the DROP is opt-in, doctor prompts* — the work is split:

| | |
|---|---|
| **Migration 052** (automatic, first startup) | All 8 writer triggers + `event_search_projection` view + `search_maintenance_control`. Constant cost. |
| **`traceary store search-retire`** (explicit) | The three large tables, in one transaction. Idempotent. |

The three triggers on `event_search_documents` go with the five on `events` / `command_audits`. That column is `ON DELETE CASCADE` and the DSN enables `foreign_keys`, so `store gc` and retention still reached the index through the foreign key; leaving those triggers alive would have made every deleted event append an FTS5 delete marker, growing the index this work removes.

Startup applies every pending migration unconditionally, so a multi-GiB `DROP TABLE` inside a migration would block the first `traceary` invocation after upgrading, with no way to decline. Startup stays fast; reclaiming the space is the operator's call. The v0.34 store-size gate is therefore met *by operators who run the command*, not automatically.

## Two measurements that shaped the design

1. **Straight `DROP`, not `DELETE`-then-`DROP`.** My original brief specified batched deletes for resumability. Measurement disproved it: emptying an FTS5 content table appends delete markers into new segments rather than reclaiming anything, so the file grows **+14%** and the index **+47%** before shrinking, and the whole thing takes **~8× longer**. `DROP` is one transaction, interruptible with a clean rollback.
2. **The bounded projection's own FTS index grows without bound across generations** (10.17 → 59.56 MB over 4 generations with a constant live corpus) and is excluded from the byte budget. Out of scope here — filed as #1740.

## Surface reduction

Six `store search-maintenance` subcommands (`adopt-target`, `start-retire`, `resume-retire`, `start-restore`, `resume-restore`, `status`) driving a resumable two-authority transition → **one** `store search-retire`. Net −5. The `store-benchmark --search-parity-manifest` / `--validate-search-parity` / `--authorize-search-parity` modes are removed too: they existed to compare the two indexes against each other, and there is no longer a second index. ~5,900 lines deleted against ~1,100 added.

## Ordering is enforced

```sh
traceary store search-retire        # drops the tables; frees pages
traceary store compact plan         # refuses if the family is still present
traceary store compact apply RUN_ID # file actually shrinks here
```

`DROP` returns pages to SQLite's free list; `auto_vacuum` is `NONE`, so the file keeps its size until `store compact` (`VACUUM INTO`). Compacting first would copy 16 GiB of dead index into the new file and bake it in, so `compact plan` fails closed — **before the source digest**, so it errors in seconds rather than after hashing a 24 GiB store. `apply` re-checks behind its exclusive lease. `doctor` warns while the family is resident, naming the bytes and the fix command; the ≥2 GiB metadata-only doctor mode gives the same advice unconditionally, since it deliberately never opens SQLite and `search-retire` is a no-op when the family is already gone.

## Rollback

**Roll-forward only.** `git revert` is not a rollback: with the writer triggers gone, a downgraded binary would query an index that stopped receiving writes at the moment of upgrade and silently return incomplete results. Restore from a backup taken before the retirement.

## Tests

- Migration 052 applies over all four `authority`/`phase` states migration 040 permitted, including one interrupted mid-retirement, and search answers afterwards
- Migration 052 leaves the large tables and removes the constant-cost objects
- `search-retire` **drops** rather than empties (asserts the tables are gone), search results are byte-identical across the retirement, and a rerun reports `already_removed` with unchanged physical bytes
- `compact plan` refuses with an error naming `search-retire`; `requireStaticSearchState` guards candidate construction
- The `store search-maintenance` group no longer exists and `store search-retire` does

Two incidental fixes surfaced while testing: the plan-time guard was gated on `run.Operation == compaction`, but `store compact` leaves `Operation` empty, so it never fired — rewritten as an exclusion of the prepared-migration publication. And the MCP server test's hand-transcribed schema had drifted far enough to fail; it now uses the real migration set.

## Verification

`go build ./...`, `go vet ./...`, `go test ./...`, `go tool golangci-lint run` — all clean.

Not exercised against the live 24.7 GiB store: repo-built binaries apply unreleased migrations irreversibly, so that measurement belongs to the #1620 gate run on a released build.

## Follow-ups filed

- #1740 — bounded projection FTS index grows without bound across generations
- #1736 / #1737 / #1739 — surfaced during this work


duck8823:
## Multi-AI review

**Claude Opus 5 (main reviewer)** — implementation by Grok 4.5 high, reviewed and corrected here.

**gpt-5.6-sol (Codex verifier)** — four adversarial passes. Five HIGH findings across them, all reproduced against the code before fixing:

1. `store search-retire` did not migrate the store before dropping the family, so running it as the first command after upgrading would leave the migration-032 source-side triggers pointing at dropped tables — every subsequent event insert would fail. Fixed by migrating first, plus a defensive trigger drop so correctness does not depend on call order.
2. Migration 052 dropped only the five source-side triggers, so the family was not inert. `event_search_documents.event_id` is `ON DELETE CASCADE` and the DSN enables `foreign_keys`, so `store gc` and retention still fired `event_search_documents_after_delete` and appended FTS5 delete markers — growing the index this work removes. 052 now drops all eight triggers.
3. A compaction run journaled at `candidate_verified` by an older binary resumes straight to the exchange without revisiting Plan or Build, bypassing both existing guards.
4. So does a run journaled at `swap_intent`: `RecoveryActions` answers `[ActionExchange, ActionRecordSwapped]` and `resumeLeased` performs that exchange before `applyLeased` is reached, so a phase-entry check never saw it. The refusal was moved to sit immediately before every forward exchange, and `Phase.SwapObserved()` was deleted.
5. The sibling prepared-store-upgrade protocol published unguarded — its `executeRecoveryAction` handled `ActionExchange` and `ActionRollbackExchange` in one case arm, so the forward direction inherited the rollback's exemption. Not reachable today (only the payload-rehearsal migration recipe is registered, and that operation is exempt by design), but `Operation.Known()` admits compaction, so a future recipe would have inherited an unguarded path. Arm split; both forward exchanges now ask.

Each fix carries a regression test that was verified to fail without it — including the FTS shadow-table measurement (`event_search_fts_data` gains a row per deleted event without fix 2) and the two resume paths.

Final verdict — **gpt-5.6-sol: マージ可**. All eight `files.Exchange` call sites re-enumerated: the four forward ones are guarded, the four rollback ones are exempt by design.

Verification: `go build ./...`, `go vet ./...`, `go test ./...`, `go tool golangci-lint run` (0 issues). CI green.

Issue #1765, opened 2026-08-10
v0.34.0-50: deprecate the sessions interactive dashboard and give notice


duck8823:
## Problem

`traceary sessions` with no `--snapshot` opens a five-pane live Bubble Tea dashboard in a TTY (`presentation/cli/top.go:90-108,176-229`). A live watch surface serves neither v0.34 pillar: it does not consolidate a session and it does not supply a summary at wake time. It is the same class of surface as the cockpit (#1687), split out because the two have different implementations and must be revertible independently — this one lives in the dashboard runner in `top.go`, the cockpit lives in `cockpit.go` and the root dispatch.

Removing the live mode is **behaviour-preserving for every non-interactive caller**. `presentation/cli/top.go:727-738` already falls back to `writeTopSnapshotText` — the exact output `--snapshot` produces, from the same loader with the same pane caps — whenever stdout is not a TTY, and `docs/cli/README.md:832-834` documents that fallback. Only a human at a terminal sees a change.

Context, not grounds: of 282,710 recorded shell audits, 54 begin with `traceary ` and none of them invoke `sessions`, `top`, or `tui`. That audit only sees agent-driven shells, so it says nothing about a human typing `traceary` in their own terminal. The grounds are the pillar mismatch and the owner's decision, not the count.

## Scope (v0.34 — notice only, nothing is removed)

1. Emit the stderr notice from `traceary sessions` **only when it is about to open the live dashboard**, naming `traceary sessions --snapshot` and removal target v0.35.0. A non-TTY bare invocation is not notified: it is already taking the snapshot path and loses nothing.
2. Annotate the live-dashboard description in `docs/cli/README.md` / `.ja.md` and `docs/interactive/README.md` / `.ja.md`, and add the changelog pair entry — the three documentation updates `docs/cli-stability.md:133-139` requires.
3. Add the entry to the removal log with target v0.35.0.

Uses the shared deprecation helper landed by **#1688**; do not reimplement it.

### Exactly one notice per invocation

`traceary top` runs the same live runner (`presentation/cli/top.go:168-176,229`), and #1688 gives `top` its own notice. Without care a single `traceary top` emits two lines, which `docs/cli-stability.md:115` forbids. The `top` notice wins — it names the whole command's replacement, which subsumes the mode. Whichever of #1688 / #1765 lands second owns the test that pins exactly one line for each of `sessions` (TTY), `top` (TTY), and `top --snapshot`.

[...]

Issue #1690, opened 2026-08-07
v0.35.0: remove `traceary top`


duck8823:
Removal half of #1688, after one full minor of overlap per `docs/cli-stability.md`.

- Remove the command and its `--snapshot` / `--json` forms
- `traceary sessions` is the replacement and is unchanged
- Move the entry from the deprecation list to the historical removal log

Depends on the v0.34.0 notice shipping first.

Parent: #1606

Issue #1689, opened 2026-08-07
v0.34.0-49: deprecate memory admin graph and session label


duck8823:
## Rescoped 2026-08-10 — two of the four rows did not survive verification

Each remaining command depends on a table that has never held a row. This is an objective criterion, not a usage argument — low invocation counts alone are **not** grounds for removal. That rule is what removed two of the original four rows.

| command | evidence (live store, `PRAGMA query_only=1`) |
|---|---|
| `memory admin graph add` / `graph list` | `memory_edges` = **0 rows** since v0.9.0 |
| `session label` | set on **0 of 27,197** sessions |

### `memory admin graph`

Verified genuinely dead: `presentation/cli/memory_graph.go:155` is the only write path, nothing creates an edge automatically, and no MCP tool exposes the graph (`presentation/mcpserver/server.go:149-190` lists all eight).

The table itself survives the deprecation and stays referenced by `gc --target memory_edges` (`application/types/gc_target.go:16`) and by bundle export/import (`application/usecase/bundle_tables.go:85`). Whether the gc target is still earning its surface is #1761; coordinate so the two do not contradict each other.

### `session label`

The write path is dead, but the **read path is not** and this issue must say which side it is deprecating. The label is filtered on (`presentation/cli/session_list.go:101`), printed as a tabular column (`:127`), and emitted in JSON by `session list` (`:167`), `session tree` (`presentation/cli/session_tree.go:313`), and MCP (`presentation/mcpserver/output.go:101,142`). Deprecating only `session label` leaves a filter that can never match and a column that is always blank.

Decide and record in the issue before implementing: deprecate the whole label concept including `--label` and the output fields, or keep the read side and accept a permanently empty column. The MCP fields are `omitempty`, so they simply disappear either way.

## Removed from scope

- **`list --failures` → #1767.** The evidence row said "1 exit code out of 255,677". True, but the predicate is `ca.failed = 1 OR (ca.exit_code IS NOT NULL AND ca.exit_code != 0)` (`infrastructure/sqlite/sql/select_recent_events.sql:20`), and `failed = 1` matches **1,943** rows — still being written today. It fails this issue's own criterion.
[...]

Issue #1687, opened 2026-08-07
v0.34.0-47: deprecate the cockpit and the bare-traceary TTY default


duck8823:
## Rescoped 2026-08-10 — notice in v0.34, removal in v0.35

The original scope ("remove `traceary cockpit` and its aliases `tui` / `dashboard`") was wrong on two counts, and the owner has since decided both open questions.

**The command is not called `cockpit`.** `presentation/cli/cockpit.go:63-64` declares `Use: "tui"` with `Aliases: []string{"dashboard"}`. There is no `traceary cockpit`. The names a user can type are `tui` and `dashboard`, plus the bare `traceary` entrypoint.

**`docs/cli-stability.md` carrying no promise is not the whole picture.** It is true that neither `tui` nor `dashboard` appears in that file. But `docs/interactive/README.md:137` says "`traceary tui` remains a **stable** explicit entrypoint", and `README.md:60,187` / `docs/cli/README.md:100` call it an "explicit compatibility entrypoint" / "stable compatibility path". Removing it in the same minor that withdraws those words gives users no notice. **Owner decision: withdraw and give one minor of notice, matching #1688's flow for `top`.**

**The `sessions` interactive mode is a sibling, not part of this issue.** The owner asked whether the Sessions dashboard is needed either; the answer is that it goes the same way, but in its own issue — #1765. gpt-5.6-sol's objection is right: the cockpit and the Sessions dashboard have different implementations, different contracts, and different rollback boundaries. This issue touches `cockpit.go` and the root dispatch; #1765 touches the dashboard runner in `top.go`. One must be revertible without the other.

**One notice per invocation, across issues.** `traceary top` runs the same live dashboard runner as `sessions` (`presentation/cli/top.go:168-176,229`). Once #1688 gives `top` its own notice and #1765 gives the Sessions live mode one, a single `traceary top` would emit two — which `docs/cli-stability.md:115` forbids. Whichever of the three lands last owns the test that pins exactly one line.

## Scope (v0.34 — notice only, nothing is removed)

1. Reintroduce the shared deprecation helper **landed by #1688** and use it here (see Dependencies).
2. Withdraw the "stable explicit entrypoint" / "compatibility path" wording for `tui` in `README.md`, `docs/cli/README.md`, `docs/interactive/README.md` and their `.ja.md` pairs, recording the reason.
3. Emit a stderr deprecation notice, naming v0.35.0 and the replacement, from each of:
[...]

PR #1769, opened 2026-08-10
feat: withdraw the permanent-alias promise for traceary top

### LINKED ISSUE #1688: v0.34.0-48: withdraw the permanent-alias promise for `traceary top`
## Problem

`docs/cli-stability.md:52` states:

> `traceary top` is not deprecated in v0.19.0; it remains a **permanent** compatibility alias for every `traceary sessions` form. Removing it later would require the deprecation flow below.

`top` is a plain alias of `sessions`. Removing it requires withdrawing that promise first, then one full minor of overlap.

## Scope (v0.34 — notice only, no removal)

1. **Land the shared deprecation helper.** See below — this issue is the first user of it and therefore owns it.
2. Withdraw the permanence wording in `docs/cli-stability.md` and its `.ja.md` pair, with the reason recorded
3. Emit a stderr deprecation notice from `traceary top` naming `traceary sessions` and the removal version
4. Add the entry to the removal log with target v0.35.0, plus the two documentation updates the flow itself requires (`docs/cli-stability.md:133-139`): the CLI reference pair and the changelog pair

Removal is #1690 (v0.35.0).

### Why the helper lands here (added 2026-08-10)

There is **no deprecation machinery in the tree at all**: `presentation/cli/` contains zero uses of cobra's `Deprecated:` field and no helper function. The previous `applyMemoryAliasDeprecation` was removed when the v0.15 flat memory aliases went. Three Wave 4 issues need it (#1687, #1688, #1689), and under 1 issue = 1 branch = 1 PR exactly one of them has to land it. This one does, because it is the smallest single command and its notice is the least ambiguous.

The behaviour is already fully specified by `docs/cli-stability.md:96-116` and the helper must implement it verbatim:

- exactly one stderr line per invocation, in the documented English and `TRACEARY_LANG=ja` forms
- the notice names the canonical replacement **leaf** and the removal target version
- it goes to stderr, not stdout — the doc states explicitly at `:114` that cobra's built-in `Deprecated` field is unusable because it routes through stdout
- a parent group whose subcommand is the real entry point fires once, for the executing leaf

`#1687` and `#1689` consume the helper and must not reimplement it.

## Done when

- No document calls `top` permanent
- `traceary top` still works and prints the notice to stderr
- stdout and the `--json` envelope are unchanged

Parent: #1606

duck8823:
Closes #1688

## What

`docs/cli-stability.md:52` promised `traceary top` as a **permanent** compatibility alias for `traceary sessions`. v0.34 narrows Traceary to the 記録 / 記憶 pillars, and a second name for one command serves neither. This PR withdraws that promise and starts the one-minor deprecation window.

**Nothing is removed here.** `traceary top` keeps working, byte-for-byte, through v0.34.x. Removal is v0.35 (#1690).

## How

- New `presentation/cli/deprecation.go` — `applyCommandDeprecation` attaches the notice via cobra's `PreRunE`, not `PersistentPreRunE`, so it fires exactly once for the executing command (`docs/cli-stability.md:119`). Cobra's built-in `Deprecated:` field is unusable here because it routes through **stdout** (`docs/cli-stability.md:114`).
- `traceary top` now emits one stderr line per invocation, in the form the policy already specifies:
  - `DEPRECATED: this command is deprecated, use `traceary sessions` instead. Removal target: v0.35.`
  - JA under `TRACEARY_LANG=ja`.
- Docs: CLI reference, CHANGELOG (EN/JA), `docs/cli-stability.md` prose, a new **"Currently deprecated"** registry, `README`, `docs/interactive/README`, and `docs/operations/json-contract-tests.md` — which also still called the alias permanent.

The shared helper is reused by #1687, #1765, and #1689.

## Contract preservation

`traceary top --snapshot` and `--snapshot --json` stdout is asserted byte-identical to `traceary sessions`' — `TestRootCLI_TopSnapshotOutputMatchesSessions`. `sessions` is untouched by this PR, so equality with `sessions` is equality with today. This is the assertion that matters: the notice must not leak into stdout or perturb the frozen envelope (`docs/cli-stability.md:48`).

`traceary sessions` emits no notice.

## Tests

- `TestApplyCommandDeprecation` — EN/JA notice text, exactly one line, stdout unaffected, a pre-existing `PreRunE` still runs.
- `TestRootCLI_DeprecationNotice` — `top --snapshot`, `top --snapshot --json` emit exactly one line; `sessions --snapshot` emits none.
- `TestRootCLI_TopSnapshotOutputMatchesSessions` — text and JSON stdout equality.

## Verification

## Review notes

Implementation delegated to gpt-5.6-luna; reviewed by Claude Opus 5. Review found two docs gaps, fixed in `20f1eca2`:

1. `docs/operations/json-contract-tests.{md,ja.md}` still asserted "permanent compatibility" in the very doc that pins the snapshot contracts — the exact claim this issue withdraws.
2. The new "Currently deprecated" registry omitted the already-announced `traceary search --json` shape change (#1717), making it wrong on the day it was introduced.


duck8823:
**gpt-5.6-sol review** — 2 HIGH findings. One adopted with a different remedy, one rejected.

### 1. `presentation/cli/deprecation.go:12` — HIGH: `--help` and argument errors emit no notice — **adopted, different remedy** (`715a2594`)

The finding is factually correct. Cobra v1.10.2 `command.go` resolves `--help` at `:934` and runs `ValidateArgs` at `:968`, both before `PreRunE` at `:999`. So `traceary top --help`, `traceary top extra`, and `traceary top --not-a-flag` emit zero notices.

I did not change the hook. Moving the notice earlier is not possible within cobra: `PersistentPreRunE` runs at `:985`, still after both short-circuits, and cobra's own `Deprecated` field is already ruled out by `docs/cli-stability.md:114`. Every deprecation mechanism cobra offers has this property, so the policy sentence at `:103` was never satisfiable as literally written.

The honest fix is to state the boundary and pay for it:

- `docs/cli-stability.md` / `.ja.md` gain a notice rule: the notice fires when the command actually runs, and in exchange a deprecated command's `Short` / `Long` must name the deprecation and the removal target so `--help` still tells the caller. `top` already does.
- `TestRootCLI_DeprecationNoticeSkipsNonRunningPaths` pins both halves — no notice on either path, and help stdout must mention the removal version. The boundary is now asserted rather than accidental.

An argument-error path emitting no deprecation notice is also correct on its own terms: the command did not run.

### 2. `CHANGELOG.md:387,391` still say "permanent compatibility alias" — **rejected**

Those lines are inside `## [v0.19.0] - 2026-05-26` (`CHANGELOG.md:379`), a shipped release. The changelog records what each release announced, and its own header says it mirrors the GitHub release notes — which are immutable. Editing them would falsify the historical record, not withdraw a promise.

The promise that is actually in force lives in `docs/cli-stability.md:52`, the CLI reference, the READMEs, and `docs/operations/json-contract-tests.md`. All four are updated by this PR. The withdrawal is announced under `## [Unreleased]` → `### Deprecated`, which is where a change of position belongs.

### Non-blocking observations, all accepted as-is

Sol agreed with the stdout-identity argument (shared writer and `runTopNamed` untouched, existing goldens still pin the same fixtures) while noting the new test alone would not catch a future shared-writer change — the existing goldens cover that. Sol also confirmed the helper's current signature does not block #1687 / #1765 / #1689, and that this PR does not make #1765's exactly-one-notice guarantee harder, since `commandName` already distinguishes `top`.

Issue #1764, opened 2026-08-10
v0.35.0: remove the cockpit and make bare traceary print help


duck8823:
## Scope

Remove the cockpit, one minor after #1687 gives notice.

| surface | today | after |
|---|---|---|
| `traceary tui` | opens the cockpit | unknown command, exit 1 |
| `traceary dashboard` | alias of `tui` | unknown command, exit 1 |
| bare `traceary` in a TTY | opens the cockpit | prints help — identical to the non-TTY path |

The `sessions` live dashboard is **not** in scope: it is #1766, split out so the two are independently revertible. This issue touches `cockpit.go` and the root dispatch; that one touches the dashboard runner in `top.go`.

`memory inbox review` stays. It is a TUI, but it is a 記憶 pillar surface with no non-interactive equivalent.

## Why bare `traceary` changes the way it does

It already has the target behaviour on the non-TTY path, so this deletes a branch rather than inventing one. `presentation/cli/root.go:529-547` — `runRootDefault` opens the cockpit only when stdin/stdout are an interactive pair and otherwise calls `cmd.Help()`. Owner decision (2026-08-10): the TTY branch goes and help becomes the single answer.

## Done when

- `presentation/cli/cockpit*.go` (≈9,400 lines including tests) and the cockpit state file store are gone
- Root loses `--reset-state`, which only ever configured the cockpit. `--db-path` stays.
- `presentation/cli/tui/` stays — `top_dashboard.go` and `memory_inbox_review*.go` still use it
- The orphaned state file is handled: `~/.local/state/traceary/cockpit.json` (`cockpit_state_file.go:43-49`, XDG-aware with a `.traceary/` fallback) is removed on upgrade or documented as safe to delete
- `docs/operations/cockpit-ui-ux-design.{md,ja.md}` and `docs/operations/cockpit-dogfood.{md,ja.md}` are deleted or moved under a historical heading
- `docs/cli-stability.md` moves the three entries from the deprecation list to the removal log
- Every `--json` envelope is unchanged

## Capability that goes away

The cockpit is the only surface that tracks "new events since your last live-tail visit" — that is what the state file holds. Nothing else offers it, and nothing replaces it. This is a deliberate loss, recorded here rather than discovered at release.

## Depends on

#1687 (v0.34 notice). Related: #1766 removes the `sessions` live mode; #1690 removes the `top` **name**, which is orthogonal — after both, `top` is still an alias, just an alias of a text command.

Parent: #1693

PR #1770, opened 2026-08-10
feat: deprecate the cockpit TUI and the bare-traceary TTY default

### LINKED ISSUE #1687: v0.34.0-47: deprecate the cockpit and the bare-traceary TTY default

duck8823:
Closes #1687

## What

v0.34 gives one minor of deprecation notice to the cockpit TUI and to the bare-`traceary` TTY default. **Nothing is removed here**; removal is #1764 in v0.35.

Note the names: there is no `traceary cockpit`. `presentation/cli/cockpit.go:62` declares `Use: "tui"` with `Aliases: []string{"dashboard"}`.

## Why

The two pillars of v0.34 are 記録 (recording and consolidation) and 記憶 (memory reuse at wake time). A live operator cockpit serves neither: it is a watch surface, and watching is not one of the pillars. The surviving TUI is `memory inbox review`, which is memory work.

Owner decision on the bare-`traceary` default: in a TTY it becomes **help**, the same as non-TTY today.

## How

- `tui` / `dashboard` get the shared `applyCommandDeprecation` notice (replacement `traceary sessions --snapshot`, target `v0.35`). The alias is the same cobra command, so it is covered — asserted by test.
- The bare-`traceary` TTY default gets a **behaviour-change** notice, emitted only on the branch that is about to open the cockpit, before the alt-screen starts. Non-TTY bare `traceary` emits nothing new: it already prints help and loses nothing at removal.

### The spec gap this opens, and how it is closed

`traceary` is not being deprecated — only its default rendering in a TTY is, so "this command is deprecated" would be false. `docs/cli-stability.md` now states that the same single notice form also covers a default-behaviour change, with the behaviour named as the subject instead of the command path. No second message format is introduced: `applyCommandDeprecation` and the behaviour case share one implementation and one format string.

```
DEPRECATED: opening the cockpit from a bare `traceary` is deprecated, use `traceary --help` instead. Removal target: v0.35.
```

#1765 reuses this same form for the `sessions` interactive default.

## Docs

`tui` was called a "stable explicit entrypoint" / "stable compatibility path" in `README`, `docs/cli/README`, and `docs/interactive/README`. All replaced, EN and JA. Plus the three updates `docs/cli-stability.md:133-139` requires: CLI reference, CHANGELOG, and the "Currently deprecated" registry — `tui`, `dashboard`, and the bare-`traceary` TTY default, each targeting v0.35.0.

## Tests

- `tui` and `dashboard` each emit exactly one notice; help text carries the deprecation.
- Bare `traceary` with an interactive stdin/stdout pair emits exactly one behaviour-change notice (through the existing `cockpitInteractive` / `cockpitRunner` seam).
- **Bare `traceary` outside a TTY emits no notice and its stdout is byte-identical to `--help`.** This is the regression that matters most — scripts calling bare `traceary` see nothing change.
- Japanese forms under `TRACEARY_LANG=ja`.

## Verification

## Review notes

Implementation delegated to gpt-5.6-luna; reviewed by Claude Opus 5. Two findings, both fixed in `31b209b5`:

1. `applyBehaviorDeprecation` was a pure pass-through to `writeDeprecationNotice` with an identical signature — three functions where two suffice. Worse, `apply*` elsewhere in the file means "install a cobra hook", while this one wrote immediately. #1765 needs the immediate-write form too, so the name had to stop lying before a second caller arrived.
2. The docs called `traceary sessions --snapshot` a "replacement cockpit". It is a text snapshot, and #1765 deprecates the `sessions` live view as well, so **no cockpit survives v0.35**. Telling operators otherwise would be a promise this wave exists to stop making.


duck8823:
**gpt-5.6-sol review** — 1 HIGH. Rejected as stated, but it exposed a real contradiction in the policy, which is now fixed (`984acc57`).

### `presentation/cli/cockpit.go:65` — HIGH: non-TTY bare `traceary` stdout changed, 2114 → 2161 bytes — **rejected, with a policy fix**

The measurement is correct. Changing `tui`'s `Short` changes the subcommand listing that bare non-TTY `traceary` prints:

```diff
-  tui         Open the Traceary operator cockpit TUI
+  tui         Deprecated cockpit TUI; removed in v0.35. Use `traceary sessions --snapshot` instead.
```

This is not a compatibility violation, because the rule it is measured against cannot mean what it would have to mean here. `docs/cli-stability.md` requires a deprecated command's `Short` and `Long` to name the deprecation and removal target — that rule was added by #1688 precisely because cobra resolves `--help` before any hook, so help text is the only thing that tells a `--help` caller. If changing `Short` also violated the stdout byte-identity clause, then **no subcommand could ever be deprecated**: the flow would forbid its own required step. #1688 already shipped the same change for `top`, so main has the same property today.

The two clauses genuinely contradicted each other, and that was worth finding. Resolved by naming the boundary rather than by reverting:

> Help and usage text is deliberately outside this guarantee. The notice rule above *requires* a deprecated command to change its `Short` and `Long`, which changes the parent's command listing, so freezing help bytes would make the flow self-contradictory. Automation must not parse `--help`; it is the one output shape that announces deprecations rather than preserving them.

Added to `docs/cli-stability.md` and its Japanese pair.

On the test: pinning bare non-TTY stdout to a recorded golden would fail on every future subcommand `Short` change — noise, not signal. The invariant that carries meaning is the one asserted: **bare non-TTY stdout equals `--help` stdout**. That is the property #1764 relies on when it makes the TTY case print help too.

`applyDeprecation` was a single-caller indirection; folded into `applyCommandDeprecation`. Sol confirmed folding does not obstruct #1765 (conditional hook for the flag notice) or #1689 (`writeDeprecationNotice` for the behaviour notice).

Sol also confirmed by pseudo-TTY run: the notice is written before the alt-screen control sequence and survives on the restored screen; the full matrix emits one notice for TTY bare / `tui` / `dashboard`, zero for non-TTY bare and both help paths, and one plus exit `2` for non-TTY `tui` — the refusal being a separate error, not a second notice. EN/JA pairs match and no surviving stable/permanent claim for `tui` exists outside released changelog sections.

### Found in my own second pass

`presentation/cli/root.go:444` — the root `Long` still advertised the cockpit default and `traceary top` with no deprecation marker. A TTY user who runs `traceary --help` would have learned nothing about either. That is exactly the gap the help-text rule exists to close, so both are now marked.

### Unrelated CI failure

`TestMemoryHygieneScan_RevisionChurnReturnsPartialOnlyAfterDurationExhausts` failed once in `application/usecase`, a package this branch does not touch. It uses a 1 ms wall-clock budget; a re-run with no code change went green. Filed as #1771 rather than widened into this PR.

Issue #1768, opened 2026-08-10
v0.36-7: decide what a workspace conflict is for before touching store workspace-alias


duck8823:
## Why this exists

#1689 planned to deprecate `traceary store workspace-alias`, on the criterion that `session_workspace_aliases` has never held a row. It has not. But the row was removed from #1689 on 2026-08-10.

**Corrected 2026-08-10 (Wave 4 close).** The reason originally given here was that the deprecation flow could not express it — every notice had to name a canonical replacement, and there is none. **That reason no longer holds.** #1689 shipped the **no-replacement notice form** in the same wave, and `docs/cli-stability.md` now allows a notice with no successor provided the deprecation entry states the evidence that nothing is lost. The mechanical obstacle is gone.

What actually blocks retirement is everything below: the alias mechanism is not dormant, so the "nothing is lost" evidence cannot be produced. That is the question this issue exists to settle.

## What the alias mechanism still does

It is not dormant data. It participates in reads and in future writes:

- reclassifies a stored conflict as `explicit_alias` in the current report (`infrastructure/sqlite/workspace_identity_datasource.go:100-121`)
- suppresses reviewed-away pairs from conflict samples (`:202-217`)
- makes future observations land as `explicit_alias` directly when an alias exists (`infrastructure/sqlite/event_delivery_store.go:400-422`)
- `explicit_alias` is a documented relationship state (`docs/operations/workspace-identity-contract.md:106-119`)

`presentation/cli/store_workspace_alias.go:15-60,78-127` is the only public way to add, remove, or audit one. Removing the CLI would freeze the mechanism: existing aliases keep their meaning, but nobody can create, withdraw, or inspect one.

## The measurement that reframes the problem

Live store, `PRAGMA query_only=1`:

| | count |
|---|---|
| `observed_relationship = 'conflict'` rows | 14,684 |
| distinct `(session_id, workspace)` pairs | **147** |
| distinct sessions | 117 |
| distinct workspaces | 36 |

The row count is dominated by repetition, not by distinct problems — `codex` / `post_tool_use` alone contributes 13,664 rows from 16 distinct pairs. By source:

| client | hook | distinct pairs | rows |
|---|---|---|---|
| antigravity | stop_transcript | 91 | 104 |
| antigravity | stop | 82 | 93 |
| codex | user_prompt_submit | 20 | 381 |
| codex | stop | 19 | 386 |
| codex | post_tool_use | 16 | 13,664 |
[...]

Issue #1767, opened 2026-08-10
v0.36-6: establish what failed=1 means before deciding the future of list --failures


duck8823:
## Why this exists

#1689 listed `list --failures` as a command whose backing data has never existed, on the evidence "1 exit code out of 255,677". That evidence is real but only covers half the predicate, so the row was removed from #1689 on 2026-08-10. This issue holds what is actually true, which is a data-quality question rather than a dead-surface one.

## What the flag actually matches

`infrastructure/sqlite/sql/select_recent_events.sql:20`:

```sql
AND (? = 0 OR ca.failed = 1 OR (ca.exit_code IS NOT NULL AND ca.exit_code != 0))
```

The `failed` flag is set from a host's structured failure reason, with no exit code involved (`presentation/cli/hook_runtime.go:570-583`, `presentation/cli/hook_resolve.go:358-399`), and `infrastructure/sqlite/command_audit_store_test.go:426-430,512-529` pins `failed=true` with `exit_code=NULL` as an intended contract, not an accident.

Measured on the live store (`PRAGMA query_only=1`, 282,710 audits):

| | count |
|---|---|
| `exit_code IS NOT NULL` | 1 |
| `failed = 1` | 1,943 |

All 1,943 come from client `hook`, source hook `post_tool_use`:

| failure_reason | count | first | last |
|---|---|---|---|
| `unknown` | 1,054 | 2026-06-17 | 2026-07-21 |
| `host_error` | 889 | 2026-07-24 | 2026-08-10 |

So the surface is not dead, and it is not historical — `host_error` is still being written today.

## The actual question

`unknown` accounts for 54% of flagged failures and stops on 2026-07-21, exactly where `host_error` starts. That looks like a classification improvement landing mid-corpus rather than two coexisting kinds of failure. Worth establishing:

1. Is `unknown` the schema default (`command_audits.failure_reason DEFAULT 'unknown'`) leaking onto rows that were flagged before the classifier existed, or is it a genuine unclassified failure?
2. What changed on 2026-07-21/24, and does anything still produce `unknown` with `failed=1`?
3. Breakdown by host and source hook for `host_error` — is one integration generating most of it?
4. Are these failures the operator actually wants surfaced, or host-transport noise? That decides whether `list --failures` is a 記録 surface worth keeping or a diagnostic that belongs in `doctor`.
[...]

Issue #1691, opened 2026-08-07
v0.35.0: remove the commands deprecated in v0.34


duck8823:
Removal half of #1689.

- `memory admin graph add` / `graph list`
- `session label` — **scope resolved by #1689 / #1772: the whole label surface**, not just the write path. What carries a notice in v0.34 and must go here: `traceary session label`, `traceary session list --label`, the `LABEL` column in text output, and the `label` field in JSON output. Both halves are in the "Currently deprecated" registry in `docs/cli-stability.md`.

Move the entries to the historical removal log.

## Rescoped 2026-08-10 (Wave 4 spike)

Two entries left this issue with #1689:

- **`list --failures` → #1767.** Not dead. The predicate is `ca.failed = 1 OR (exit_code IS NOT NULL AND exit_code != 0)` (`infrastructure/sqlite/sql/select_recent_events.sql:20`) and `failed = 1` matches 1,943 rows on the live store, still being written today.
- **`store workspace-alias` → #1768.** No notice ships in v0.34, so no removal is due here.

  ⚠️ **The stated reason has since become false.** It read "cannot be deprecated under the current flow, which requires a canonical replacement." #1689 itself then added the **no-replacement notice form**, and `docs/cli-stability.md` now says: *"When a surface is removed without a successor, the notice says that there is no replacement instead of naming one; the deprecation entry must state the evidence that nothing is lost."* The flow can express it. What still blocks retirement is the other half of #1768's evidence — the alias mechanism is **live**, participating in report reclassification, conflict-sample suppression, and future `explicit_alias` promotion. Test retirements against *"can we show nothing is lost"*, not *"is there a successor."*

## On dropping the tables

The original text said "drop the now-unused tables in the same migration". Handle that as a separate decision, not a reflex:

- `memory_edges` is still referenced by `gc --target memory_edges` (`application/types/gc_target.go:16`) and by bundle export/import (`application/usecase/bundle_tables.go:85`). Dropping it means changing both, and #1761 is already asking whether that gc target earns its surface — settle that first.
- `sessions.label` is a column on a live table, not a table. Dropping it is a rewrite of `sessions` and is only in scope if #1689 chose to retire the whole label concept.

Parent: #1693

PR #1772, opened 2026-08-10
feat: deprecate memory admin graph and the session label surface

### LINKED ISSUE #1689: v0.34.0-49: deprecate memory admin graph and session label

duck8823:
Closes #1689

## What

Two surfaces depend on data that has never existed. v0.34 gives them one minor of notice; v0.35 removes them (#1691). **Nothing is removed here.**

| surface | evidence on the reference store |
|---|---|
| `memory admin graph add` / `graph list` | `memory_edges` = **0 rows** since v0.9.0 |
| `session label`, `session list --label`, the `LABEL` column, the `label` JSON field | set on **0 of 27,197** sessions |

## The label decision

Deprecating only the write command would leave a filter that can never match and a tabular column that is always blank. That is worse surface than either keeping the feature or removing it whole, and this wave exists to reduce surface. So all four label surfaces get the same notice and all four go in v0.35.

Nothing observable is lost: every JSON field is already `omitempty` (`presentation/cli/output.go:82,232,303`), so with zero labels anywhere the field never appears — removing it changes no emitted envelope. The `session list` table header is not part of the frozen contract (`docs/cli-stability.md:48` freezes the `doctor` / `sessions --snapshot --json` / `timeline --json` / `session tree --json` envelopes and the `session handoff` field labels; this is none of those).

The `sessions.label` **schema column** stays. Dropping it is a rewrite of a live table and belongs to #1691.

## The third notice form

There is no replacement for either surface — the features are going away, not moving. `traceary memory store list` lists memories, not edges, so naming it would be false, and `docs/cli-stability.md` requires a named replacement to be canonical.

So this adds the third and final case to the shared notice family, alongside the command form (#1688) and the behaviour-change form (#1687):

```
DEPRECATED: <subject> is deprecated with no replacement. Removal target: v0.35.
```

One implementation, one stderr path, one format family. The policy is amended to match: when a surface is removed without a successor, the notice says so, and the deprecation entry must state the evidence that nothing is lost.

## Notice placement

- `graph add` and `graph list` — the notice is on the **leaves**, not the `graph` parent, so it fires once for the executing leaf.
- `session label` — on the command.
- `session list --label` — only when the flag is actually passed (`cmd.Flags().Changed("label")`). A plain `session list` emits nothing.

## Tests

Exactly one notice for each of `graph add`, `graph list`, `session label`, and `session list --label foo`; none for `session list`; stdout byte-identical throughout; Japanese forms under `TRACEARY_LANG=ja`.

## Verification

## Review notes

Implementation delegated to gpt-5.6-luna; reviewed by Claude Opus 5. Two findings, fixed in the follow-up commit:

1. `presentation/cli/session_list.go:104` — the `--label` flag's own help text still read "filter by label" with no deprecation marker. `session list --help` would have told the caller nothing, which is exactly the gap the help-text rule added earlier in this wave exists to close.
2. Call sites passed a bare `""` to mean "no replacement", which reads as an omission rather than a decision. Now a named `noReplacement` constant, and the two message branches no longer duplicate the stderr handling.

Out of scope and deliberately untouched: `memory_edges` table, `gc --target memory_edges`, bundle export/import, the `sessions.label` column, and `presentation/mcpserver/`. `list --failures` and `store workspace-alias` were split out to #1767 and #1768 during the Wave 4 spike.


duck8823:
**gpt-5.6-sol review** — 1 HIGH, adopted. It found a real defect that my review missed.

### `presentation/cli/memory_graph.go:58` — HIGH: a notice fires for an invocation that never runs — **fixed** (`52bdda27`)

`graph add` has required flags (`--to`, `--relation`). Cobra validates required flags in `command.go` **after** `PreRunE` and before `RunE`, so `traceary memory admin graph add memory-a` emitted a notice and then failed with a required-flag error. Sol measured it: exit code 1, stdout 0 bytes, one notice. That breaks the rule this wave itself added — the notice fires when the command actually runs.

I had checked this ordering while reviewing #1688 and dismissed it, because `top` has no required flags. `graph add` does. The dismissal was the error, not the rule.

The fix is not a special case for `graph add`. `applyCommandDeprecation` now attaches the notice to the **run step** instead of `PreRunE`, which makes the policy sentence literally true for every short-circuit cobra has — help resolution, argument validation, and required-flag validation all precede it. The rule text in `docs/cli-stability.md` and its Japanese pair now say so explicitly.

`TestRootCLI_DeprecationNoticeSkipsRequiredFlagFailures` pins it, for a missing `--to` and a missing `--relation`. I verified the test has teeth by reverting the helper to `PreRunE`: both cases fail with the notice present.

`docs/cli/README.md:660,672` and the Japanese pair stated "no replacement" without the evidence that the CHANGELOG and the registry both carry. Added: the reference store has zero `memory_edges` rows.

### Confirmed by Sol's own measurement

One notice each for `graph add`, `graph list`, `session label`, `session list --label foo`; zero for `memory admin graph` and plain `session list`; zero on every `--help` path, with the deprecation visible in the help body or flag description instead. `session list --label ""` correctly emits one — `Changed` is the right predicate and no environment binding exists for the flag. `session list --label foo --offset -1` emits the notice before any stdout write.

Against the same zero-label database, compared with `main`: text output 183 bytes identical, JSON output 247 bytes identical, zero `"label"` keys emitted, and `presentation/cli/output.go:303` confirmed `json:"label,omitempty"`. The `noReplacement` sentinel is sound for its three callers. None of the out-of-scope surfaces — memory-edge usecase, SQLite datasource and schema, GC, bundle, MCP server — were touched.

Issue #1766, opened 2026-08-10
v0.35.0: remove the sessions live dashboard


duck8823:
## Scope

Remove the `sessions` live dashboard, one minor after #1765 gives notice. `traceary sessions` becomes a plain text command; `traceary sessions --snapshot [--json]` is unchanged.

Split from #1764 (cockpit removal) so the two are independently revertible: this issue touches the dashboard runner in `top.go`, that one touches `cockpit.go` and the root dispatch.

## Why this is a no-op for scripts

`presentation/cli/top.go:727-738` already falls back to `writeTopSnapshotText` for non-TTY callers — the same writer, loader, and pane caps `--snapshot` uses (`:198-215`) — and `docs/cli/README.md:832-834` documents it. Deleting the live branch makes bare `sessions` byte-identical to `sessions --snapshot` for every caller. Only a human at a terminal sees a change.

## Done when

- `top_dashboard.go` and `top_dashboard_internal_test.go` (≈2,600 lines) are gone
- `top_data.go` and `top_json.go` stay — they feed the snapshot
- `presentation/cli/tui/` stays if `memory_inbox_review*.go` still uses it; if #1764 has also landed, check whether anything but memory inbox review imports it and shrink accordingly
- `docs/cli-stability.md` moves the entry from the deprecation list to the removal log
- The deprecation notice from #1765 is removed along with the mode it described

## Must not change

- `sessions --snapshot` / `--snapshot --json` and `top --snapshot` / `--snapshot --json` — frozen contract (`docs/cli-stability.md:48`)
- `sessions --json` without `--snapshot` stays an error (`presentation/cli/top.go:217-218`)
- `--profile ai` still requires `--snapshot --json` (`presentation/cli/top.go:220-227`)
- snapshot text bytes, exit codes, and the `--idle` / `--limit` / `--allow-stale` / workspace / client / agent filters

## Depends on

#1765 (v0.34 notice). Related: #1764, #1690.

Parent: #1693

PR #1773, opened 2026-08-10
feat: deprecate the sessions interactive dashboard default

### LINKED ISSUE #1765: v0.34.0-50: deprecate the sessions interactive dashboard and give notice

duck8823:
Closes #1765

## What

`traceary sessions` with no `--snapshot` opens a five-pane live Bubble Tea dashboard in a TTY. v0.34 gives one minor of notice that this **default** is going away; v0.35 removes it (#1766) and `sessions` becomes a plain text command. **Nothing is removed here.**

`sessions` itself is **not** deprecated — only its interactive rendering. It is the surviving command, and the replacement `traceary top` points at.

## Why

A live watch surface serves neither of v0.34's two pillars, 記録 and 記憶. The surviving TUI is `memory inbox review`, which is memory work. Everything the dashboard shows is already available as `sessions --snapshot [--json]`, which is what scripts and non-TTY callers get today.

## How

`presentation/cli/top.go` — `runTopTUI` already falls back to `writeTopSnapshotText` for non-TTY callers. The notice is emitted only on the branch that is about to run the live dashboard, immediately before the alt-screen starts, so the terminal restores it when the user quits. A non-TTY invocation emits nothing new: it is already taking the snapshot path and loses nothing at removal.

Replacement `traceary sessions --snapshot`, removal target `v0.35`, using the behaviour-change notice form added by #1687.

## Exactly one notice for `traceary top`

This is the part that had to be right. `traceary top` runs the **same** live runner, and #1688 already attached a command-deprecation notice to it. Without care, one `traceary top` in a TTY would emit two lines, which `docs/cli-stability.md` forbids.

The `top` notice wins — it names the replacement for the whole command, which subsumes the mode. The exclusion now lives in a named predicate that carries its own reason, so it does not get deleted as a stray condition before `top` itself goes in v0.35:

```go
func interactiveDashboardNoticeApplies(commandName string) bool {
	return commandName != "top"
}
```

## What must not change, now or at removal

Pinned as tests, not just as care:

- `sessions --json` without `--snapshot` stays an error; `--profile ai` still requires `--snapshot --json`
- `sessions --snapshot` and `--snapshot --json` stdout is byte-identical — frozen contract (`docs/cli-stability.md:48`)
- **non-TTY bare `sessions` stdout is byte-identical to `sessions --snapshot`, with empty stderr.** #1766 deletes the live branch on the strength of exactly this property, so it is pinned here rather than assumed.

## Verification

## Review notes

Implementation delegated to gpt-5.6-luna; reviewed by Claude Opus 5. Two findings, fixed in the follow-up commit:

1. The `sessions` `Short` began "Deprecated live Sessions dashboard", so `traceary --help` listed `sessions` as going away. That is backwards — `sessions` is the replacement `top` points at, and only its interactive default is deprecated. A user scanning the command list would have drawn exactly the wrong conclusion, and the two Shorts contradicted each other.
2. The `top` exclusion was a bare `commandName != "top"` with no explanation. Named, documented, and pinned by a test. The TTY branch reads the real stdin/stdout and has no injection seam, so the decision is the testable part; the non-TTY path is covered by the byte-identity test above.


duck8823:
**gpt-5.6-sol review** — **no CRITICAL or HIGH findings.**

Sol verified the critical property under a real pseudo-TTY, launching the built binary and quitting with `q`:

| invocation | notices |
|---|---:|
| `sessions` in a TTY | 1 — the mode notice |
| `top` in a TTY | 1 — the command notice only |
| `top --snapshot` | 1 |
| `sessions --snapshot` | 0 |
| non-TTY `sessions` | 0 |

`interactiveDashboardNoticeApplies("top") == false` is what prevents the double notice.

**Alt-screen ordering**, from the captured PTY control sequences: `notice → ?1049h → dashboard paint → ?1049l` for both `sessions` and `top`. The notice is written to the main screen before the alt-screen opens, is never swallowed by it, never lands after the dashboard paints, and survives on the restored screen after `q`.

**Byte identity**, building `main` and this branch separately against the same SQLite store: `sessions --snapshot`, `sessions --snapshot --json`, and non-TTY bare `sessions` all produce identical stdout with exit 0, and no new notice on any of them. The frozen `--snapshot --json` envelope did not move.

**On threading `*cobra.Command`** through `runTop` / `runSessions` / `runTopNamed` / `runTopTUI`: acceptable. It stays inside the presentation layer, the repository already has precedent (`runRootDefault`, `runSessionRefine`, `runStoreCapacity`), and `writeDeprecationNotice` is itself cobra-dependent, so narrowing to an `io.Writer` would not remove the dependency — only move it. Not worth the churn.

`presentation/cli/top.go:76` — `top`'s `Long` told new interactive callers to prefer `traceary sessions`, which is the very rendering this PR deprecates. Pointing migrating users at a second deprecated surface defeats the purpose of the notice. Now points at `--snapshot`, the only form that survives v0.35.

### Not adopted

Sol confirmed there is no injection seam for a genuine end-to-end TTY test of the combined command-notice plus mode-notice result — `withCockpitRuntimeForTest` is cockpit-specific and does not transfer, so a real test would need an external PTY subprocess or a new interactive/runner seam for `top`. He explicitly does not ask for that churn, given the manual PTY verification above and the unit test pinning the decision. Agreed: adding a seam for a surface being deleted in v0.35 would be work with a two-month lifespan.

Issue #1606, opened 2026-08-01
v0.34.0: consolidate and reuse — the two pillars


duck8823:
# v0.34.0 — Consolidate and reuse

A restart, not an increment. Traceary narrows to **two pillars**, both of which must work **reliably and automatically**.

1. **記録** — capture already works. Keeping whole conversations bloats, so: summarise, compress, evict from the index.
2. **記憶** — the same, plus **automatic use of the summary**.

The second pillar is the "AI should sleep too" idea: consolidation happens without being asked, forgetting is a feature, and by the time you wake the selection is already done.

Spec: [`.ai/spec/1606.md`](../blob/main/.ai/spec/1606.md) (landed by #1671).

## Where we actually are (measured 2026-08-07, live store)

24.71 GiB · 534,086 events · 26,728 sessions · 2.5 months.

| | |
|---|---|
| capture | works — 99.99% via hooks |
| growth | **+9.9 GiB/month**; five reduction mechanisms exist, none enabled |
| index | **15.83 GiB (64.1%)** — trigram over the entire corpus |
| duplication | **57.7% of `events.body` is an exact duplicate** (Kimi transcripts: 96.6%) |
| summaries | **4 of 26,728** sessions have one |
| automatic use | **no wiring at all** — `additionalContext` appears zero times in the codebase |

Speeding up search by indexing everything is what turned a 4 GiB problem into a 24 GiB one.

## What changed from the original plan

| was | now | why |
|---|---|---|
| hot/cold segmented history, external storage | withdrawn | external storage was never wanted |
| refine at `SessionStart` | **refine at `stop`** | handing raw material at wake time pollutes the whole session |
| consolidate at compact | **compact is observation-only on every host** | verified against Claude / Codex / Kimi primary docs |
| discard bodies to save disk | **fix duplication and index tiering** | discardable transcripts are only 10% of the store |
| attest the audit record including output | **attest instruction + command identity only** | `output_text` is 1.47 GiB at 6,153 B/row and proves nothing extra |
| attestation chain in v0.34 | **moved to v0.35** (#1677, #1678) | serves neither pillar; v0.34 stays small |

### Wave 0 — fix the premise
Every threshold and budget below is sized on byte counts that are currently 57.7% duplicate.

- #1681 make transcript recording idempotent per turn — the extraction is correct; nothing prevents the same unchanged turn being re-recorded, and Kimi's `Stop` fires ~24x per completed turn
[...]

Issue #1083, opened 2026-05-25
v0.19-17: promote Sessions dashboard and sessions command


duck8823:
## Scope
Promote the session dashboard as the primary Sessions surface: rename the cockpit Top tab to Sessions, remove the placeholder Sessions tab, and add a `traceary sessions` command while keeping `traceary top` as a compatibility command.

Parent: #1052

## Acceptance criteria
- Cockpit tabs are `Tail | Sessions | Memory | Settings`; there is no separate placeholder Sessions tab.
- The former Top cockpit surface is labeled Sessions and keeps the dashboard/detail behavior.
- Add `traceary sessions` with the same dashboard, filters, `--snapshot`, and `--json` behavior as `traceary top`.
- `traceary top` remains available for compatibility; help copy points new users to `traceary sessions --snapshot [--json]`.
- Session rows include an operator-friendly `name="..."` derived from label, summary, or stable workspace/agent fallback before raw metadata.
- Dynamic dashboard rendering still fills terminal height for both `sessions` and compatibility `top`.

## Non-goals
- Do not change the JSON schema emitted by `top --snapshot --json` or `sessions --snapshot --json`.
- Do not remove `traceary session ...` lifecycle subcommands.

## Validation
- `go test ./presentation/cli -run 'Top|Sessions|Cockpit|Root'`
- `go test ./...`
- `go tool golangci-lint run`

Issue #1060, opened 2026-05-24
v0.19-8: make bare traceary open the Tail-first TUI


duck8823:
## Goal
Make the default interactive entrypoint Tail-first, while preserving non-interactive and discoverability contracts.

## Requirements
- Running `traceary` with no subcommand in an interactive TTY opens the TUI.
- Running `traceary` with no subcommand in a non-TTY must not corrupt output or hang; keep help/fallback behavior explicit and tested.
- Keep `traceary tui` as a compatibility alias/explicit entrypoint.
- Preserve `--help` behavior.
- Document the new default in CLI docs and release notes.

## Acceptance criteria
- Interactive no-args path starts the Tail-first TUI.
- Non-interactive no-args path is covered by a test.
- Existing subcommands and help output remain compatible.

Parent: #1052

duck8823:
Closed by merged PR #1067.

PR #1774, opened 2026-08-10
docs: record what wave 4 settled about deprecation notices

duck8823:
Records Wave 4 in the spec. Refs #1606 — **closes nothing**; all four Wave 4 sub-issues (#1688 / #1687 / #1689 / #1765) are already closed by their own PRs, and #1606 is the parent, which stays open until the tagged release.

## What

One row (D15) and one prose section in `.ai/spec/1606.md`. Spec-only; no code, no docs under `docs/`.

## Why this and not just the PR bodies

The four Wave 4 PRs each record their own issue. What none of them records is the thing that outlived them: the notice flow changed shape three times while Wave 4 was being implemented, and v0.35's four removal issues (#1764 / #1766 / #1690 / #1691) inherit the result.

- **Three notice forms**, one implementation, one stderr path — command, behaviour-change, no-replacement.
- **The notice is attached to the run step**, not a pre-run hook. This came from a measured defect, not a preference: cobra v1.10.2 validates required flags at `command.go:1007`, *after* `PreRunE` at `:999`, so `traceary memory admin graph add memory-a` emitted a notice and then exited 1. `--help` (`:934`) and `ValidateArgs` (`:968`) short-circuit the same way. I had seen that ordering during the #1688 review and waved it off because `top` has no required flags; `graph add` has two.
- **Help text is exempt from the stdout byte-identity guarantee.** Requiring a deprecated command to change its `Short` is what makes `--help` still useful once the notice only fires on real runs — but changing `Short` moves the parent's command listing. Freezing help bytes would have made the flow unable to deprecate anything.

D15 also records the two `gpt-5.6-sol` HIGH findings I rejected and why, so neither gets re-raised in v0.35: the `CHANGELOG.md:387,391` "permanent" lines sit inside the shipped `## [v0.19.0]` section and rewriting them would falsify the record, and the `tui` `Short` byte-count change is the flow working, not violating it.

## Verification

Spec-only change; `git show --stat` is one file. No build or test surface is touched.

### DIFF
--- a/.ai/spec/1606.md
+++ b/.ai/spec/1606.md
@@ -386,6 +386,7 @@ stop(畳む) ── やりとり ── stop(スルー) ── compact
 | D12 | compact 到達時に未精製範囲を stop 圧力から**切り離して gc 所有へ移し、圧力をリセット**する。compact hook を持たない Antigravity は対象外（残存リスクを受け入れ） | compact 後の素材は文脈にないので、以後 stop を繰り返しても畳めない |
 | D13 | 対話 surface は `memory inbox review` **だけ**残す。cockpit・`sessions` live dashboard・bare `traceary` の TTY 既定は **v0.34 告知 / v0.35 削除**。削除後の bare `traceary` は TTY でも help | D10 は「即削除」を前提にしていたが、`docs/cli-stability.md` に公約が無いだけで README / docs/interactive が `tui` を stable と明記していた。`top` (D11) と同じ扱いに揃える |
 | D14 | 「backing table が空」を根拠にした退役は、**述語全体**と**deprecation 契約**の両方を満たす場合に限る | Wave 4 spike で 4 件中 2 件が脱落した。`list --failures` は `failed = 1` で 1,943 件ヒットする（exit code だけを見た evidence が誤り）。`store workspace-alias` は canonical replacement が無く、`docs/cli-stability.md:98-115` の notice 形式を満たせない |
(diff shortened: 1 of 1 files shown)


duck8823:
**gpt-5.6-sol review** — 3 HIGH, all three real, all three fixed in `4fd70b52`.

Sol reviewed the section against the merged history and the current tree rather than against plausibility, which is the right test for a spec section whose whole job is to be accurate.

### 1. The HIGH count was wrong — **fixed**

I wrote "3 findings, 2 rejected, only the required-flag one adopted." The actual tally, from the review comments on the four PRs: **4 HIGH, 2 adopted, 2 rejected.**

| PR | finding | outcome |
|---|---|---|
| #1769 | `--help` / argument errors emit no notice | **adopted**, different remedy |
| #1769 | CHANGELOG still says permanent | rejected |
| #1770 | `tui` `Short` change moved bare non-TTY stdout | rejected — but produced the help-text exemption |
| #1772 | required-flag failure emits a notice | **adopted** |

Sol's point about *why* this matters is the sharp one: undercounting erased the origin of the `Short` / `Long` rule. That rule is the part of the policy v0.35 is most likely to question, and the spec no longer said where it came from. The section now carries the table.

He also called the `tui` entry self-serving as written, and he is right. The measurement was correct and, under the policy as it stood, so was the violation. Only the conclusion — revert the code — was wrong. That is now what the section says.

### 2. Stale `CHANGELOG.md` line numbers — **fixed**

`:387,391` with the section header at `:379` was true when #1769 was reviewed. Wave 4's own changelog entries moved them; they are now `:390,394` under a header at `:382`, and `:387` / `:391` are unrelated Memory entries. A reader following the citation lands somewhere else.

Fixed by naming the release section and its two entries (Sessions dashboard #1083, bare CLI #1060) rather than line numbers. A file this section keeps appending to cannot carry a durable line citation.

### 3. D15 silently invalidated part of D14 — **fixed**

The one I would not have caught. D14 gave two reasons `store workspace-alias` cannot be retired: the table is not really dead, **and** the deprecation flow requires naming a canonical replacement, which the alias does not have. #1689 then added the no-replacement form and `docs/cli-stability.md:123` now explicitly allows a notice with no successor — so the second reason stopped being true, in this very wave, without D14 being revised.

Left alone, v0.35 would test retirements against "is there a successor" when the actual test is now "can we show nothing is lost." D15 states the revision, D14's row is annotated, and the surviving reason (measured in #1768: the alias is live in report reclassification, conflict-sample suppression, and future `explicit_alias` promotion) is called out as the one that still stands.

### Verification

Spec-only; no build surface. CI on the first commit: Test / Lint / Docs / Integrations all green.

Issue #1676, opened 2026-08-07
v0.34.0-36: define the gc state machine for eviction, discard, and compression


duck8823:
## Scope

Today gc conflates decisions that change for different reasons. Separate them.

| separation | when | what is lost | what remains |
|---|---|---|---|
| **evict from index** | over the window budget | full-text search | session-tier search, body, statistics |
| **discard body** | folded **and** non-attested | the text | bytes, timestamps, counts |

```
body_state:  present --(over budget)----------> (index only; body stays)
                     --(folded, non-attested)--> discarded
```

## Compression is deferred out of v0.34

The original scope had a third branch — `attested -> compressed, nothing is lost`. **v0.34 has no way to evaluate it.**

- The attestation chain was withdrawn from v0.34 (see the spec's "撤回した設計"); `#1677` / `#1678` are v0.35 work under #1693
- The only `attested` concept in the code today is `OneShotRepairEvidenceOperatorAttested` (`application/types/one_shot_repair.go:20`), which attests a process exit during repair. It says nothing about whether an event body is attested

Shipping a three-state machine whose third state can never be reached would bake a dead branch into the gc contract and into `body_state`'s CHECK constraint. So v0.34 ships two states, and compression lands with attestation in v0.35.

`body_state` must therefore be defined so a later value can be added without rewriting the column — say so in the migration.

## Never removed by gc

- `events` skeleton (timestamps, kinds, body byte counts, counts)
- `session_refinements`
- the session-tier index
- `prompt` bodies (human instructions) — never discarded
- `command_text` / `input_text` — never discarded

## Removed once folded

- `transcript` bodies (2.28 GiB measured)
- `command_outputs` (1.36 GiB measured, #1675)

## Done when

- An unfolded transcript survives gc
- A folded transcript is discarded and every statistic is unchanged
- Index eviction leaves the body readable through `show`
- A `prompt` body is never discarded by any gc target
- Adding a third `body_state` value in v0.35 needs no data migration

Depends on #1685 (`body_state` must exist). Compression follow-up tracked under #1693.

Spec: [`.ai/spec/1606.md`](../blob/main/.ai/spec/1606.md) (landed by #1671)

Parent: #1606

PR #1776, opened 2026-08-10
fix: guard session refinement lower bounds

### LINKED ISSUE #1758: v0.34.0-x: session refinement coverage can regress at its lower bound

## Problem

`SaveIfAdvances` writes all of `covers_from_event_id`, `covers_to_event_id`, `summary`, `keywords`, `produced_by`, `produced_at` and `degraded` from `excluded`, but the guard in `infrastructure/sqlite/sql/upsert_session_refinement.sql` only checks that `covers_to_event_id` strictly advances under `(ts_norm(created_at), id)` order. Nothing requires the *lower* bound not to move.

The table comment states the intended invariant plainly:

> a newer refinement strictly subsumes its predecessor (coverage only advances; re-summarisation is cumulative)

The SQL enforces half of it.

## Failure scenario

Session S has generation 1 covering `[e1, e3]`. A generation 2 write arrives covering `[e2, e4]`. `covers_to` advances from e3 to e4, so the guard passes, and `covers_from` is overwritten from e1 to e2.

`e1` is now outside the stored range. It was covered a moment ago and is not covered now — coverage went backwards, and the row records no trace of the range it used to hold.

## Why it matters more after #1676

`gc --target events` discards a transcript body only if a refinement covers it. Coverage is now the sole authority for an irreversible discard, and it is also the thing that keeps a body *safe*. A lower bound that can regress means an event can be un-covered by an ordinary write — silently re-arming nothing, but making the fold's promise ("everything in this range is summarised") false for a range the summary text still describes.

## Suggested fix

Extend the `WHERE` guard on the `DO UPDATE` branch to require the lower bound not to regress, mirroring the existing `covers_to` comparison — i.e. `excluded.covers_from_event_id` must be at or before `session_refinements.covers_from_event_id` in `(ts_norm(created_at), id)` order. The existing comparison sub-select can be parameterised rather than duplicated.

Add a test that a write which advances `covers_to` while moving `covers_from` forward is rejected, and one that a write advancing both ends of the range in the subsuming direction is accepted.

## Scope

Guard + test only. No new column, command or migration.

duck8823:
Closes #1758

## What

`infrastructure/sqlite/sql/upsert_session_refinement.sql` states this invariant in its own header:

> a newer refinement strictly subsumes its predecessor (coverage only advances; re-summarisation is cumulative)

It enforced half of it. The `DO UPDATE` guard compared only `covers_to_event_id` under canonical `(ts_norm(created_at), id)` order; `covers_from_event_id` was written unconditionally from `excluded`. A write covering `[e2, e4]` over a stored `[e1, e3]` passed, and `e1` silently left the covered range.

This adds the mirrored comparison for the lower bound: `excluded.covers_from_event_id` must be **at or before** the stored one. Equality is accepted — the ordinary re-summarisation keeps the same start and extends the end.

## Why it is worth the pre-gate slot

After #1676, `gc --target events` discards a transcript body **only if a refinement covers it**. Coverage is now the sole authority for an irreversible discard and simultaneously the thing that keeps a body safe.

## Honest scope: this is defence at the boundary, not a live data-loss bug

I checked the production path before accepting the change rather than after. `application/usecase/session_refinement_usecase_impl.go:141` passes `existing.CoversFromEventID()` through on every supersede, with the comment "Coverage only advances: keep the earlier covers_from". **No current caller can trigger the defect.**

That does not make the fix pointless — `SaveIfAdvances` is a domain repository contract whose SQL promises an invariant it did not enforce, and the guarantee now has to hold for any future caller as well as this one — but the PR should not claim to be fixing observed corruption. It is not.

The same check also confirmed the fix cannot reject a legitimate write, which was the real risk: had a producer legitimately computed a later start, this guard would have thrown away the new summary text entirely. It cannot, because the usecase never moves the lower bound.

## Implementation notes

- The comparison is a second explicit scalar `CASE` rather than a CTE. gpt-5.6-luna's reasoning, which I accept: there are no new placeholders either way, and the explicit form keeps the fail-closed behaviour legible next to the existing one.
- **No `?` was added or moved.** `session_refinement_datasource.go:85-102` binds arguments positionally under a comment describing the layout; a silent placeholder shift there writes the wrong columns, so this was the highest-risk part of the change and it was avoided entirely.
- Missing event rows yield NULL and therefore fail closed, matching the existing upper-bound behaviour. `gc --target events` blanks bodies with `UPDATE` and does not delete rows — but **`store dedupe` does**: `content_event_dedupe_datasource.go:801` runs `DELETE FROM events WHERE id = ?` after archiving. So the NULL branch is reachable, and reaching it freezes that session's refinement permanently. That is pre-existing behaviour of the upper-bound guard, not something this PR introduces; filed as **#1777** rather than widened into here.

## Verification

Five subtests, table-driven: forward lower bound rejected, widening accepted, unchanged accepted, and the two equal-timestamp tie-break cases.

I re-ran the new test against the pre-fix SQL myself rather than taking the implementer's word:

```
--- FAIL: .../rejects_advancing_upper_bound_with_a_forward_lower_bound
    SaveIfAdvances() written mismatch (-want +got):
      bool(
    - 	false,
    + 	true,
      )
```

The tie-break cases came out of review: the original three all passed with the guard's `<=` mutated to `>=`, because none of them placed two events at the same instant. Confirmed the new pair closes that by mutating the SQL and watching exactly those two fail:

```
--- FAIL: .../accepts_an_equal_timestamp_with_a_lower_id
--- FAIL: .../rejects_an_equal_timestamp_with_a_higher_id
```

## Review notes

### DIFF
--- a/infrastructure/sqlite/session_refinement_datasource.go
+++ b/infrastructure/sqlite/session_refinement_datasource.go
@@ -62,8 +62,10 @@ func (d *SessionRefinementDatasource) FindBySessionID(
 }
 
 // SaveIfAdvances stores refinement only when the persisted row is still at
-// expectedGeneration (0 means "no row yet") and the incoming covers_to
-// strictly advances coverage under canonical event order.
+// expectedGeneration (0 means "no row yet") and the incoming range subsumes the
+// stored one under canonical event order: covers_to strictly advances, and
+// covers_from does not move forward. Coverage authorises an irreversible body
+// discard, so it may widen but never shrink.
 func (d *SessionRefinementDatasource) SaveIfAdvances(
 	ctx context.Context,
 	refinement *model.SessionRefinement,
--- a/infrastructure/sqlite/session_refinement_datasource_test.go
+++ b/infrastructure/sqlite/session_refinement_datasource_test.go
@@ -475,6 +475,21 @@ func TestSessionRefinementDatasource_SaveIfAdvances_LowerBoundGuard(t *testing.T
 			wantWritten:    true,
 			wantStoredFrom: "evt-1",
 		},
+		// The tie-break is the part of the comparison a mutation would slip
+		// through: with equal timestamps the guard falls back to id order, and
+		// both cases below pass under either <= or >= unless they are paired.
+		{
+			name:           "accepts an equal timestamp with a lower id",
+			coversFrom:     "evt-0-tie",
+			wantWritten:    true,
+			wantStoredFrom: "evt-0-tie",
+		},
+		{
+			name:           "rejects an equal timestamp with a higher id",
+			coversFrom:     "evt-1-zz",
+			wantWritten:    false,
+			wantStoredFrom: "evt-1",
+		},
 	}
 
 	for _, tt := range tests {
@@ -484,6 +499,9 @@ func TestSessionRefinementDatasource_SaveIfAdvances_LowerBoundGuard(t *testing.T
 				{id: "evt-1", at: time.Date(2026, 8, 1, 10, 0, 0, 0, time.UTC)},
 				{id: "evt-2", at: time.Date(2026, 8, 1, 11, 0, 0, 0, time.UTC)},
 				{id: "evt-3", at: time.Date(2026, 8, 1, 12, 0, 0, 0, time.UTC)},
+				// Same instant as evt-1, sorting either side of it by id.
+				{id: "evt-0-tie", at: time.Date(2026, 8, 1, 10, 0, 0, 0, time.UTC)},
+				{id: "evt-1-zz", at: time.Date(2026, 8, 1, 10, 0, 0, 0, time.UTC)},
 			})
 			seed, err := model.NewSessionRefinement(
 				sessionID, 1, "evt-1", "evt-2", "seed summary", "", "agent",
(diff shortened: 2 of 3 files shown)


duck8823:
## gpt-5.6-sol independent review: no CRITICAL, no HIGH

Reviewed read-only against `dfef93f3`. Three non-blocking observations; two adopted in `ff6d967e`, one escalated to its own issue.

**Adopted — the tie-break was untested.** sol: 「異なるIDで同一タイムスタンプの場合の tie-break をテストしていません。現状は `<=` を誤って `>=` にしても3ケースすべて通ります。」 This is the exact branch I had flagged as the failure mode I cared most about, and the tests did not reach it — none of the three placed two events at the same instant. Added the paired equal-timestamp cases and verified by mutating the SQL that both fail together; either one alone still passes under the wrong operator.

**Adopted — the contract was not stated where callers read it.** `SaveIfAdvances` now says what the guard promises and why: coverage may widen but never shrink, because it authorises the irreversible discard in #1676.

**Escalated — my NULL-branch claim in the original body was wrong.** I had written that the fail-closed path is unreachable because gc never deletes event rows. sol pointed at `content_event_dedupe_datasource.go:801`, which runs `DELETE FROM events WHERE id = ?`. I checked: correct, and it is the only production path that deletes event rows. So the branch is reachable, and hitting it freezes that session's refinement permanently rather than transiently, since nothing repairs the dangling reference. Pre-existing for the upper bound since #1683, so → **#1777**, not widened into this PR. Body corrected.

sol confirmed independently that no placeholder was added or moved, and reached the same conclusion I did on `session_refinement_usecase_impl.go:141` — no current caller can be rejected by the new guard.

Review notes: implementation gpt-5.6-luna, review Claude Opus 5, independent review gpt-5.6-sol.

Issue #1742, opened 2026-08-09
v0.34.0-46: compress the command_audits text columns


duck8823:
## Problem

#1685 applies the payload codec to `events.body`. That is only **-0.584 GiB** of the 1.67 GiB the
#1620 store gate is counting on. The other **-1.087 GiB** is in `command_audits`:

| source | GiB |
|---|---|
| `command_audits.output_text` | -0.925 |
| `command_audits.command_text` + `input_text` | -0.162 |

Splitting it out of #1685 keeps that PR reviewable — it carries a new command, a new checkpoint
table, migration 053, a fixpoint protocol, and the SQL sweep in D6 of the design note. This issue is
the remainder, and it is deliberately small: three more `(table, field)` rows through the same engine.

## This is a dependency of the gate arithmetic

**#1679's 1.43 GiB index budget is derived from the full 1.67 GiB.** With only #1685 merged the
residual is ~3.656 GiB, and the 4 GiB gate leaves the index ~0.344 GiB.

So this issue is an explicit dependency of **#1679** and **#1620**. "#1685 merged" is not sufficient
for either.

## Why this half is cheap

The hazards that make #1685 High risk are all on the `events` side and mostly do not recur here.
Swept with `rg` over `infrastructure/`, `application/`, `presentation/`, `domain/`:

- **No trigger reads audit text.** The only ones that did were migration 032's search writers, and
  migration 052 (#1718) dropped every one of them.
- **No projection carries audit text.** `event_metadata_projection` projects event metadata only, so
  there is no equivalent of #1685's `legacy_source_hook` / `body_stored_bytes` corruption.
- **No SQL interprets audit text.** Nothing calls `json_extract` or `LIKE` on the three columns —
  #1685's D6 category (A) is empty here.
- The codec metadata columns already exist for all three fields
  (`schema/sqlite/migrations/000036_add_payload_codec_foundation.sql:22-31`), the archive v1
  descriptor already carries them (`domain/archive_segment.go:60`), and
  `payload_codec_compatibility.go:31-40` already counts all three lanes in both modes.

What does recur is #1685's D6 category (B) — byte accounting that must say which byte it means:

| site | note |
|---|---|
| `infrastructure/sqlite/search_projection_rebuild.go:375` | already prefers `*_plaintext_bytes` for the logical figure; the physical `length()` terms are intentional — verify, do not "fix" |
[...]

PR #1778, opened 2026-08-10
fix: measure rehearsal plaintext payloads in bytes

### LINKED ISSUE #1749: v0.34-follow-up: payload rehearsal sizes text columns in characters, not bytes
Follow-up from #1742 / PR #1748, raised by the adversarial review.

## Problem

`payload_rehearsal.go` sizes the corpus with SQLite's `length()` applied to TEXT columns:

```go
// infrastructure/sqlite/payload_rehearsal.go:209
SELECT count(*), coalesce(sum(length(body)),0) FROM events

// infrastructure/sqlite/payload_rehearsal.go:212
SELECT count(*), coalesce(sum(length(command_text)+length(input_text)+length(output_text)),0) FROM command_audits
```

`length(X)` returns **characters** for a TEXT value and **bytes** only for a BLOB:

```
sqlite> SELECT length('あいう'), length(CAST('あいう' AS BLOB));
3|9
```

So every non-ASCII identity payload is under-counted, by up to 3× for CJK and 4× for emoji. Traceary stores Japanese prompts, transcripts and command output routinely, so this is not a corner case.

## Why it matters now

The rehearsal exists to predict the cost of the payload codec before running it against a real store — it is the input to the #1620 gate arithmetic. Under-counting the plaintext side makes the predicted saving look smaller than it is, and the prediction is what the gate decision rests on.

It also becomes ambiguous rather than merely wrong once the backfill (#1685, #1742) has run: a zstd body is a BLOB, so `length()` switches to returning bytes — the physical size, not the plaintext size the rehearsal wants. The same expression then means two different things depending on the row's codec.

## Fix

Both sites need to say which byte they mean, the same way #1685's D6 sweep and #1742 did for the other accounting sites:

- prefer `*_plaintext_bytes` where the codec metadata is present,
- fall back to `length(CAST(<column> AS BLOB))` for legacy all-NULL rows,
- and never `length(<text column>)` bare.

## Acceptance

- Rehearsal plaintext totals are byte counts for both ASCII and non-ASCII corpora.
- The figures are stable across a corpus before and after the codec backfill runs (the plaintext total must not change just because rows became zstd).
- A test with a non-ASCII corpus that fails against the current expression.

duck8823:
Closes #1749

## What

`PayloadRehearsalAdapter.Preview` sized the corpus with `length()` applied to TEXT columns. SQLite returns **characters** for TEXT and bytes only for BLOB, so every non-ASCII payload was under-counted — up to 3× for CJK, 4× for emoji. Traceary stores Japanese prompts and command output routinely.

The rehearsal is the input to the #1620 gate arithmetic. Under-counting the plaintext side makes the predicted compression saving look smaller than it is, and that prediction is what the gate decision rests on.

## How

Metadata first, bytes as the fallback:

```sql
sum(CASE WHEN body_plaintext_bytes IS NOT NULL
         THEN body_plaintext_bytes
         ELSE length(CAST(body AS BLOB)) END)
```

This is better than casting everything to BLOB, which was my first sketch. Once #1685's backfill has run, a zstd body's BLOB length is the *physical* size, not the plaintext size the rehearsal wants — the same expression would then mean two different things depending on the row's codec. Reading the recorded `*_plaintext_bytes` where it exists keeps `PlaintextBytes` genuinely plaintext on a mixed store, which is the issue's second acceptance criterion.

Legacy rows have no metadata and are identity-encoded, so their stored bytes *are* their plaintext bytes; the fallback is exact rather than approximate.

## The pre-036 target

A rehearsal target only has to be a distinct file from the live store — nothing requires it to be current, and `rehearsalMigrationsPending` exists precisely to detect one that is behind. Such a target has no `*_plaintext_bytes` columns at all, so the aggregate is chosen after a `pragma_table_info` probe.

That branch was reachable (`VerifyStoreCompatibility` returns nil when `store_format_state` is absent, so an old store is not rejected) and untested, which meant a typo in it would have surfaced only as a confusing `no such column` during a real preflight. Added a test that builds the target from a truncated migration catalog — a genuine pre-036 store, not one with columns removed — and asserts the column really is absent before measuring.

One probe covers both tables because migration 036 adds every `*_plaintext_bytes` column in a single step; noted in the code so the asymmetry does not read as an oversight.

## Scope check

`rg 'length\((body|command_text|input_text|output_text)\)'` over non-test code returns only these two sites on `main`. Both are fixed here; the other accounting sites were already converted by #1685's D6 sweep and #1742. `command_audits.command_text` / `input_text` / `output_text` are all `NOT NULL`, so no row is silently dropped from the sum by a NULL propagating through the addition.

## Verification

Three cases, all failing on `main` against a corpus of `あいうえお` / `入力` / `出力` (27 bytes, 9 characters):

```
legacy multibyte text                  -27 +9
encoded row uses recorded plaintext    -27 +6
target predating the codec columns     -27 +9
```

The second is the interesting one: on `main` it reports 6 — the two-byte zstd blob plus 4 characters of audit text — which is neither the plaintext size nor a consistent physical size. That is the ambiguity the issue describes, not just an under-count.

## Review notes

### DIFF
--- a/infrastructure/sqlite/payload_rehearsal.go
+++ b/infrastructure/sqlite/payload_rehearsal.go
@@ -206,10 +206,26 @@ func (a *PayloadRehearsalAdapter) Preview(ctx context.Context, c apptypes.Payloa
 		return apptypes.PayloadRehearsalMetrics{}, xerrors.Errorf("inspect pending rehearsal migrations: %w", err)
 	}
 	var eventRows, auditRows, eventBytes, auditBytes int64
-	if err = db.QueryRowContext(ctx, `SELECT count(*),coalesce(sum(length(body)),0) FROM events`).Scan(&eventRows, &eventBytes); err != nil {
+	// Codec metadata is authoritative for encoded rows. Legacy rows have no
+	// plaintext byte metadata and are identity-encoded, so their BLOB length is
+	// the plaintext byte count. Never use SQLite length(TEXT), which counts
+	// characters rather than UTF-8 bytes.
+	eventAggregateQuery := `SELECT count(*),coalesce(sum(CASE WHEN body_plaintext_bytes IS NOT NULL THEN body_plaintext_bytes ELSE length(CAST(body AS BLOB)) END),0) FROM events`
+	auditAggregateQuery := `SELECT count(*),coalesce(sum(CASE WHEN command_plaintext_bytes IS NOT NULL THEN command_plaintext_bytes ELSE length(CAST(command_text AS BLOB)) END+CASE WHEN input_plaintext_bytes IS NOT NULL THEN input_plaintext_bytes ELSE length(CAST(input_text AS BLOB)) END+CASE WHEN output_plaintext_bytes IS NOT NULL THEN output_plaintext_bytes ELSE length(CAST(output_text AS BLOB)) END),0) FROM command_audits`
+	var codecMetadata bool
+	if err = db.QueryRowContext(ctx, `SELECT EXISTS(SELECT 1 FROM pragma_table_info('events') WHERE name='body_plaintext_bytes')`).Scan(&codecMetadata); err != nil {
+		return apptypes.PayloadRehearsalMetrics{}, xerrors.Errorf("inspect payload codec metadata: %w", err)
+	}
+	if !codecMetadata {
+		// The copied target may predate the codec foundation. Its migration plan
+		// is still measurable, but the metadata columns do not exist yet.
+		eventAggregateQuery = `SELECT count(*),coalesce(sum(length(CAST(body AS BLOB))),0) FROM events`
+		auditAggregateQuery = `SELECT count(*),coalesce(sum(length(CAST(command_text AS BLOB))+length(CAST(input_text AS BLOB))+length(CAST(output_text AS BLOB))),0) FROM command_audits`
+	}
+	if err = db.QueryRowContext(ctx, eventAggregateQuery).Scan(&eventRows, &eventBytes); err != nil {
 		return apptypes.PayloadRehearsalMetrics{}, xerrors.Errorf("inspect event aggregates: %w", err)
 	}
-	if err = db.QueryRowContext(ctx, `SELECT count(*),coalesce(sum(length(command_text)+length(input_text)+length(output_text)),0) FROM command_audits`).Scan(&auditRows, &auditBytes); err != nil {
+	if err = db.QueryRowContext(ctx, auditAggregateQuery).Scan(&auditRows, &auditBytes); err != nil {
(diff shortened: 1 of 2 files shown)


duck8823:
## gpt-5.6-sol independent review: no CRITICAL, no HIGH

Reviewed read-only against `da113958` + `77ecf0e2`. Three non-blocking observations; one adopted, two recorded as understood rather than fixed.

**Confirmed the central claim.** sol traced all three row shapes on a mixed store and reached the same result: legacy rows (all-NULL metadata) are identity because the initial schema is plain `TEXT NOT NULL` *and* because `payloadRow.decode`'s contract returns stored bytes unchanged for all-NULL metadata — a stronger basis than the one I used, which was the initial schema alone. Backfilled identity rows contribute their recorded `len(plaintext)`; zstd rows contribute recorded plaintext, not physical length.

Also verified independently: migration 036 adds all four `*_plaintext_bytes` columns in one transactional migration and no later migration drops or renames them (048 only updates values), so the single probe is sound; the three audit text columns have been `NOT NULL` since the initial migration and no migration relaxed that, so nothing drops out of the sum; and `PlaintextBytes` reaches the outside only as JSON `plaintext_bytes` with no ratio arithmetic and no snapshot or documented character-count contract to update.

**Adopted — "a genuine pre-036 store" overstated it.** The helper applies the *current catalog's* prefix 001–035 to an empty database. sol's point: v0.33.1 shipped through 034, and 035 was added later, so this is the catalog's own prefix rather than a reconstruction of any released schema. The columns and NULL constraints the fallback depends on are identical either way, so the test proves what it needs to, but the comment now says what the helper actually builds.

**Recorded, not fixed — metadata values are trusted unverified.** `*_plaintext_bytes IS NOT NULL` is the only test, so a corrupted or hand-edited store could feed a negative or wrong value straight into the total; migration 036 has no non-negativity or tuple-completeness constraint. sol did not rate this HIGH and I agree: every real writer emits complete metadata, and the run itself rejects partial metadata, negative lengths, and checksum mismatches through `payloadRow.decode`. Preview is a preflight estimate, not an integrity gate. Constraining the metadata tuple in the schema is a separate change with its own migration cost.

**Recorded — the fully-backfilled identity row is not pinned by its own case.** It takes the same `*_plaintext_bytes IS NOT NULL` branch as zstd, so there is no untested code path; sol says the same. The fabricated two-byte zstd body with a placeholder sha256 is legitimate here precisely because Preview neither decodes nor verifies — the assertion is that it reports the recorded plaintext length rather than the physical one.

sol hit a network failure on `gh pr diff` and reviewed the two commits directly with `git show`.

Review notes: implementation gpt-5.6-luna, review Claude Opus 5, independent review gpt-5.6-sol.

Issue #1779, opened 2026-08-10
v0.36-8: decide the remaining payloadCodecIdentity call sites


duck8823:
## Context

#1618 activates the zstd payload codec for **canonical writes only** — `insertEventAndAudit`. That scoping was deliberate: #1685 owned rewriting existing rows, #1618 owns new ones, and widening either would have made the rollback boundary untestable.

Four production call sites still pass `payloadCodecIdentity` unconditionally. Each was left alone on purpose and each has a different reason, so this issue is to decide them individually rather than sweep them.

## The sites

| Site | Current behaviour | Why it was left | Question |
|---|---|---|---|
| `infrastructure/sqlite/bundle_datasource.go` | bundle import re-encodes as identity | keeps imported rows in the representation the bundle format describes | should import adopt the canonical policy, so an imported store does not silently grow relative to a natively-written one? |
| `infrastructure/sqlite/store_archive.go` | archive restore re-encodes as identity | preserves the representation the archive recorded | same question, plus: does a restored store then disagree with `payload_codec_compatibility_state`? |
| `infrastructure/sqlite/content_event_dedupe_datasource.go` | dedupe restore rewrites as identity | the quarantine archive stores decoded plaintext, so restore re-encodes from plaintext | restoring a deduped row un-compresses it; harmless for correctness, but it means `store dedupe --restore` inflates the store |
| `infrastructure/sqlite/raw_body_retention_datasource.go`, `store_management_datasource.go` | retention markers written as identity | a marker is a short synthetic string that would not compress | probably correct as-is; confirm and record why rather than leaving it implicit |

## Why it matters

The #1620 gate measures growth after compression is on. Every path above writes payload bytes that bypass the policy. None of them is a normal steady-state path — imports, restores and markers are occasional — so this is not expected to move the gate. That expectation should be measured, not assumed.

The second reason is drift: with the policy expressed in `encodeCanonicalPayload`, a call site that bypasses it is a place where the codebase can quietly grow a second, divergent answer to "how do we store a payload".

## Acceptance

- Each site above has a recorded decision: adopt `encodeCanonicalPayload`, or stay identity with the reason stated in a comment at the call site.
- Any site that adopts it has a test showing a compressible payload lands as zstd and reads back byte-identical.
- Sites that stay identity are covered by a comment explaining what would break if they changed — enough that a future reader does not "fix" them.

## Not in scope

Changing the bundle or archive **formats**. This is about what the writer does on the way in, not about the interchange representation.

PR #1781, opened 2026-08-10
feat: activate compressed canonical payload writes

### LINKED ISSUE #1618: v0.34.0-51: activate compressed canonical writes

duck8823:
Closes #1618

## What

`infrastructure/sqlite/payload_codec.go` has implemented a versioned zstd codec since #1685, and it has never encoded a canonical write — every call site passed `payloadCodecIdentity`. The comment in `insertEventAndAudit` said so outright:

> v0.34 deliberately encodes canonical writes as identity.

This retires that comment. New event bodies and the three `command_audits` payloads are now encoded by policy: zstd when it shrinks the value, identity otherwise. Existing rows are untouched — that was #1685's job, and keeping the split is what makes the rollback boundary testable.

Against #1620's `≤ +0.5 GiB/month`, projected growth is **2.08 GiB/month** without compression and **0.40** with it. No other lever closes that gap; byte caps and workspace exclusions were considered on #1620 and rejected because compression saves more and discards nothing.

## The one-way upgrade, stated plainly

**v0.33.1 — the current released version — has no store-format compatibility check at all.** `store_format_state` and `VerifyStoreCompatibility` arrived after it; `git show v0.33.1:infrastructure/sqlite/migrate.go` shows a migration loop that silently skips versions it does not know. A v0.33 binary opening a v0.34 store therefore reads a zstd BLOB as TEXT and shows the operator garbage, and **nothing v0.34 ships can prevent it.**

The owner accepted this rather than gating it behind a switch. So the job was to make it impossible to hit unknowingly, which the CHANGELOG and `docs/operations/payload-retention.{md,ja.md}` now say in the only place the advice can still be acted on: take `traceary store backup` **before first running v0.34**. A backup taken afterwards is a backup of an already-compressed store and is no fallback at all.

Migration 036 already set `minimum_reader_version = 34`, and `currentReaderVersion` is 34, so there was nothing to set here — the compatibility gate was already in place and durable.

## Implementation

`encodeCanonicalPayload` holds the policy in one place, shared with the backfill so the two cannot drift:

- **Codec metadata is the capability boundary.** `insertEventAndAudit` takes `codecMetadata bool`; when false the codec columns are not written at all, and a zstd body with no codec metadata would be unrecoverable. `encodeCanonicalPayload` returns identity unconditionally in that case, so zstd is structurally unreachable without metadata rather than merely conventionally avoided.
- **Identity stays legal per row.** Short and incompressible payloads must not grow, so a zstd encode that does not shrink falls back to identity with full metadata. Extracted from `payload_backfill.go`'s existing recipe rather than duplicated.
- **TEXT/BLOB affinity is part of the stored representation.** `storedBodyArg` binds identity as `string` and zstd as `[]byte`. The write path previously bound `string(eventPayload.Bytes)` unconditionally, which was the single most likely way to corrupt this change; it now reuses the same helper the backfill uses.

## Review findings, all fixed

Three HIGH findings from my review of the first implementation. All were confirmed by measurement, not by reading.

**1. `body_stored_bytes` meant two different things.** Migration 053's header assigns this to this issue explicitly — *"INSERT triggers are intentionally untouched (#1618 owns non-identity inserts)"* — and the first implementation did not touch them. Measured on the branch, an 8704-byte compressible body produced:

```
events.body_stored_bytes                    = 58     <- compressed
events.body_plaintext_bytes                 = 8704
event_metadata_projection.body_stored_bytes = 58
```

A row backfilled to zstd by #1685 keeps ~8704, because 053 preserves it on the UPDATE path. Same column, same codec, same content, differing by a factor of 150 depending on how the row became compressed — and it is published on the MCP contract as "persisted payload bytes" and read by the retention candidate queries. `prepared_migration_verifier_test.go:53` states the intended meaning: "body_stored_bytes is canonical pre-codec provenance".

Migration `000056` mirrors 053's guards onto both INSERT triggers. Re-measured after the fix:

```
compressed  codec=zstd     events.stored=8704 plaintext=8704 encoded=58 | proj.stored=8704
short       codec=identity events.stored=5    plaintext=5    encoded=5  | proj.stored=5
```

The same migration puts an explicit codec condition on the `legacy_source_hook` `LIKE` inference. It was already failing closed — zstd's frame magic cannot match `'[phase:subagent]%'` — but by accident rather than construction, which is precisely the distinction 053 drew. That column is NULL for all 607,989 rows in the reference store and is on the v0.35 removal path, so this is about construction, not about restoring behaviour anyone depends on.

**2. The doctor check made `doctor` unusable on a large store.** It ran four `COUNT(*) ... WHERE <lane>_codec='zstd'` queries. None of those columns is indexed, and they ran *before* `isLargeStoreForBoundedDoctor` — the guard that exists because "default doctor is filesystem-metadata-only for stores at or above 2 GiB".

Measured on the reference store (28 GiB, 609,795 events, 287,382 audits) via `sqlite3` with `PRAGMA query_only=1`, using an unindexed predicate as the equivalent scan:

```
COUNT(*) FROM events         WHERE <unindexed>   54.2 s
COUNT(*) FROM command_audits WHERE <unindexed>   17.9 s
COUNT(*) FROM events                              0.4 s   (indexed, for contrast)
```

One events scan plus three audit scans is roughly **110 seconds added to every `traceary doctor` run**. The counts already existed in O(1): migration 036 creates `payload_codec_compatibility_state` with exactly these four counters, maintained by triggers. The inspector now reads the singleton row, and reports `state='invalid'` or a non-counter `mode` as a warning rather than silently falling back to scanning.

**3. The warning fired when it could no longer be acted on.** The check returned WARN whenever any compressed row existed. Compression is on by default, so every healthy store acquires compressed rows within a day — a permanent WARN on every correctly-functioning store, burying the real ones. Worse, the hint said to back up "before upgrading", which by the time the check can see compressed rows has already happened.

It now reports PASS and states the downgrade consequence in the message, in EN and JA. The pre-upgrade advice moved to the CHANGELOG and the operations docs, where someone can still act on it.

## Read-path walk

Every path that reads `events.body` or a `command_audits` text column was walked: CLI list/show/context, MCP, hooks, archive, backup, retention, and search all go through codec-aware hydration or do not interpret the payload. Search was verified by test — a compressed body is found by its plaintext content. No unfixed path reads these columns raw.

`LIKE` / `instr` / `length()` applied directly to those columns: the only production sites are the derivation triggers fixed above and byte-accounting expressions already using `length(CAST(x AS BLOB))`.

## Verification

Failing on `main`:

```
--- FAIL: TestDatasource_SaveWithAuditCanonicalPayloadPolicy
    compressed event storage mismatch: want ["zstd","blob"] got ["identity","text"]
```

The trigger tests need migration 056 rather than the write-path change, so I checked them by removing the migration instead:

```
--- FAIL: .../nonidentity_insert_preserves_provenance_and_skips_hook_inference
--- FAIL: .../nonidentity_insert_without_stored_bytes_falls_back_to_plaintext_bytes
```

The second case is mine. The first supplies `body_stored_bytes` explicitly, so it only exercises the first arm of the trigger's `COALESCE`; `insertEventAndAudit` never writes that column, and that is the shape every canonical write actually has.

Also covered: an incompressible or short body stays identity and stays TEXT affinity; mixed identity and zstd rows in one store read correctly through the same reader; `codecMetadata = false` never produces a zstd row; the doctor check reports state in EN and JA.

## Out of scope, filed separately

`payloadCodecIdentity` remains at bundle import, archive restore, dedupe rewrite, and retention markers. Each was left deliberately and each has a different reason → **#1779**.

## Review notes


duck8823:
## gpt-5.6-sol review: **no CRITICAL or HIGH findings**

Reviewed read-only from a branch archive in `/private/tmp`; the repository and the live store were untouched. `go test ./infrastructure/sqlite ./presentation/cli`, `go test ./...`, `go vet ./...` and `go tool golangci-lint run` (0 issues) were re-run independently.

sol confirmed each question in the review packet, including the two I was least sure of:

- **A zstd payload cannot be written without recoverable codec metadata.** `codecMetadata` is one flag for both the encode decision and the column list, every `insertEventAndAudit` caller propagates the same value, and a failure rolls back the whole transaction. There is no partial-failure path that lands a compressed body without its metadata.
- **Trigger firing order does not matter — correct by construction, not by luck.** This is what I explicitly asked for rather than a measured end state. Whichever `AFTER INSERT` trigger on `events` runs first, the projection's `body_stored_bytes` settles on plaintext provenance, because migration 053's projection UPDATE preserves the same value that migration 056's INSERT arm writes.
- `COALESCE(NEW.body_stored_bytes, NEW.body_plaintext_bytes)` staying NULL when both are NULL is right: it refuses to fabricate provenance from a compressed length. Canonical writes always set `body_plaintext_bytes`.
- Dedupe, archive restore, bundle import, retention and GC/cascade delete all pass through the counter triggers, so the singleton cannot silently drift from reality. `VACUUM INTO` copies the singleton, the rows and the triggers from one consistent snapshot.
- Empty and one-byte payloads stay identity, and an empty value round-trips to an empty **string** — `storedBodyArg` binds it as `string([]byte{})`, not as a zero-length BLOB.
- Integrity metadata is still computed from post-redaction model values in the same INSERT. The deleted comment's guarantee survives the change it was written to guard against.

### The three non-blocking observations

**1. Test gap on the highest-risk change — adopted, `acf2a2fa`.** `TestDatasource_SaveWithAuditCanonicalPayloadPolicy` pinned only the event body and the audit command lane, plus one round trip. `input_text` / `output_text` codec and affinity, the empty payload, and every integrity-metadata column were unpinned. Fair criticism of a test guarding what the PR body calls the highest-risk change in the release, so all four lanes are now pinned, together with the metadata that actually decides whether a compressed row is readable at all: format version, plaintext provenance, encoded size against the stored value, and the checksum over the plaintext.

Mutation-verified rather than asserted:

```
# input/output bound through string(...) instead of storedBodyArg
--- FAIL: compressed audit input storage mismatch
--- FAIL: compressed audit output storage mismatch

# output lane's plaintext and stored byte counts swapped
--- FAIL: audit output integrity metadata mismatch
--- FAIL: audit output encoded bytes = 896, want fewer than plaintext 30
--- FAIL: decode audit compressed output: payload integrity error
          (codec=zstd, row=compressed, field=output): stored length mismatch
```

The third line is the reason this matters: get the metadata wrong and the row does not degrade, it becomes unreadable.

**2. Doctor does not reconcile the singleton against reality — not fixed, deliberately.** sol states there is no production bypass; the divergence needs external SQL that drops the counter triggers. A reconciliation check would be exactly the four full table scans this PR removed for measured cause — 110 s added to every `doctor` run on the reference store — to defend against an operator who has already taken a `sqlite3` shell to their own store. Wrong trade. `state='invalid'` still catches every in-product path.

**3. `*Zstd` fields carry non-identity counts — adopted as a rename, `acf2a2fa`.** The columns are `event_body_nonidentity` and friends and the query never checks the codec, so the field names asserted something the data does not. Renamed to `*NonIdentity` with the reason recorded on the type. Internal type, no JSON tags, no contract change. The doctor message still says "compressed rows", which remains accurate while zstd is the only codec `encodePayload` will produce.

`go build ./...` ✅ / `go test ./...` ✅ / `go tool golangci-lint run` → 0 issues ✅ after both changes.

PR #1782, opened 2026-08-10
feat: inject wake summaries on Antigravity PreInvocation

### LINKED ISSUE #1714: v0.34.0 follow-up: verify the Antigravity PreInvocation output contract before injecting at wake

duck8823:
Closes #1714

## What

#1684 shipped wake-time injection on every host except Antigravity, and left this issue behind for a specific reason: Antigravity parses `PreInvocation` stdout as JSON, and the only output vocabulary this repository had verified was `decision` plus "an empty object so Antigravity injects no steps". Guessing a field name would have shipped a hook that `doctor` reports as installed while it silently injects nothing — the issue says outright that this is worse than not shipping it.

So the first half of this issue was research, not code.

## The contract, confirmed rather than guessed

The field is **`injectSteps`**. Confirmed in the vendor's own hook documentation, shipped with the Antigravity CLI at `~/.gemini/antigravity-cli/builtin/skills/agy-customizations/docs/hooks.md`, whose `PreInvocation` section is titled *"Use to inject context or instructions before the model runs"* and gives:

```json
{
  "injectSteps": [
    { "ephemeralMessage": "Remember to check for lint errors before proposing changes." }
  ]
}
```

`injectSteps` is optional; the supported step types are `toolCall`, `userMessage` and `ephemeralMessage`, the last documented as a transient system message.

This resolves #1714 to its **"if a field exists"** branch, so the spec's host table entry `Antigravity | ● PreInvocation` stands and is now stated positively rather than as an open question.

**`ephemeralMessage`, not `userMessage`.** The payload is recalled context — `hook_wake_injection.go` says so on its own first line — and `userMessage` would present Traceary's text to the model as something the operator typed.

Also recorded from the same document, because it constrains this hook: `PreInvocation` ignores the matcher, takes a flat handler list, and handlers default to a **30-second** timeout. This is the path that blocks every model call.

## Implementation

The obstacle was that `maybeInjectWakeSummaries` writes text straight to an `io.Writer`, which a JSON-parsing host cannot use. Rather than duplicate the selection, budget and marker logic, the shared part is factored into `injectWakeSummaries` behind a sink: the existing hosts pass a sink that writes to stdout, and Antigravity passes one that captures the text for embedding.

Nothing about selection, the byte budget, or the once-per-session marker changed. That matters because `PreInvocation` fires **before every model call** — the marker is the only thing standing between one injection and one per model call, and it is the same marker the other hosts use.

With nothing to inject the response stays exactly `{}`, which the vendor contract treats as valid and which is what this handler emitted before.

## The review finding this PR nearly shipped with

Independent review (gpt-5.6-sol) found a HIGH issue that the first implementation introduced, and it was worse than the one already named below.

`hook_spool.go` replayed a spooled Antigravity `PreInvocation` with `io.Discard`. Grok and Kimi pass `nil` instead, with a comment three lines away explaining exactly why: `maybeInjectWakeSummaries` returns early on `output == nil`, so replay cannot consume the marker. This branch's handler called the shared function directly, so it had no such guard — and `io.Discard` is not nil anyway.

The result: a failed `PreInvocation` gets spooled, a later hook replays it, the replay selects the summaries, the capture sink succeeds, the marker is written, and the JSON carrying `injectSteps` goes to `io.Discard`. Every subsequent `PreInvocation` for that conversation then returns `{}`. The conversation never receives its wake summaries **at all** — not once lost, permanently silenced.

Before this branch the same replay was harmless, because the handler never injected anything.

Fixed on both halves, because neither alone is sufficient: replay passes `nil`, and the handler returns before touching selection, the budget or the marker when the writer is nil.

`TestHookAntigravityPreInvocationWakeInjectionSurvivesSpoolReplay` drives a real spool record through the replay entry point. It was verified against both mutations independently:

```
# guard removed from the handler
--- FAIL: TestHookAntigravityPreInvocationWakeInjectionSurvivesSpoolReplay
# io.Discard restored in hook_spool.go
--- FAIL: TestHookAntigravityPreInvocationWakeInjectionSurvivesSpoolReplay
```

## The same defect exists elsewhere, and is not fixed here

Checking the finding turned up that the **generic** spool replay still passes `io.Discard` for `session` records, and `runHookSession` injects on `start` (`hook_session.go:106`, `:149`). So Claude, Gemini and Codex lose wake injection permanently for any session whose hook spools — and `hook_session.go:95` says replay of an already-committed session start is a common path, not a rare one.

That is pre-existing, on a path this branch does not touch, and it affects a different set of hosts, so it is #1785 rather than a widening of this PR. The comment stating the nil-writer rule names the exception so the file does not claim something untrue.

It is a #1620 gate item: "wake injection on every host" cannot be measured honestly while three hosts silently drop it.

## Verification

The injection test was mutation-verified rather than trusted: deleting the `response["injectSteps"] = ...` assignment fails both meaningful subtests.

```
--- FAIL: TestHookAntigravityPreInvocationWakeInjection/eligible_summaries_use_ephemeral_message
--- FAIL: TestHookAntigravityPreInvocationWakeInjection/second_PreInvocation_is_empty
```

Covered: eligible summaries produce one `ephemeralMessage` carrying exactly the text the shared formatter produces; the **second** firing for the same conversation produces `{}`; no eligible summaries, a zero budget and a negative budget all produce `{}`; and a spool replay does not consume the injection. Assertions decode the JSON and compare structure, not the encoding.

## One behaviour worth naming

The shared code marks the session as injected after the **sink** succeeds, and for Antigravity the sink only captures the text; the JSON write happens afterwards in the handler's deferred writer. So a failure of that final write would lose the injection for that conversation, where the plaintext hosts would retry on the next firing. The comment in the shared code — *"a failed write may be retried on the next firing"* — is therefore slightly weaker for this host. Left as is: the alternative is marking after the response is serialised, which would push marker state into the hook's exit path for a failure mode that costs one injection.

That is a different and much smaller problem than the spool replay one above, which cost every injection rather than one.

## Review notes

Contract research by Claude Opus 5; implementation delegated to gpt-5.6-luna; reviewed by Claude Opus 5 and gpt-5.6-sol.

### DIFF
--- a/.ai/spec/1606.md
+++ b/.ai/spec/1606.md
@@ -90,7 +90,7 @@ Supersedes `docs/architecture/hot-cold-segmented-history.md` (+ `.ja.md`), which
 | Codex CLI | ● `SessionStart` stdout / additionalContext | ● `{"decision":"block","reason":...}` → continuation prompt | ✕ 注入不可 | 39.9% |
 | Kimi Code | ✕ SessionStart は fire-and-forget → **初回 `UserPromptSubmit`** を使う | ● exit 2 →「a message can be appended to let the model continue」 | ✕「return values are completely ignored」 | 38.0% |
 | Grok Build | ● SessionStart | ● Stop | ✕ | 2.4% |
-| Antigravity | ○ PreInvocation（#1714 — 本 issue #1684 では未実装。stdout が JSON で context 載せ先が未検証） | ● Stop | **✕ compact hook 自体が無い** | 1.2% |
+| Antigravity | ● PreInvocation（`injectSteps.ephemeralMessage`、#1714 で実装） | ● Stop | **✕ compact hook 自体が無い** | 1.2% |
 | Gemini CLI | ● SessionStart | ● AfterAgent | ✕ | ~0% |
 
 **結論.**
@@ -311,7 +311,7 @@ stop(畳む) ── やりとり ── stop(スルー) ── compact
 | 起床注入 | 要約が 3 件ある | SessionStart | additionalContext に要約のみ。生本文を含まない | integration |
 | 要約なしは無音 | 要約 0 件 | SessionStart | 何も出力しない | unit |
 | Kimi の注入先 | Kimi ホスト | SessionStart | 何も出さない。初回 UserPromptSubmit で出す | integration |
-| Antigravity の注入先 | Antigravity ホスト | PreInvocation | **#1714 で扱う**（#1684 では未実装。stdout JSON の context フィールドが未検証） | integration |
+| Antigravity の注入先 | Antigravity ホスト | PreInvocation | **`injectSteps.ephemeralMessage` で実装済み（#1714）** | integration |
 | 索引退避で本文は残る | 窓外の event | projection rebuild | 全文検索に出ない。`show` では本文が読める | integration |
 | まだ畳めるうちは捨てない | 継続中で stale でないセッションの未精製 transcript | gc | 本文が残る | unit |
 | 終了済みなら畳めないので捨てる | 終了済み・要約なし（`covers_to` 無し）のセッション | gc | セッション全体が orphan になり `degraded=1` 要約後に本文が捨てられる | unit |
--- a/docs/integrations/antigravity.ja.md
+++ b/docs/integrations/antigravity.ja.md
@@ -12,12 +12,22 @@ Antigravity の `hooks.json` は *hook-group 名* から event config への top
 
 | Antigravity event | Traceary の効果 |
 | --- | --- |
-| `PreInvocation` | `conversationId` を key にした冪等な session 開始/更新（Antigravity に `SessionStart` はありません）。最初の workspace path を workspace とする |
+| `PreInvocation` | `conversationId` を key にした冪等な session 開始/更新（Antigravity に `SessionStart` はありません）。最初の workspace path を workspace とし、最初の対象発火で `injectSteps` を通じて想起要約を注入する |
 | `PreToolUse` (`run_command`) | 提案された `{CommandLine, Cwd}` を `conversationId + stepIdx` で永続化。ブロックしない（`{"decision":"allow"}`） |
 | `PostToolUse` (`run_command`) | 同一 step の `PreToolUse` が永続化した command を突き合わせ、`command_executed` audit を記録（step の `error` 付き）。pending が無ければ fail soft |
 | `Stop` | `transcriptPath` から最新の user prompt と model response を復元し、`prompt` + `transcript`、続いて turn 境界を記録。session は**閉じない** |
 
 Antigravity の payload は camelCase フィールド（`conversationId`, `workspacePaths`, `transcriptPath`, `toolCall.name`, `toolCall.args.CommandLine`, `toolCall.args.Cwd`, `stepIdx`, `terminationReason`）を使います。Traceary は共有 runtime（session / audit / transcript）を再利用する前に、これらを内部形式へ正規化します。
+
+### hook の出力語彙
+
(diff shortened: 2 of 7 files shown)


duck8823:
**gpt-5.6-sol: merge blocking — 1 HIGH.** Verified against the code, accepted, fixed.

> `hook_spool.go:223` / `hook_antigravity.go:206-213` — Antigravity spool replay passes `io.Discard`, but the new wake injection writes the marker when the **capture sink** succeeds, not when `output` is written. A spooled `PreInvocation` replayed by another hook sets the marker while the JSON is discarded, so every later `PreInvocation` for that conversation returns `{}` and the wake summary never reaches the host. Grok/Kimi replay avoid this with a `nil` writer. No test covers the spool replay path.

Confirmed rather than taken on trust:

- `hook_spool.go:223` did pass `io.Discard`; `replayGrokSpoolRecord` three lines below passes `nil` and already carried a comment explaining why.
- `writeAntigravityJSON` (`hook_antigravity.go:562-565`) already no-ops on a nil writer, so switching all four Antigravity actions is safe.
- Passing `nil` alone is not enough — the handler calls `injectWakeSummaries` directly and so never had the `output == nil` guard that protects the other hosts. Both halves were needed.

`TestHookAntigravityPreInvocationWakeInjectionSurvivesSpoolReplay` drives a real spool record through the replay entry point. Verified non-vacuous against each half separately: removing the handler guard fails it, and restoring `io.Discard` in `hook_spool.go` fails it.

**Checking it turned up the same defect on a path this PR does not touch.** The generic `case "session"` replay still passes `io.Discard`, and `runHookSession` injects on `start` (`hook_session.go:106`, `:149`), so Claude, Gemini and Codex are silenced whenever their session hook spools — and `hook_session.go:95` says that replay path is common. Filed as #1785 rather than widened into this PR; it is a #1620 gate item, since "wake injection on every host" cannot be measured while three hosts drop it. The nil-writer comment now names the exception so the file does not assert something untrue.

sol's other requested areas returned no CRITICAL or HIGH findings: the once-per-session marker semantics, the JSON shape against the vendor contract, and the refactor's effect on the existing hosts.

`go build ./...` ✅ / `go test ./...` ✅ / `go tool golangci-lint run` → 0 issues ✅ / CI green on `4db8c187`.

PR #1783, opened 2026-08-10
fix: bound the search projection FTS index with resumable reclaim

### LINKED ISSUE #1740: v0.34-60: bounded search projection's FTS index accumulates tombstones with no deterministic reclaim

duck8823:
Closes #1740

## What

`search_projection_recent_fts` accumulates FTS5 delete markers. Eviction and old-generation cleanup delete rows continuously, each firing `search_projection_recent_ad`, which appends a marker, and nothing reclaims them. `store compact` cannot: measured on a real index, `VACUUM` returned **1%** of the bloat where `('optimize')` returned **60%** — segments merge, they do not vacuum.

This runs FTS5's incremental merge alongside the existing cleanup path, and fixes `FTSLogicalBytes`, which was a straight copy of the source-text byte total published as `fts_logical_bytes`.

## Two review rounds, both about the same mistake

Worth stating plainly, because both defects were found by measurement rather than by reading, and both were more serious than the bug being fixed.

### Round one: the loop was not bounded, only the step

The first implementation ran `merge` until the `total_changes()` delta dropped below 2 — a **full optimize**, in 16-page steps. Bounding each step does not bound the loop. Measured on the branch, shadow tables via `dbstat`, one full reclaim in one transaction:

```
shadow =  13,717,504 B   reclaim =  35.8 ms
shadow =  45,404,160 B   reclaim = 126.2 ms
shadow = 181,338,112 B   reclaim = 533.2 ms      <- past the 250 ms lock cap
```

Linear, ~2.9 ms/MB, against a 250 ms cap at a 1464 MiB design ceiling. And `p.Completed` is set only on the `cleanup → complete` transition, so reclaim sat on the batch that finalises the generation: an overrun rolled back the cleanup deletes with it and the generation never completed.

### Round two: bounding the loop did not stop it failing the batch

The fix bounded the loop by the remaining lock budget with 100 ms of headroom. gpt-5.6-sol pointed out that this cannot bound anything, because a merge step is **indivisible** — the between-step deadline checks cannot stop a step already issued, and the 16-page cap bounds the step's *output*, not the input it scans.

The PR body then claimed "reclaim can never be the reason a batch fails". Measured, that claim was false:

```
merge error: context deadline exceeded
post-interrupt statement error: <nil>          <- statements on a dead transaction appear to succeed
commit error: cannot commit - no transaction is active (1)
cleanup delete survived commit: false          <- the cleanup DELETE was discarded
```

SQLite rolls the **whole transaction** back when an interrupted statement is a write inside an explicit transaction. So an overrunning step destroys the cleanup deletes sharing its transaction, and the retry wrapper turns the result into `SearchProjectionNoProgressError` — the round-one stall, behind a narrower trigger.

**How narrow, measured.** Single indivisible merge step, tombstone-heavy fixture, high-entropy 64 KiB documents:

```
shadow =  62,291,968 B   start_step = 3.16 ms   continue_step = 2.73 ms
shadow = 194,846,720 B   start_step = 1.36 ms   continue_step = 1.29 ms
shadow = 375,091,200 B   start_step = 2.75 ms   continue_step = 1.69 ms
shadow = 536,301,568 B   start_step = 1.07 ms   continue_step = 0.97 ms
```

Flat, not linear: a step costs 1-3 ms regardless of index size, because the 16-page cap bounds what it produces. So the margin against 100 ms of headroom is roughly thirtyfold and there is no trend to extrapolate to the ceiling. sol's estimate of the magnitude was wrong; the mechanism was right.

Remote is not impossible, and the consequence is committed cleanup work being discarded, so the structure changed rather than the claim.

## What ships

- **Reclaim runs after the batch commits, in its own transactions.** One merge step per transaction. An interrupted step now loses only itself; every earlier step is already durable and nothing else shares its transaction.
- **Steps are not cancellable mid-write.** They run on the caller's context stripped of the batch deadline, so a step can overshoot the budget by its own 1-3 ms but can never be interrupted. `busy_timeout` still bounds lock contention.
- **Reclaim failure cannot reach the caller.** It is logged at debug. The batch has already committed by then; reclaim is maintenance.
- **The headroom constant is gone.** Nothing follows reclaim inside a transaction, so there is no window left to protect. The budget is simply how long the reclaim series may run, and it stays the batch's existing lock duration rather than a new knob.
- **Reclaim runs on every non-source batch**, since eviction and old-generation cleanup both create tombstones continuously.
- **The classification is one function.** The first attempt left a dead sub-expression and converted a **committed** batch into `NoProgressError` whenever the lock deadline fired between commit and check.
- **`FTSLogicalBytes` measures the shadow tables** instead of copying the source-text total.
- The 250 ms lock cap is now `apptypes.DefaultSearchProjectionLockTime`, shared by the CLI flag default, the catch-up path and the test.

## What bounded reclaim actually buys

Bounding the loop raises the plateau — the price of not holding the write lock, visible rather than implied. Same fixture, only the budget changed:

```
budget = time.Minute   11,395,072 -> 11,423,744 -> 11,423,744 -> 11,423,744
budget = 250 ms        11,395,072 -> 13,733,888 -> 19,509,248 -> 19,509,248
```

Ten generations at the production budget, with one transaction per step:

```
11,395,072  13,733,888  19,509,248  19,509,248  19,505,152  19,513,344
19,517,440  19,505,152  19,509,248  19,488,768  19,501,056
```

Steady state ~19.5 MB, about **1.71×** the unbounded steady state, flat from generation 2 through 10. Identical to the pre-restructure numbers, because per-step transactions change when work is committed and not what the merge does. The index is bounded; the plateau is higher.

## Tests

The original churn assertion was "the sizes are not strictly increasing", which any single tie satisfies — `[10, 20, 20, 40, 80]` passes it — and it ran at `time.Minute`, measuring a plateau nobody gets. It now runs at `DefaultSearchProjectionLockTime` for ten generations and asserts that every generation after steady state stays within **5%** of it and the final size stays under **2.0×** the initial, with the measured numbers in a comment so a later reader can tell a deliberate bound from a fudge factor. Verified by stubbing `reclaimSearchProjectionFTS` to `return nil`:

```
shadow size generation 4 deviated 39.3% from steady state beyond 5% tolerance
```

The old assertion **passes** on that same sequence.

`TestSearchProjectionFailingReclaimDoesNotFailCommittedBatch` pins the property round two is about, through a function-variable seam that exists only so it can be pinned. Verified by propagating the reclaim error instead of logging it:

```
--- FAIL: TestSearchProjectionFailingReclaimDoesNotFailCommittedBatch
    cleanup batch 0 returned reclaim failure (-want +got): + e"reclaim failed in test"
```

`FTSLogicalBytes` was checked by restoring `s.FTSLogicalBytes = s.RecentBytes`, which fails its test.

## Scope

One FTS5 table: `search_projection_recent_fts`. Legacy `event_search_fts` belongs to #1718 and `store search-retire`. No new command, subcommand or MCP tool — reclaim runs inside the existing cleanup path.

sol also observed that `fts_logical_bytes` sums `length(CAST(x AS BLOB))` over integer columns, which counts decimal digits rather than storage width. Real, non-blocking, and a question about what that field should measure rather than a bug in this change — filed as #1787.

## Review notes


duck8823:
**gpt-5.6-sol: 1 HIGH.** The mechanism is right and the magnitude is wrong. Fixed on the mechanism.

> `search_projection_rebuild.go:406` — the deadline check does not make reclaim time-bounded, because the FTS5 merge statement is indivisible. A step may begin with almost no budget left, and the 16-page limit bounds output, not input scanned. If a step exceeds the 100 ms headroom, the state update or the commit meets an expired lock context, rolls back both cleanup and merge, and the batch can stall repeatedly. Measurements up to 181 MB do not establish safety at 1464 MiB.

**The mechanism is real, and worse than stated.** I measured what SQLite does when a merge step is cancelled mid-statement:

```
merge error: context deadline exceeded
post-interrupt statement error: <nil>          <- statements on a dead transaction appear to succeed
commit error: cannot commit - no transaction is active (1)
cleanup delete survived commit: false          <- the cleanup DELETE was discarded
```

The transaction is rolled back automatically. It is not that a later statement meets an expired context — the transaction is already gone, and the cleanup deletes with it. The PR body's claim that "reclaim can never be the reason a batch fails" was false.

**The magnitude is not supported.** sol asked for single-step latency across index size, so I measured it — tombstone-heavy fixture, two thirds of each round deleted, high-entropy 64 KiB documents:

```
shadow =  62,291,968 B   start_step = 3.16 ms   continue_step = 2.73 ms
shadow = 194,846,720 B   start_step = 1.36 ms   continue_step = 1.29 ms
shadow = 375,091,200 B   start_step = 2.75 ms   continue_step = 1.69 ms
shadow = 536,301,568 B   start_step = 1.07 ms   continue_step = 0.97 ms
```

Flat, not linear — a step costs 1-3 ms regardless of size, because the 16-page cap bounds what the step produces. There is no growth trend to extrapolate to 1464 MiB, and the margin against 100 ms of headroom is roughly thirtyfold. The trigger is a cold or contended disk, not the steady state.

Remote is not impossible and the blast radius is committed cleanup work, so the structure changed rather than the claim: reclaim now runs **after** the batch commits, one merge step per transaction, on a context without the batch deadline. An interrupted step loses only itself. `searchProjectionFTSReclaimHeadroom` is deleted — nothing follows reclaim inside a transaction any more.

`TestSearchProjectionFailingReclaimDoesNotFailCommittedBatch` pins it, verified by propagating the reclaim error instead of logging it:

```
--- FAIL: TestSearchProjectionFailingReclaimDoesNotFailCommittedBatch
    cleanup batch 0 returned reclaim failure (-want +got): + e"reclaim failed in test"
```

The churn numbers are byte-identical to the pre-restructure run, which is the expected result: per-step transactions change when work commits, not what the merge does. Re-verified non-vacuous against the `return nil` stub.

sol's other requested areas returned no further CRITICAL or HIGH findings. Of the non-blocking observations:

- **`fts_logical_bytes` counts integer columns as decimal digit counts** — real. It is a question about what that field should measure rather than a defect in this change, so it is #1787 rather than a widening of this PR.
- **Eviction deletes, the `ad` trigger, merge, state update and commit are serial in one transaction** — confirmed, and now moot for reclaim, which has left that transaction.
- **A caller-originated `context.DeadlineExceeded` becomes `SearchProjectionNoProgressError`** — kept. An operator cancelling the run is a run that made no further progress, which is what the caller acts on.

`go build ./...` ✅ / `go test ./...` ✅ / `go tool golangci-lint run` → 0 issues ✅.

PR #1784, opened 2026-08-10
fix: report encoded bytes in raw-body retention plans

### LINKED ISSUE #1759: v0.34.0-x: retention plan overstates reclaimable bytes by summing decoded plaintext

## Problem

`store retention plan` tells the operator how much space applying the plan would reclaim. That number is computed from the **decoded plaintext length** of each candidate body, not from the encoded extent actually stored on disk.

- `infrastructure/sqlite/raw_body_retention_datasource.go:52-92` — the candidate list carries the decoded body
- `application/usecase/raw_body_retention_usecase.go:53-65,90-93` — the reclaim figure sums those decoded lengths

Since #1685 backfilled bodies to zstd, the decoded length is routinely several times the stored length. The plan therefore promises several times the space that applying it frees.

## Failure scenario

An operator sees `plan` report ~3 GiB reclaimable, accepts the plan on that basis, applies it, and recovers a fraction of that after `store compact`. The decision to discard was made against a number that was never about disk.

The direction is the dangerous one: the estimate overstates the benefit of an irreversible operation, so it argues for discarding more than the situation warrants.

## Suggested fix

Sum the encoded extent instead. `body_encoded_bytes` already records it per row, so the plan can select and sum that column rather than measuring the decoded value it loads for verification. `body_stored_bytes` is deliberately pre-codec provenance (see migration 053's trigger guard) and is **not** the right column here.

Also worth stating in the output or the docs that the figure is logical bytes freed inside the SQLite file, which only becomes free disk space after `store compact`.

## Scope

Query + usecase arithmetic + test. No schema change — the column already exists.

duck8823:
Closes #1759

## What

`store retention plan` tells the operator how much space applying the plan would reclaim, and that number was the **decoded plaintext length** of each candidate body. `raw_body_retention_datasource.go` even scanned `body_stored_bytes`, validated it, then threw it away and used `len(plain)` instead.

Since #1685 backfilled bodies to zstd — and since **#1618 landed**, so canonical writes are compressed too — the decoded length is routinely several times the stored one. The direction is the dangerous one: it overstates the benefit of an **irreversible** discard, so it argues for discarding more than the situation warrants.

## Two rounds, because the first fix was still gross

The encoded extent is not what applying the plan frees either. Apply does not empty the body — `raw_body_retention_datasource.go:243` replaces it with `EventBodyUnavailableRetentionMarker`, identity-encoded, 37 bytes. The body column loses the candidate's encoded extent and gains the marker back.

The test's own candidates make it concrete. Two candidates of 44 and 9 encoded bytes were reported as 53 bytes reclaimed. What actually happens:

```
Current   53      encoded bytes the body column holds today
Projected 74      encoded bytes it holds afterwards (2 x 37)
Net      -21      applying this plan costs 21 bytes
```

A plan that reports a 53-byte gain in fact costs 21. Same failure as the original bug, same direction.

## Where the figure lives, and where it does not

The first attempt at the net figure stored it in `Candidates[].LogicalExtent.Bytes` — the canonical, hashed payload. That required widening the published `retention-plan/v1` grammar to accept negative integers without changing the version, and forced the apply path to invert the arithmetic to recover the real encoded extent that `verifyRawBodyCandidateState` compares against. It also made the same store state produce a different plan identity across versions that touch the marker text.

The net figure is now **not persisted at all**. The payload stores what it is derived from:

- `Candidates[].LogicalExtent.Bytes` — the candidate's **encoded extent**. A measurement, non-negative, passed straight to candidate verification with no inversion.
- `Ceilings[].Current` — the **gross total**: what the body column holds today.
- `Ceilings[].Projected` — the **sum of the marker extents**: what it holds afterwards. It was hardcoded `"0"`, which was already wrong.

The operator-facing figure is `Current - Projected`, computed where the plan is displayed. `canonicalSignedInteger` is gone; both stored values are non-negative again.

**Bumped to `retention-plan/v2`.** `Projected` changed meaning from "always zero" to "bytes remaining after apply", which is a published contract change — and the bump makes a plan created under the old semantics fail as an unsupported version rather than silently applying against a figure that meant something else. For an irreversible operation that is the right failure and the right diagnosis.

## The query

```sql
CASE WHEN e.body_encoded_bytes IS NOT NULL
     THEN e.body_encoded_bytes
     ELSE length(CAST(e.body AS BLOB))
END
```

`body_stored_bytes` is deliberately pre-codec **provenance**, not a physical size — migration 053's trigger guard and migration 056's INSERT arms both preserve it as plaintext bytes on purpose — so it is not the column for this and its meaning is unchanged. Legacy rows have no codec metadata and are identity-encoded, so their on-disk extent is the BLOB length; the `CAST` is load-bearing, because `length()` on TEXT counts characters and understated the reference store by 20.3%.

`RawBodyCandidate.StoredBytes` named one thing and held another, so it is split: `EncodedBytes` is the reclaimable extent, `PlaintextBytes` stays provenance. The retention **ledger** keeps recording plaintext bytes and restore keeps validating against them.

## Findings, all fixed

**1. The number the issue is about had no test.** The first implementation tested the datasource snapshot — one layer below the arithmetic. Reverting the usecase to the pre-fix behaviour and running the **whole** suite left every package green. There is now a test at the layer that produces the plan, asserting exact values, and the same mutation fails it.

**2. A safety check was dropped and needed justifying rather than accepting.** `loadRawBodyRecovery` no longer compares the recovery body's length. That loses nothing — a SHA-256 match already implies equal length, and the digest comes from the reviewed plan — and it was necessary because the field it compared against no longer holds a plaintext length. The new `stored.Int64 != plaintextBytes` check was accepted only after confirming migrations 021, 026 and 053 all derive that column with `length(CAST(... AS BLOB))`, so no existing row carries a character count.

**3. The figure was still gross (gpt-5.6-sol, HIGH).** Above.

**4. The net figure was persisted in the hashed payload (gpt-5.6-sol, design).** Above.

Mutation-verified independently, both directions:

```
# display uses Current alone
retentionNetChange() = "44", want "7"
retentionNetChange() = "9",  want "-28"

# Projected restored to "0"
retention plan extents mismatch (-want +got):
-   Projected: "74"
+   Projected: "0"
```

A `retention-plan/v1` payload is rejected by name: `unsupported retention plan schema "retention-plan/v1"`.

## One behavioural narrowing, stated deliberately

`verifyRawBodyCandidateState` now also compares the encoded extent, so a plan created **before** a re-encode — the #1685 backfill, for instance — becomes unappliable and must be re-created, where previously only a content or timestamp change did that. Kept: an irreversible discard should not apply against a figure that has since changed. Pinned by `TestRawBodyRetention_rejectsReencodedCandidate`.

## Not in scope, filed instead

A candidate whose net contribution is zero or negative arguably should not be selected for an irreversible discard at all. That is a change to selection policy rather than to the reported figure, and it turns on what the retention ledger is for — #1786.

## Verification

No schema change to the database: the column already existed. No new command, subcommand or MCP tool.

## Review notes


duck8823:
**gpt-5.6-sol: 1 HIGH, plus a design review on the fix for it.** Both accepted.

> `raw_body_retention_usecase.go:59` — the reclaim figure sums each candidate's full encoded extent, but apply does not empty the body: `raw_body_retention_datasource.go:243-259` replaces it with a 36-byte retention marker. The real reduction is at most `EncodedBytes - 36`, and is negative for short bodies. The new test's candidates of `44` and `9` report "53 bytes reclaimed" while the bodies alone become `36 x 2 = 72` afterwards — 19 bytes more, before the codec metadata legacy rows also gain.

Verified against the code and accepted. The marker is 37 bytes, not 36 — `"[traceary:body-unavailable:retention]"` — so the arithmetic is slightly worse than stated:

```
Current   53      encoded bytes the body column holds today
Projected 74      encoded bytes it holds afterwards (2 x 37)
Net      -21      applying this plan costs 21 bytes
```

Same failure as the original bug, same direction: an irreversible discard sold with an inflated number.

**The first fix put the net value in the wrong place, and sol was asked to attack my objection to it.** Storing net in `Candidates[].LogicalExtent.Bytes` meant widening the published `retention-plan/v1` grammar to signed integers without a version bump, and inverting the arithmetic in `prepareExecution` to recover the extent that candidate verification actually compares against.

sol agreed the per-candidate extent must stay a non-negative measurement, and improved on where the net should go: **do not persist it at all.** Store `Current` (gross total) and `Projected` (sum of marker extents) as before and after, and subtract at display time. `Projected` was hardcoded `"0"`, which was already wrong — applying the plan does not leave zero bytes behind.

That is what ships. `canonicalSignedInteger` is deleted, `retentionGrossBytes`/`retentionNetBytes` are deleted, and `prepareExecution` passes the stored extent straight through.

**Schema bumped to `retention-plan/v2`**, per sol's recommendation. `Projected` changed meaning, which is a published contract change, and the bump makes a plan created under the old semantics fail as an unsupported version instead of applying against a figure that meant something else. Rejected by name: `unsupported retention plan schema "retention-plan/v1"`.

Mutation-verified independently in both directions:

```
# display uses Current alone
retentionNetChange() = "44", want "7"
retentionNetChange() = "9",  want "-28"

# Projected restored to "0"
retention plan extents mismatch (-want +got):
-   Projected: "74"
+   Projected: "0"
```

**One thing sol raised that is not being fixed here.** A candidate whose net contribution is zero or negative arguably should not be selected for an irreversible discard at all. That is selection policy rather than the reported figure, and it turns on what the retention ledger is for — filed as #1786.

`go build ./...` ✅ / `go test ./...` ✅ / `go tool golangci-lint run` → 0 issues ✅.

PR #1788, opened 2026-08-10
fix: stop spool replay consuming the wake-injection marker

### LINKED ISSUE #1785: fix: spool replay of session hooks consumes the wake-injection marker
Found while reviewing #1714 (PR #1782). This is a **pre-existing** defect on a different code path, so it is filed separately rather than folded into that PR.

## What is wrong

`presentation/cli/hook_spool.go:193` replays a spooled `session` hook with `io.Discard`:

```go
case "session":
    return c.runHookSession(ctx, io.Discard, input, client, action, dbPath)
```

`runHookSession` performs wake injection on the `start` action, at `hook_session.go:106` and `hook_session.go:149`. `maybeInjectWakeSummaries` returns early only when `output == nil` — and `io.Discard` is not nil.

So a replayed session start selects the summaries, writes them into `io.Discard`, and then calls `markHookWakeInjected`. The marker is consumed and the text is thrown away. Every later `SessionStart` for that client and session returns nothing, because the marker says the injection already happened.

The session never receives its wake summaries at all. Not one lost injection — all of them, permanently, for that session.

## Why it is likely rather than theoretical

`hook_session.go:95-98` says so itself:

> Spool replay (and hosts that re-fire SessionStart) often hit a session that already committed before the kill.

Replay of a spooled session start is an expected path, and the `ErrInvalidSessionState` branch at line 106 is precisely the branch that path takes.

## Hosts affected

Every host routed through the generic `case "session"` replay: Claude, Gemini and Codex. Grok and Kimi are not affected — `replayGrokSpoolRecord` and `replayKimiSpoolRecord` already pass `nil`, with a comment stating the reason. Antigravity is fixed in #1714.

## Fix

Pass `nil` rather than `io.Discard` for the generic replay, matching what Grok, Kimi and Antigravity already do. `runHookCompact` at line 197 does not inject, but it should switch too so the file carries one convention rather than two — and so the comment above `replayGrokSpoolRecord`, which now states the rule for spool replay generally, is true of the whole file.

A test must pin it at the replay entry point, so that changing it back to `io.Discard` fails: after a spool replay of a session start for a session with eligible summaries, a subsequent live `SessionStart` still emits the summary text. #1714's `TestHookAntigravityPreInvocationWakeInjectionSurvivesSpoolReplay` is the model — it was verified non-vacuous against both halves of its fix.

## Why this blocks the v0.34 gate

"Wake injection on every host" is a #1620 gate row. This defect silences injection on three hosts whenever their session hook spools, so the row cannot be measured honestly until it is fixed.

duck8823:
Closes #1785

## What

Spool replay of a `session` hook passed `io.Discard`. `runHookSession` injects wake summaries on the `start` action (`hook_session.go:106`, `:149`), and `maybeInjectWakeSummaries` returns early only when `output == nil` — `io.Discard` is not nil.

So a replayed session start selected the summaries, wrote them into `io.Discard`, and then wrote the marker. The marker is consumed, the text is gone, and every later `SessionStart` for that client and session returns nothing because the marker says the injection already happened.

The session never receives its wake summaries. Not one lost injection — all of them, permanently.

Claude, Gemini and Codex were affected. Grok, Kimi and Antigravity already pass `nil`; Antigravity was fixed in #1714, which is where this defect was found.

## Not a rare path

`hook_session.go:95-98` says so itself:

> Spool replay (and hosts that re-fire SessionStart) often hit a session that already committed before the kill.

The `ErrInvalidSessionState` branch that comment describes is the branch containing the injection at line 106.

## The fix

`replayHookSpoolRecord` passes `nil` for `session` and for `compact`, matching what the Grok, Kimi and Antigravity replays already do. `runHookCompact` does not inject, but it already carries its own `output == nil` guard for exactly these callers, so switching it costs nothing and leaves the file with one convention instead of two.

The comment stating the rule moves to `replayHookSpoolRecord`, where it governs every replay rather than reading as documentation for one host's helper.

## Test

`TestHookSessionStartWakeInjectionSurvivesSpoolReplay` drives a real spool record through the replay entry point and then asserts a subsequent live `SessionStart` still emits the summary text. Verified non-vacuous by restoring `io.Discard`:

```
--- FAIL: TestHookSessionStartWakeInjectionSurvivesSpoolReplay
    - true
    + false
    stdout=""
```

The empty `stdout` is the defect exactly: the live firing produces nothing because the replay consumed the marker.

## Why now rather than after the release

"Wake injection on every host" is a #1620 gate row. It cannot be measured honestly while three hosts silently drop it whenever their session hook spools.

## Verification

No new command, subcommand, flag or MCP tool.

## Review notes

Found by Claude Opus 5 while checking a gpt-5.6-sol finding on #1714; implementation delegated to gpt-5.6-luna; reviewed by Claude Opus 5.

### DIFF
--- a/presentation/cli/hook_spool.go
+++ b/presentation/cli/hook_spool.go
@@ -183,6 +183,10 @@ func (c *RootCLI) drainHookSpoolRecordsDetailed(ctx context.Context, limit int) 
 	return result
 }
 
+// Spool replay passes a nil writer, not io.Discard: wake injection (#1684) marks
+// the session as injected once it has written, so a discarded write would
+// consume the marker and silence the live firing. A nil writer makes injection
+// a no-op.
 func (c *RootCLI) replayHookSpoolRecord(ctx context.Context, record hookSpoolRecord) error {
 	input := newExplicitHookPayloadReader([]byte(record.Payload))
 	dbPath := record.DBPath
@@ -190,11 +194,11 @@ func (c *RootCLI) replayHookSpoolRecord(ctx context.Context, record hookSpoolRec
 	action := record.Action
 	switch strings.TrimSpace(record.Command) {
 	case "session":
-		return c.runHookSession(ctx, io.Discard, input, client, action, dbPath)
+		return c.runHookSession(ctx, nil, input, client, action, dbPath)
 	case "audit":
 		return c.runHookAudit(ctx, input, client, dbPath)
 	case "compact":
-		return c.runHookCompact(ctx, io.Discard, input, client, action, dbPath)
+		return c.runHookCompact(ctx, nil, input, client, action, dbPath)
 	case "subagent-start":
 		return c.runHookSubagentStart(ctx, input, client, dbPath)
 	case "subagent-stop":
@@ -232,14 +236,6 @@ func (c *RootCLI) replayAntigravitySpoolRecord(ctx context.Context, input io.Rea
 	}
 }
 
-// Spool replay passes a nil writer, not io.Discard: wake injection (#1684) marks
-// the session as injected once it has written, so a discarded write would
-// consume the marker and silence the live firing. A nil writer makes injection
-// a no-op.
-//
-// The generic session replay above is the one exception left, and it is a
-// defect rather than a decision: it still passes io.Discard and consumes the
-// marker for Claude, Gemini and Codex. Tracked by #1785.
 func (c *RootCLI) replayGrokSpoolRecord(ctx context.Context, input io.Reader, action, dbPath string) error {
 	switch strings.TrimSpace(action) {
 	case "session-start":
--- a/presentation/cli/hook_wake_injection_test.go
+++ b/presentation/cli/hook_wake_injection_test.go
@@ -346,6 +346,56 @@ func TestHookAntigravityPreInvocationWakeInjectionSurvivesSpoolReplay(t *testing
 	}
 }
 
+func TestHookSessionStartWakeInjectionSurvivesSpoolReplay(t *testing.T) {
+	const workspace = "github.com/duck8823/traceary"
+	const replaySessionID = "sess-generic-replay"
+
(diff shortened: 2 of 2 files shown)


duck8823:
**gpt-5.6-sol: no CRITICAL or HIGH findings.** Two of the three questions I asked were answered by tracing rather than asserted, which is what I wanted:

- **`nil` is safe on every `runHookSession` path.** The writer is used only by `maybeInjectWakeSummaries`, which guards on nil at `hook_wake_injection.go:88`. No unguarded use in `end`, `stop`, the one-shot start/end wrappers, or any error path.
- **Skipping rather than discarding changes no state in `runHookCompact`.** The nil branch at `hook_runtime.go:713` skips only the read-only `ContextUsecase.Handoff`, its formatting, and the `fmt.Fprint`. Markers, events and progress state are untouched, and the `pre-compact` / `post-compact` persistence does not depend on the writer at all.
- **The wake-injection marker is the only write-once state written after a successful output** on these paths. Session state writes and cleanup happen before injection and are unaffected by the writer.

sol also confirmed the test routes through the generic `replayHookSpoolRecord` rather than a host-specific replay, which was the failure mode I was most concerned about — a test that passes only because Grok and Kimi were already correct.

**One non-blocking observation, not adopted here.** The test would not catch a regression that moved the marker write *before* the sink, because replay never calls injection at all now and the live write then succeeds. That is true, and it is a property of #1684's design rather than of this change — the ordering it describes is what makes a failed write retryable on the next firing. A failing-writer test belongs with that ordering, not with the replay fix, and widening this PR to add one would put two unrelated properties in one change.

`go build ./...` ✅ / `go test ./...` ✅ / `go tool golangci-lint run` → 0 issues ✅ / CI green on `97958172`. Mutation-verified independently: restoring `io.Discard` fails the test with an empty `stdout`, which is the defect exactly.

Issue #1791, opened 2026-08-10
v0.34.0 follow-up: payload rehearsal bounds Pause/ReleaseScrub on some paths and not others


duck8823:
Noticed while checking whether #1789's checkpoint-anchoring defect appears elsewhere. The anchoring is fine here — every context is derived at the point of use — but the **bounding** is not consistent, and the same call is bounded on some paths and unbounded on others in the same function.

`application/usecase/payload_rehearsal.go`:

| line | call | bound |
|---|---|---|
| 99 | `u.workflow.Pause(context.WithoutCancel(ctx), handle)` | none |
| 116 | `u.workflow.Pause(pauseCtx, handle)` | 1s |
| 122 | `u.workflow.Pause(context.WithoutCancel(ctx), handle)` | none |
| 126 | `u.workflow.Pause(pauseCtx, handle)` | 1s |
| 168 | `u.scrubber.ReleaseScrub(context.WithoutCancel(ctx), handle)` | none |
| 176 | `u.scrubber.ReleaseScrub(releaseCtx, handle)` | 1s |

`context.WithoutCancel` strips the caller's deadline as well as its cancellation, so the unbounded rows wait on nothing. The reasoning `infrastructure/sqlite/payload_backfill.go` gives for bounding the equivalent write applies here too:

> acquiring the store lease waits on nothing but the context. An unbounded checkpoint would turn "someone else holds the store" into a process that never returns. Losing the checkpoint is recoverable […] and hanging is not.

`Pause` and `ReleaseScrub` both reach the store, so a concurrent holder makes those three call sites hang with no ceiling — including line 99 and line 168, which run on the failure path of a `liveResult` check, exactly when something is already wrong.

The three bounded sites already show the intended shape. Either the unbounded ones should adopt it, or the difference should be justified in a comment; right now it reads as an oversight rather than a decision.

Worth checking `#1780` (payload rehearsal aborts on "cannot disable rehearsal WAL autocheckpoint") for overlap while this is being looked at — that is also a rehearsal-path failure whose recovery goes through these calls.

No measurement attached: this is read from the code, not observed in a hang.

PR #1792, opened 2026-08-10
fix: anchor payload backfill checkpoints per transition

### LINKED ISSUE #1789: v0.34.0: payload backfill cannot record a terminal transition on any run longer than 10 seconds
Found while rehearsing the #1620 release gate against a 30 GB copy of the live store. It blocks the gate: `store payload-backfill run` is the compression lever the store-size row depends on, and it cannot report completion on any real store.

## What happens

`infrastructure/sqlite/payload_backfill.go:250` derives the checkpoint context **once, at `execute()` entry**:

```go
closeCtx, cancelClose := context.WithTimeout(context.WithoutCancel(ctx), payloadBackfillCheckpointTimeout)
defer cancelClose()
```

`payloadBackfillCheckpointTimeout` is 10 seconds. Every terminal transition in the loop then runs on that same, already-aged context:

| line | call |
|---|---|
| 297 | `completeBackfillRun` |
| 308 | `resetBackfillCursor` (end of a dirty pass) |
| 321 | `failBackfillRun` |
| 347 | `pauseBackfillRun` (`--stop-after-batches`) |
| 380 | `pauseBackfillRun` (via `abortRun`, on cancellation) |

So a run that takes longer than 10 seconds of wall clock cannot record **any** terminal transition. The window is consumed by the work it is supposed to outlive.

The comment above it is right about why the checkpoint must survive cancellation, and right that it must be bounded. It is the anchoring that is wrong: the deadline has to start when the checkpoint is taken, not when the run starts.

## Measured on the rehearsal store

30 GB copy of the live store, binary built from `50b953e9`, `--batch-rows 2048`:

```
$ traceary store payload-backfill run --batch-rows 2048 --db-path .../rehearsal.db
Error: failed to execute CLI command: payload backfill run failed: run payload backfill:
       advance payload backfill run: context deadline exceeded
6:23.71 total
```

`store payload-backfill status` afterwards:

```
"state": "running",
"high_water_rowid": 796522,
"cursor_rowid":     796522,
"pass_count":       1,
"scanned_rows":     1279931,
"rewritten_rows":   1279931,
"plaintext_bytes":  3340966020,
"stored_bytes":     1307244443,
"more_pending":     true
```

The cursor reached the high-water mark, so this is `resetBackfillCursor` at the end of pass 1 — the pass rewrote rows, so the walk restarts, and the restart is exactly what could not be recorded.

Note what did work: the compression itself is durable. Batches commit on the caller's context, so 1,279,931 fields were rewritten and 3,340,966,020 plaintext bytes became 1,307,244,443 stored bytes. Only the bookkeeping is lost. The operator sees a non-zero exit, a run stuck at `running`, and `more_pending: true` — with no way to tell that from a genuine failure.

## Reproduction

Deterministic, in `infrastructure/sqlite`. Seed three compressible events and burn the window inside a committed batch:

```go
var once sync.Once
ds.onAfterCommitBatch = func(int64) {
    once.Do(func() { time.Sleep(payloadBackfillCheckpointTimeout + time.Second) })
}
result, err := ds.Run(context.Background(), apptypes.PayloadBackfillConfig{BatchRows: 1})
```

```
Run err          = advance payload backfill run: context deadline exceeded
result state     = ""
persisted state  = "running"
```

Same error string as the 30 GB run. No cancellation is involved: the caller's context is `context.Background()`.

Every existing checkpoint test cancels within milliseconds, so none of them ever reaches the window.

## What to change

Derive the checkpoint context **per transition**, so each terminal write gets its own bounded budget measured from the moment it is taken. `abortRun` should derive its own rather than receiving one.

Shipping a test for this must not sleep ten seconds. Make the window overridable on the datasource in the style of the existing `onSelectedBatch` / `onBeforeCommitBatch` / `onAfterCommitBatch` seams, then pin the property directly: with a small window and a batch hook that sleeps longer than it, the run must still complete. That fails before the change and passes after.

Also worth a look in the same pass: `application/usecase/payload_rehearsal.go:99-176` derives several `WithoutCancel` budgets, and `infrastructure/sqlite/search_projection_rebuild.go:1356` / `:1438` do too. Check whether any of them is anchored the same way before a long stretch of work, rather than assuming this site is the only one.

duck8823:
Closes #1789

## What

Found by rehearsing the #1620 release gate against a 30 GB copy of the live store. `store payload-backfill run` is the compression lever the store-size row depends on, and it could not report completion on any real store.

`infrastructure/sqlite/payload_backfill.go` derived the checkpoint context **once, at `execute()` entry**, then ran every terminal transition on it — `completeBackfillRun`, `resetBackfillCursor`, `failBackfillRun`, and both `pauseBackfillRun` call sites. The window is 10 seconds. Any run longer than that consumed its own checkpoint budget doing the work the checkpoint exists to record.

The comment above it was right about why the checkpoint must outlive cancellation and why it must be bounded. Only the anchoring was wrong, and the anchoring was easy to get wrong because `abortRun` took the context as a parameter.

## Measured, before

30 GB rehearsal store, `--batch-rows 2048`, binary built from `50b953e9`:

```
Error: failed to execute CLI command: payload backfill run failed: run payload backfill:
       advance payload backfill run: context deadline exceeded
6:23.71 total
```

```
"state": "running",   "pass_count": 1,   "more_pending": true,
"high_water_rowid": 796522,   "cursor_rowid": 796522,
"scanned_rows": 1279931,   "rewritten_rows": 1279931,
"plaintext_bytes": 3340966020,   "stored_bytes": 1307244443
```

Cursor at the high-water mark, so this is `resetBackfillCursor` at the end of a dirty pass — the pass restart is what could not be recorded.

Note what did work: batches commit on the caller's context, so 1,279,931 fields were rewritten and 3.34 GB of plaintext became 1.31 GB stored. Only the bookkeeping was lost. The operator saw a non-zero exit, a run stuck at `running`, and no way to tell that from a genuine failure.

## Measured, after

Same store, same run row, binary built from this branch:

```
"state": "completed",   "pass_count": 2,   "more_pending": false,
"completed_at": "2026-08-10T17:47:08.950515Z",
"scanned_rows": 1962377,   "rewritten_rows": 1279931
```

Exit 0. The stranded run resumed, walked the second (clean) pass, and completed.

## The change

`checkpointContext()` derives the budget when a transition is taken rather than when the run starts, and `abortRun` derives its own instead of receiving one. The window is overridable through a `checkpointTimeout` field in the style of the existing `onSelectedBatch` / `onBeforeCommitBatch` / `onAfterCommitBatch` test seams — no flag, environment variable or config surface.

## Non-vacuity

Verified independently of the implementer, by restoring the shared context on the two paths the test covers:

```
--- FAIL: TestPayloadBackfillCheckpointWindowStartsAtEachTransition
    --- FAIL: .../completion_after_a_slow_committed_batch
        payload_backfill_test.go:1130: Run: advance payload backfill run: context deadline exceeded
    --- FAIL: .../pause_after_a_slow_committed_batch
        payload_backfill_test.go:1130: Run: advance payload backfill run: context deadline exceeded
```

Same error string as the 30 GB run. The test does not sleep ten seconds: it shrinks the window and sleeps past it inside a committed batch.

Every pre-existing checkpoint test cancels within milliseconds, which is why none of them ever reached the window.

## The other WithoutCancel sites

Checked, and none has this defect — each derives its context immediately before the work it bounds:

- `search_projection_rebuild.go:1356` / `:1438` — budgets are explicitly `measureTimeout() + evidenceWriteTimeout`, sized for the measure-and-write they wrap.
- `application/usecase/payload_rehearsal.go:99-176` — derived inside the loop, at the point of use.

`payload_rehearsal.go` did turn up something adjacent: the same `Pause` / `ReleaseScrub` call is bounded at 1s on three paths and completely unbounded on three others. That is a separate decision on a separate file — filed as #1791, deliberately not touched here.

## Verification

`go build ./...` ✅ / `go test ./... -count=1` ✅ / `go tool golangci-lint run` → 0 issues ✅ — run independently of the implementer.

No schema change, no new command, subcommand, flag or MCP tool.

## Review notes

### DIFF
--- a/infrastructure/sqlite/payload_backfill.go
+++ b/infrastructure/sqlite/payload_backfill.go
@@ -58,6 +58,9 @@ const payloadBackfillCheckpointTimeout = 10 * time.Second
 // codec: events.body and command_audits.{command_text,input_text,output_text}.
 type PayloadBackfillDatasource struct {
 	db *Database
+	// checkpointTimeout bounds test-injected checkpoint windows. Production
+	// leaves it zero, which uses payloadBackfillCheckpointTimeout.
+	checkpointTimeout time.Duration
 	// onBeforeCommitBatch is a test hook that fires after a batch is selected
 	// and before the verify/encode transaction. Production leaves it nil.
 	onBeforeCommitBatch func(batch []backfillCandidate)
@@ -237,23 +240,10 @@ func (d *PayloadBackfillDatasource) execute(ctx context.Context, c apptypes.Payl
 		return apptypes.PayloadBackfillResult{}, err
 	}
 
-	// Checkpoint writes outlive the cancellation that triggers them. Every
-	// terminal transition — complete, reset, pause, fail — records what the
-	// worker already did durably, so running it on the caller's context would
-	// lose that record exactly when the run needs it most.
-	//
-	// Bounded, though: WithoutCancel drops the deadline too, and acquiring the
-	// store lease waits on nothing but the context. An unbounded checkpoint
-	// would turn "someone else holds the store" into a process that never
-	// returns. Losing the checkpoint is recoverable — resume reads the cursor
-	// the last committed batch left — and hanging is not.
-	closeCtx, cancelClose := context.WithTimeout(context.WithoutCancel(ctx), payloadBackfillCheckpointTimeout)
-	defer cancelClose()
-
 	var batchCount int64
 	for {
 		if err = ctx.Err(); err != nil {
-			return d.abortRun(ctx, closeCtx, db, run, batchCount, err)
+			return d.abortRun(ctx, db, run, batchCount, err)
 		}
 
 		// Begin a new pass when the cursor is at the origin.
@@ -268,7 +258,7 @@ func (d *PayloadBackfillDatasource) execute(ctx context.Context, c apptypes.Payl
 		}
 		batch, selectErr := d.selectBackfillBatch(ctx, db, run.CursorRowID, ceilings, c.BatchRows)
 		if selectErr != nil {
-			return d.abortRun(ctx, closeCtx, db, run, batchCount, selectErr)
+			return d.abortRun(ctx, db, run, batchCount, selectErr)
 		}
 		if d.onSelectedBatch != nil {
 			batch = d.onSelectedBatch(batch)
@@ -286,18 +276,22 @@ func (d *PayloadBackfillDatasource) execute(ctx context.Context, c apptypes.Payl
 			if !run.passDirty {
 				remaining, countErr := countEligibleBackfillRows(ctx, db, run.CursorRowID, ceilings)
 				if countErr != nil {
-					return d.abortRun(ctx, closeCtx, db, run, batchCount, countErr)
+					return d.abortRun(ctx, db, run, batchCount, countErr)
 				}
 				if remaining > 0 {
-					return d.abortRun(ctx, closeCtx, db, run, batchCount, xerrors.Errorf(
+					return d.abortRun(ctx, db, run, batchCount, xerrors.Errorf(
 						"payload backfill selection returned no candidates while %d eligible fields remain past cursor %d",
 						remaining, run.CursorRowID,
 					))
 				}
-				if completeErr := completeBackfillRun(closeCtx, db, run); completeErr != nil {
+				checkpointCtx, cancelCheckpoint := d.checkpointContext(ctx)
+				completeErr := completeBackfillRun(checkpointCtx, db, run)
+				if completeErr != nil {
+					cancelCheckpoint()
 					return apptypes.PayloadBackfillResult{}, completeErr
 				}
-				loaded, loadErr := loadBackfillRun(closeCtx, db, run.RunID)
+				loaded, loadErr := loadBackfillRun(checkpointCtx, db, run.RunID)
+				cancelCheckpoint()
 				if loadErr != nil {
 					return apptypes.PayloadBackfillResult{}, loadErr
 				}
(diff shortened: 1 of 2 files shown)


duck8823:
**gpt-5.6-sol: no CRITICAL or HIGH findings.** Verified by tracing rather than asserted:

- All five transitions derive `checkpointContext()` at the point of use — `completeBackfillRun` (`:287`), `resetBackfillCursor` (`:302`), `failBackfillRun` (`:318`), the `StopAfterBatches` pause (`:346`), and the cancellation pause in `abortRun` (`:402`). No path still carries a context from `execute()` entry, including the `loadBackfillRun` reads that follow each write.
- No `cancel` leaks. The in-loop derivations cancel on both the error and success branches; `abortRun` is function-scoped, so `defer` is correct there.
- `abortRun` cannot get a zero-length budget. Checked against the Go 1.26.5 source rather than assumed: `context.WithoutCancel` returns a context with no deadline, `Done() == nil` and `Err() == nil`, so the subsequent `WithTimeout` allocates a fresh ten seconds whether the parent was cancelled or deadline-exceeded.
- `checkpointTimeout` is unexported and left zero by `NewPayloadBackfillDatasource`, the only production construction path. It is not reachable from CLI config or another package.

**One observation adopted.** sol flagged the 5 ms window against two SQLite writes as narrow enough to flake on a loaded runner. That is fair — the test should fail because the anchoring is wrong, not because the scheduler was busy. Widened to a 150 ms window against a 750 ms sleep in `632a155f`, which keeps the same 5x overrun while giving the writes real headroom. Re-ran with `-count=3`: green.

**Two observations noted, not adopted.**

sol pointed out that `failBackfillRun` and the cancellation path through `abortRun` are not *directly* covered for re-anchoring. Correct, and worth stating what is: the completion subtest seeds one event with `BatchRows: 1`, so pass 1 rewrites the row, the walk hits an empty batch with `passDirty`, and `resetBackfillCursor` runs **after** the 750 ms sleep has already overrun the window. The reset path is covered, and it is the exact transition that failed on the 30 GB store. Covering the fail path needs partial-metadata injection, which is a different test's machinery.

On the budget now being per-transition rather than per-run: a cancellation during a reset can take up to ~20 s to unwind rather than ~10 s. That is the intended consequence — each write gets a budget generous against the 1 s SQLite busy timeout — and the alternative is the defect this PR fixes.

CI green on `632a155f`. Mutation-verified independently by restoring the shared context on the two covered paths, which fails both subtests with the production error string.

Issue #1798, opened 2026-08-10
v0.34.0: opening a large store costs 10-18 s on every write-capable command, and the projection's 1 s wall budget cannot cover it


duck8823:
Found while verifying #1793 on the rehearsal copy of the live store. It is not a projection defect — it affects every command that opens the store for writing, and it is the reason `store search-projection resume` cannot run at its own default budgets.

## Measured

Same binary, same machine, two stores:

| store | `traceary list --limit 1` |
|---|---|
| 729 KB (freshly created) | **0.028 s** |
| 7.16 GB rehearsal copy | **10.3 - 17.9 s** |

The file size is not SQLite's problem. On the same 7.16 GB file:

| path | time |
|---|---|
| `sqlite3 <db> "SELECT 1"` | **0.016 s** |
| `traceary store gc --dry-run` (read-only connection, no initialize / migrate) | **0.31 s** |
| `traceary list --limit 1` (initializes) | **16.8 s** |

So the cost is entirely inside `Database.initializeAt` (`infrastructure/sqlite/database.go:348`), which every write-capable command runs on every invocation. Removing the `-wal`, `-shm` and `.traceary.lock` sidecars first changes it to 10.3-11.8 s, so it is not a stale-sidecar artefact.

It is also not CPU: 1.1 s user + 1.9 s system out of ~11 s wall. `sample` on the running process puts 390 of 471 main-thread samples in `pthread_cond_wait` and 41 in `pread`. The process spends its time waiting, not computing — which points at the lease/connection path or fsync behaviour rather than at query cost.

Ruled out by direct measurement:

- **The workspace-observation catch-up.** `catchUpWorkspaceObservations` runs on every initialize, but on this store it selects **0** rows and the equivalent scan in raw `sqlite3` costs 0.26-0.74 s.
- **The projection catch-up.** A store whose generation is parked (`rehearsal.db`, abandoned) costs 15.3 s and one with an active generation costs 17.9 s. The batch itself reports `last_batch_milliseconds: 371`.

Both catch-ups are sound designs; neither accounts for the time.

## Why it blocks the projection

`SearchProjectionUsecase.Resume` (`application/usecase/search_projection_usecase.go:66`) derives its context from `b.WallTime` and then opens the store inside it:

```go
wallCtx, cancel := context.WithTimeout(ctx, b.WallTime)
status, err := u.store.SearchProjectionStatus(wallCtx)
```

[...]

Issue #1794, opened 2026-08-10
v0.34.0: a single event 38 bytes over the decoded-bytes budget permanently fails the search projection


duck8823:
Second permanent stall found in the same #1620 gate rehearsal, with a different mechanism from #1793 and a worse property: no batch size can get past it.

## What happens

After working around #1793 with a wider lock cap, the projection walked to 36.1% and then failed:

```
Error: failed to execute CLI command:
       search projection decoded_bytes exceeds batch budget (8388646 > 8388608)
```

```
"state": "failed",  "failure_class": "decoded_bytes",
"checkpoint": 147584,  "high_water": 408970
```

The store contains exactly two rows the projection cannot ever index:

```
rowid   kind        body_plaintext_bytes   created_at
364431  transcript  8388646                2026-07-18T01:13:07.556403Z
364434  transcript  8388646                2026-07-18T01:13:08.016582Z
```

`--decoded-bytes` defaults to 8388608 (8 MiB). Both rows are **38 bytes over it**, and both are the same size to the byte, half a second apart — a host truncating a transcript at 8 MiB and traceary storing it inside a small envelope. This is a shape the corpus will keep producing, not a freak row.

## Why it is worse than #1793

#1793 stalls because a *batch* does not fit its lock cap; a smaller batch would fit. Here the unit is a single row, and the budget applies to that row. **There is no batch size that makes progress**, so the generation is not slow, it is impossible.

The recovery is also the same dead end. The decoded-bytes budget is part of the generation config hash, so `resume` refuses a wider one:

```
search projection made no progress: budget does not match generation configuration
```

The only way forward is `abort` + `start` with a larger `--decoded-bytes`, discarding the 36% already built — and then the next store that produces an 8 MiB transcript hits it again at whatever the new number is.

## What to decide

A row that cannot fit the per-batch budget at any batch size should not stop the walk. The options, roughly in order of how much they preserve:

1. **Project it partially.** The FTS document does not need the entire body to be useful; indexing the first N bytes of an over-budget row keeps the row findable and keeps the budget meaningful.
2. **Skip and record it.** Advance the checkpoint past the row, count it, and surface the count in `status` so the operator knows the projection is complete-with-exclusions rather than silently partial.
3. **Fail the row, not the generation.** Whatever the choice, the typed `failure_class: decoded_bytes` should name the offending row so the operator can act on it. Right now the error gives a number and no way to find which row produced it — I had to query `events` by `body_plaintext_bytes` to identify them.

[...]

PR #1799, opened 2026-08-10
fix: adaptively split oversized projection batches

### LINKED ISSUE #1793: v0.34.0: the search projection cannot be built on a real store — default lock cap stalls it permanently at 0.16%
Found while rehearsing the #1620 release gate against a copy of the live store. On a store with 408,970 source rows, the search projection cannot be built with the shipped defaults: it stalls permanently at 640 rows (0.16%) and never advances again.

Search is served by this projection, so this is a v0.34 blocker rather than a tuning note.

## Measured

Rehearsal store, 408,893 events, freshly compacted to 2.88 GiB, binary built from `50b953e9`. `store search-projection start` with defaults, then `resume --until-complete` in a loop. Every one of **400 consecutive invocations**:

```
iter=1   rc=1 state=rebuilding phase=source done=False ckpt=640 hw=408970
...
iter=400 rc=1 state=rebuilding phase=source done=False ckpt=640 hw=408970

Error: failed to execute CLI command:
       search projection made no progress: projection lock duration cap exceeded
```

The checkpoint is byte-identical across all 400 attempts. This is not slow progress; it is no progress.

## Why

Per-batch cost at the default `--rows 128` on this store, measured from `last_batch_milliseconds` over successful batches:

```
292 ms   199 ms   267 ms   365 ms   285 ms
```

The default `--lock-time` is **250 ms**. The routine cost of a default-sized batch is above it more often than not. The last batch that did commit before the stall reported `last_batch_milliseconds: 179`, already 72% of the cap.

With `--lock-time 5s --wall-time 10s` on a fresh generation the same store advances steadily, 2,560 rows per invocation, with no stalls.

So the cap is not protecting against a pathological row. It is below the normal cost of the work it bounds.

## Why it is permanent rather than slow

Three things compound:

1. **No adaptation.** `SearchProjectionUsecase.ResumeUntil` (`application/usecase/search_projection_usecase.go:143`) propagates the first batch error and stops. Nothing re-plans the batch with fewer rows, and `applyProjectionPlanWithRetry` (`infrastructure/sqlite/search_projection_rebuild.go:727`) only retries SQLite busy/snapshot conflicts *inside* the same cap. A batch that cannot fit is a hard stop, not a degraded step.

2. **The knob is refused where it is needed.** `resume` rejects a different budget on an existing generation:

   ```
   $ traceary store search-projection resume --until-complete --lock-time 5s
   Error: failed to execute CLI command:
          search projection made no progress: budget does not match generation configuration
   ```

   (`search_projection_usecase.go:76`.) The config hash covers the lock time, so widening it is only possible on a new generation.

3. **Recovery costs everything.** The only way forward is `abort` then `start` with a wider budget, which discards the whole generation. On this store that is fine because 640 rows is nothing; on a store that stalled at 60% it would not be.

The net effect: an operator whose store crosses whatever the threshold is gets a projection that is permanently stuck, an error that names a knob, and a knob the command refuses to accept.

## What to decide

The immediate question is whether the default `--lock-time` / `--rows` pairing is right. 250 ms against a measured 199-365 ms is not a margin, and the row count is what makes the batch expensive.

The structural question matters more: **a batch that does not fit its lock cap should shrink, not stop.** Halving the row count on a lock-cap failure and retrying would turn a permanent stall into a slower walk, and would make the default's tuning much less load-bearing. That also removes the need to widen a budget mid-generation, which is the thing the config hash currently forbids.

If neither is changed, then at minimum:

- the error should say which knob and that it requires `abort` + `start`, since `resume` will refuse it;
- `doctor` should detect a generation whose checkpoint has not moved across invocations, because right now nothing distinguishes "stalled forever" from "still working".

## Related

`#1751` (make the search projection's capacity derivation trustworthy enough to bound the store) is about the byte ceiling rather than the lock cap, but both are cases of a projection bound that behaves differently on a real store than on a fixture. The capacity evidence on this same store was also unavailable:

```
"capacity_evidence": {"status":"unavailable","method":"dbstat",
  "reason":"amplification sample below minimum; recent cutoff: context deadline exceeded"}
```

and `start` warned:

```
WARN search projection recent cutoff prefilter unavailable; retention is age-only until eviction
     reason="recent cutoff: context deadline exceeded"
```

duck8823:
Closes #1793.

The projection stalled permanently on a real store because a batch that did not fit the write-lock cap had no smaller unit of work to fall back to, and the obvious workaround was refused.

## What was wrong

`applyProjectionPlanWithRetry` retries only SQLite busy/snapshot conflicts. A batch that exceeds the lock deadline returns `SearchProjectionNoProgressError` and `ResumeUntil` propagates it and stops — the same batch, at the same size, forever.

The operator cannot resume with a smaller `--rows`, because `ConfigHash` folded every budget field into the generation identity, so any change to the batch size is read as a different generation and discards all durable progress.

## What changed

**The generation hash now covers only budgets that define generation contents** (`v3:<decodedBytes>:<recentAge>:<indexFamilyBytes>`). `Rows`, `WallTime`, `LockTime`, `StoredBytes` and `WriteBytes` bound one batch's working set and cannot change the resulting index — `StoredBytes` and `WriteBytes` can reject an oversized row, but rejection fails the generation outright rather than quietly admitting a different set of rows. `DecodedBytes` is held in by choice, not by that argument: #1794 decides what an over-budget row should do, and until then a widened budget should start a new generation rather than silently change the meaning of a partial one.

**A batch that exceeds the lock cap now halves and retries**, down to a floor of one row, and the next batch is planned at the caller's original size again — one fat row must not permanently shrink the walk. A single row that still does not fit terminates with a distinct code (`single_row_lock_duration_cap_exceeded`) so it is not confused with "the batch was merely too big". `SearchProjectionNoProgressError` gained a typed `Code` for this; comparing free-text reasons across the layer boundary would have been fragile.

**The inventory phase shrinks too.** Its lock-cap error was untagged, so it kept the stall this PR removes — and it is the first phase every existing store enters on upgrade, over its whole history, at the same default cap. Its raw-deadline path now goes through the same classifier as the source path rather than returning a bare driver error. Found by gpt-5.6-sol in review.

Identifying *which* row is out of scope here — that is #1794.

## Verified on the store that reproduced it

The rehearsal copy of the live store (408,970 rows, 7.16 GB), at the **default** `--lock-time 250ms` that previously stalled the walk permanently:

```
{"batches":10,"progress":{"selected":896,"written":896,...},"stop_reason":"max_batches"}
```

896 rows over 10 batches — below the 128-per-batch default, so the shrink engaged — and the checkpoint advanced 1792 → 2688 with `rc=0`. Before this change the same store returned `projection lock duration cap exceeded` on every invocation and never moved.

`--wall-time` had to be raised to 20s to run that at all, for a reason unrelated to this PR: opening a store this size costs 10-18 s and the 1 s default wall budget covers the open. Filed as **#1798**, which also means an end-to-end default-budget build cannot be demonstrated until that is fixed.

## Tests

Injectable rather than timing-dependent — a test that depends on how long a real SQLite write takes on a loaded runner is a flake, not a test.

- `TestSearchProjectionBudgetConfigHash` — table-driven over every budget field, asserting which ones change the generation identity and which do not.
- `TestSearchProjectionResumeUntilShrinksOnlyTheFailedBatch` — asserts the planned row sizes are `[4, 2, 4]`: the failed batch halves, the next one is back at full size.
- `TestSearchProjectionResumeUntilReportsSingleRowLockCap` — `[4, 2, 1]` then the distinct terminal code, rather than looping.
- `TestSearchProjectionResumeUntilShrinksInventoryBatch` — the same `[4, 2, 4]` for the inventory phase.
- `TestSearchProjectionApplyPathsTagLockDurationCap` — at the sqlite layer, because the usecase tests fake the store and so assert nothing about what it returns; that is how the missing inventory tag survived the first round.

Implementation by gpt-5.6-luna; review, the comment correction in `0ba5b68d`, and the store verification by Claude Opus 5.

### DIFF
--- a/application/types/search_projection.go
+++ b/application/types/search_projection.go
@@ -52,8 +52,15 @@ type SearchProjectionBudget struct {
 func (b SearchProjectionBudget) Valid() bool {
 	return b.Rows > 0 && b.WallTime > 0 && b.LockTime > 0 && b.StoredBytes > 0 && b.DecodedBytes > 0 && b.WriteBytes > 0 && b.RecentAge > 0 && b.IndexFamilyBytes > 0
 }
+
+// ConfigHash contains only budgets that define generation contents. DecodedBytes
+// remains semantic because it decides which rows are rejected; RecentAge and
+// IndexFamilyBytes define retention. Rows, WallTime, LockTime, StoredBytes and
+// WriteBytes only bound one batch's working set, so they must stay out of the
+// generation identity: a batch that cannot fit its lock cap must be allowed to
+// resume with a smaller unit of work without discarding durable progress.
 func (b SearchProjectionBudget) ConfigHash() string {
-	return fmt.Sprintf("v2:%d:%d:%d:%d:%d:%d:%d:%d", b.Rows, b.WallTime.Nanoseconds(), b.LockTime.Nanoseconds(), b.StoredBytes, b.DecodedBytes, b.WriteBytes, b.RecentAge.Nanoseconds(), b.IndexFamilyBytes)
+	return fmt.Sprintf("v3:%d:%d:%d", b.DecodedBytes, b.RecentAge.Nanoseconds(), b.IndexFamilyBytes)
 }
 
 type SearchProjectionGeneration struct {
@@ -225,7 +232,17 @@ func (e *SearchProjectionOversizeError) Error() string {
 	return fmt.Sprintf("search projection %s exceeds batch budget (%d > %d)", e.Class, e.Bytes, e.Limit)
 }
 
-type SearchProjectionNoProgressError struct{ Reason string }
+type SearchProjectionNoProgressCode string
+
+const (
+	SearchProjectionNoProgressLockDurationCap          SearchProjectionNoProgressCode = "lock_duration_cap_exceeded"
+	SearchProjectionNoProgressSingleRowLockDurationCap SearchProjectionNoProgressCode = "single_row_lock_duration_cap_exceeded"
+)
+
+type SearchProjectionNoProgressError struct {
+	Code   SearchProjectionNoProgressCode
+	Reason string
+}
 
 func (e *SearchProjectionNoProgressError) Error() string {
 	return "search projection made no progress: " + e.Reason
--- a/application/types/search_projection.go
+++ b/application/types/search_projection.go
@@ -53,12 +53,21 @@ func (b SearchProjectionBudget) Valid() bool {
 	return b.Rows > 0 && b.WallTime > 0 && b.LockTime > 0 && b.StoredBytes > 0 && b.DecodedBytes > 0 && b.WriteBytes > 0 && b.RecentAge > 0 && b.IndexFamilyBytes > 0
 }
 
-// ConfigHash contains only budgets that define generation contents. DecodedBytes
-// remains semantic because it decides which rows are rejected; RecentAge and
-// IndexFamilyBytes define retention. Rows, WallTime, LockTime, StoredBytes and
-// WriteBytes only bound one batch's working set, so they must stay out of the
-// generation identity: a batch that cannot fit its lock cap must be allowed to
-// resume with a smaller unit of work without discarding durable progress.
+// ConfigHash contains only budgets that define generation contents. RecentAge
+// and IndexFamilyBytes decide what the generation retains, so resuming with a
+// different value would leave an index that no single budget describes.
+//
+// Rows, WallTime, LockTime, StoredBytes and WriteBytes only bound one batch's
+// working set, so they stay out of the generation identity: a batch that cannot
+// fit its lock cap must be allowed to resume with a smaller unit of work rather
+// than discard durable progress. StoredBytes and WriteBytes can reject an
+// oversized row, but rejection fails the generation outright — it never quietly
+// admits a different set of rows — so they change no content either.
+//
+// DecodedBytes is held in by choice, not by that argument. It rejects rows the
+// same way, and #1794 decides what an over-budget row should do instead; until
+// that lands, keeping it here means a widened budget starts a new generation
+// rather than silently changing the meaning of a partial one.
 func (b SearchProjectionBudget) ConfigHash() string {
 	return fmt.Sprintf("v3:%d:%d:%d", b.DecodedBytes, b.RecentAge.Nanoseconds(), b.IndexFamilyBytes)
 }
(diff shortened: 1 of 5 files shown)


duck8823:

Two HIGH findings. One adopted, one rejected with evidence.

### Adopted — the inventory phase was not shrinkable (`8d1b24c3`, `0936a4e9`)

sol was right, and this was the more serious of the two. `ResumeUntil` shrinks only on `SearchProjectionNoProgressLockDurationCap`, and `Database.ApplyInventoryBatch` returned its lock-cap error untagged, so the inventory phase kept the exact stall this PR set out to remove. Worse placement than the source phase: inventory is the first thing every existing store does when it upgrades, over its whole history, at the same 250 ms default cap.

There was a second, quieter gap in the same function. The source path routes results through `classifySearchProjectionApplyResult`, which turns a raw `context.DeadlineExceeded` into the typed error; the inventory path's `if err == nil || !isSearchProjectionSQLiteBusy(err)` returned that deadline as a bare driver error, neither tagged nor retried. Both paths now share the classifier.

Checked that shrinking actually helps there before changing anything: `SelectInventory` bounds its batch by `b.Rows`, so halving the budget does halve the inventory batch.

Added a regression test at the layer that broke. The usecase-level tests fake `ApplyInventoryBatch` and therefore assert nothing about what sqlite returns — which is exactly why the missing tag survived the first round. `TestSearchProjectionApplyPathsTagLockDurationCap` calls both real write paths with a zero lock duration and asserts the typed code. Verified it fails without the production change:

```
--- FAIL: TestSearchProjectionApplyPathsTagLockDurationCap/inventory_batch
      types.SearchProjectionNoProgressCode(
    - 	"lock_duration_cap_exceeded",
    + 	"",
      )
```

### Rejected — the `v2` → `v3` config hash needs no migration

sol asked for compatibility handling for persisted `v2` hashes, on the grounds that an in-flight old generation cannot resume after upgrade and must be abandoned.

The premise does not hold: the search projection is unreleased, so no store in the wild holds a `v2` hash.

```
$ git show v0.33.1:application/types/search_projection.go
fatal: path '...' does not exist in 'v0.33.1'

$ git merge-base --is-ancestor 41837ff6 v0.33.1
NOT shipped in v0.33.1
```

`41837ff6` (2026-08-02) introduced the type after `v0.33.1` (2026-08-01). The only stores that can hold a `v2` hash are pre-release dogfooding copies, where `abort` + `start` is the correct answer and a compatibility shim would be permanent cost for a transient condition.

### Unchanged

`go build` / `go vet` / `go tool golangci-lint run` clean, `go test ./application/... ./infrastructure/...` green.

PR #1800, opened 2026-08-10
fix: keep whole-family measurement off the projection control path

### LINKED ISSUE #1798: v0.34.0: opening a large store costs 10-18 s on every write-capable command, and the projection's 1 s wall budget cannot cover it

duck8823:
Closes #1798.

`traceary list --limit 1` took 15 seconds on a 7 GB store. So did every other write-capable command, including ones that then did nothing.

## What was wrong

`Database.initializeAt` runs `catchUpSearchProjection` on every store open, and that reads `SearchProjectionStatus` before deciding whether there is any work. `SearchProjectionStatus` performs two whole-family scans:

| query | time on the 7.16 GB rehearsal store |
|---|---|
| `selectSearchProjectionFTSLogicalBytesSQL` | 6.26 s |
| `selectSearchProjectionFamilyTotalSQL` (`dbstat` walk) | 6.48 s |

Everything else in the open is noise by comparison — `migrate` 0.002 s, the compatibility check 0.000 s, the workspace-observation catch-up 1.1 s. The rehearsal store's generation is `failed`, so the catch-up parks it and does no work at all; it still paid 15 s to find that out.

The original issue ruled the projection catch-up out by indirect measurement. That was wrong, and the correction is in the issue thread: it is not the projecting that costs, it is the status read before it.

## What changed

The measured fields — `FTSLogicalBytes`, `PhysicalBytes`, `PhysicalEvidence`, `MatchProbeMilliseconds`, `InspectionMilliseconds`, and the per-generation aggregates — have exactly one consumer, `Inspect`, which backs the operator-facing `store search-projection status` and `probe` output. Every other caller reads persisted control columns only:

| caller | reads |
|---|---|
| `StartGeneration` | `State` |
| `Resume` | `State`, `Phase`, `ConfigHash` |
| `CatchUp` | `State`, `Phase`, `ConfigHash`, `CapacitySemanticsVersion`, `FailureClass`, cutover columns |
| `finishCatchUpProgress` | `State`, `Phase`, cutover columns |

Those four now use `SearchProjectionControlStatus`, one read of the two singleton rows in a single read-only transaction so a half-applied transition cannot be observed. `Inspect` keeps the measured read.

A narrower return type rather than a `measure bool`: the type is what stops a future control path from quietly reintroducing the scans, and it makes the usecase's dependency on the store honest about what it actually needs. No new command, subcommand or flag — the surface should shrink, not grow.

`finishCatchUpProgress` also read `statusAfter.Completed`, which is the SQL expression `state='complete'`, so it becomes `statusAfter.State == "complete"` rather than a field on the control type.

## Measured

`traceary list --limit 1` on the 7.16 GB rehearsal store, same binary flags, alternating runs:

| | run 1 | run 2 | run 3 |
|---|---|---|---|
| `main` | 15.08 s | 14.61 s | |
| this branch | 1.82 s | 0.85 s | 0.66 s |

`store search-projection resume` at its own 1 s `--wall-time` default now reaches its batch instead of expiring during the pre-batch status read.

## Tests

`TestSearchProjectionStatusRoutingKeepsMeasurementsOffControlPaths` counts control vs measured reads for `StartGeneration`, `Resume`, `CatchUp` (including its post-batch read) and `Inspect`. The compiler cannot hold this on its own: `SearchProjectionStatus` carries every field the control paths read, so routing one back to the measured call still compiles.

Review found the control read dropped `Method: "dbstat"` from the cutover evidence it copies into the catch-up result, which is JSON-serialised — the derivation is now a shared helper used by both reads.

## Not in this PR

The store open still sits inside `Resume`'s per-batch wall budget. With the status read down to milliseconds the 1 s default is workable, so that boundary is worth re-measuring rather than changing on the same speculation that produced the original diagnosis.

Implementation by gpt-5.6-luna; profiling, review and the store measurements by Claude Opus 5.

### DIFF
--- a/application/types/search_projection.go
+++ b/application/types/search_projection.go
@@ -328,6 +328,22 @@ type SearchProjectionStatus struct {
 	CutoverAfterEvidence  CapacityEvidence `json:"cutover_after_evidence"`
 }
 
+// SearchProjectionControlStatus contains only persisted state-machine data.
+// It deliberately excludes derived measurements so lifecycle operations cannot
+// accidentally put whole-family scans on their control path.
+type SearchProjectionControlStatus struct {
+	State                    string
+	Phase                    string
+	ConfigHash               string
+	CapacitySemanticsVersion int
+	FailureClass             string
+	CutoverIndexFamily       string
+	CutoverFamilyBytesBefore int64
+	CutoverFamilyBytesAfter  int64
+	CutoverBeforeEvidence    CapacityEvidence
+	CutoverAfterEvidence     CapacityEvidence
+}
+
 // SearchProjectionCatchUpResult is one bounded unit of automatic generation
 // work performed during store initialization. It mirrors the event-search
 // backfill shape: a single open does a bounded amount of work and resumes
--- a/application/usecase/search_projection_plan_test.go
+++ b/application/usecase/search_projection_plan_test.go
@@ -10,16 +10,21 @@ import (
 	"time"
 
 	"github.com/google/go-cmp/cmp"
+	"golang.org/x/xerrors"
 
 	apptypes "github.com/duck8823/traceary/application/types"
 	"github.com/duck8823/traceary/application/usecase"
 )
 
 type wallBudgetStore struct {
-	budget      apptypes.SearchProjectionBudget
-	applied     bool
-	delayStatus bool
-	selected    bool
+	budget             apptypes.SearchProjectionBudget
+	applied            bool
+	delayStatus        bool
+	delayControlStatus bool
+	selected           bool
+	controlReads       int
+	measuredReads      int
+	state              string
 }
 
 func (s *wallBudgetStore) Start(context.Context, apptypes.SearchProjectionBudget, time.Time) (apptypes.SearchProjectionGeneration, error) {
@@ -28,11 +33,25 @@ func (s *wallBudgetStore) Start(context.Context, apptypes.SearchProjectionBudget
 
 //nolint:wrapcheck // Test fake preserves context cancellation identity.
 func (s *wallBudgetStore) SearchProjectionStatus(ctx context.Context) (apptypes.SearchProjectionStatus, error) {
+	s.measuredReads++
 	if s.delayStatus {
 		<-ctx.Done()
 		return apptypes.SearchProjectionStatus{}, ctx.Err()
 	}
 	return apptypes.SearchProjectionStatus{State: "rebuilding", ConfigHash: s.budget.ConfigHash()}, nil
+}
+
+func (s *wallBudgetStore) SearchProjectionControlStatus(ctx context.Context) (apptypes.SearchProjectionControlStatus, error) {
+	s.controlReads++
+	if s.delayControlStatus {
+		<-ctx.Done()
+		return apptypes.SearchProjectionControlStatus{}, xerrors.Errorf("control status: %w", ctx.Err())
+	}
+	state := s.state
+	if state == "" {
+		state = "rebuilding"
+	}
+	return apptypes.SearchProjectionControlStatus{State: state, ConfigHash: s.budget.ConfigHash()}, nil
 }
 
 //nolint:wrapcheck // Test fake preserves context cancellation identity.
@@ -78,7 +97,7 @@ func TestProjectionBatchPlanEnforcesStrictLogicalMutationByteCap(t *testing.T) {
 func TestResumeStatusConsumesSameWallBudgetAndPreventsSelectionOrMutation(t *testing.T) {
 	t.Parallel()
 	b := apptypes.SearchProjectionBudget{Rows: 1, WallTime: time.Millisecond, LockTime: time.Second, StoredBytes: 1, DecodedBytes: 1, WriteBytes: 1, RecentAge: time.Hour, IndexFamilyBytes: 1}
-	store := &wallBudgetStore{budget: b, delayStatus: true}
+	store := &wallBudgetStore{budget: b, delayControlStatus: true}
 	_, err := usecase.NewSearchProjectionUsecase(store).Resume(context.Background(), b, time.Now())
 	if !errors.Is(err, context.DeadlineExceeded) {
 		t.Fatalf("error=%v", err)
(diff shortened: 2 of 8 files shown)


duck8823:

No CRITICAL or HIGH findings. sol independently confirmed the three things the split turns on: that no control caller reads a field the narrow type dropped (including `SelectInventory` / `SelectSnapshot`, which re-read what they need themselves), that `State == "complete"` is equivalent to the old `state='complete'` expression in every state the machine permits, and that the `JOIN` + `CASE` reproduces the old two-query inventory-phase derivation, including failing the same way when the inventory singleton is missing.

### Adopted — the routing test could not have caught the regression it guards (`b86b3589`)

sol's MEDIUM, and worth taking. The routing test counts which fake method each caller invokes; it never runs the real SQL. Adding an FTS aggregate or a `dbstat` subquery to `select_search_projection_control_status.sql` leaves the control-read count unchanged, so the test would still pass while the 15-second cost came straight back.

`TestSearchProjectionControlStatusQueryReadsOnlySingletonState` runs `EXPLAIN QUERY PLAN` over the real query against a real migrated store and asserts the set of objects the plan touches is exactly `search_projection_state` and `search_projection_inventory_state`.

It resolves aliases from the query text rather than matching plan wording, which is the part that makes it hold: SQLite reports `SCAN s`, not `SCAN search_projection_state`, so a plain name match would assert nothing. Asserting on the resolved set rather than a denylist of forbidden tables also means a table added to the schema later is caught without anyone remembering to extend a list.

Verified it fails on the regression it exists for, by temporarily adding a `dbstat` sum to the query:

```
--- FAIL
    plan:
    SEARCH s USING INTEGER PRIMARY KEY (rowid=?)
    SEARCH i USING INTEGER PRIMARY KEY (rowid=?)
    SCALAR SUBQUERY 1
    SCAN dbstat VIRTUAL TABLE INDEX 0x2:
```

### Found while verifying, filed separately

`store search-projection resume` at defaults now reaches its batch, but the **automatic** catch-up still stalls permanently on the rehearsal store: #1793's adaptive shrink lives in `ResumeUntil`, and `CatchUp` calls `Resume` directly. Filed as **#1801** rather than widened into this PR.

PR #1803, opened 2026-08-10
fix: let the automatic projection catch-up shrink its batch

### LINKED ISSUE #1801: v0.34.0: the automatic projection catch-up still stalls permanently — the adaptive shrink is only in ResumeUntil
#1793 fixed the permanent stall for the operator, but not for the user. The adaptive shrink lives in `ResumeUntil`, and the automatic per-open catch-up does not go through it.

## Measured

On the gate rehearsal store (`identity.db`, 408,970 events, generation `rebuilding` / `source`, checkpoint 2688), with #1793 and #1798 both applied:

```
$ traceary list --limit 1
{"level":"ERROR","msg":"search projection catch-up incomplete; retrying on next initialization",
 "action":"resume","state":"rebuilding","phase":"source","batches":0,"selected":0,"written":0,
 "error":"search projection catch-up: search projection made no progress: projection lock duration cap exceeded"}
```

Every invocation, `batches: 0`. The generation never advances on its own. Meanwhile the operator path works:

```
$ traceary store search-projection resume --until-complete --wall-time 20s
{"batches":10,"progress":{"selected":896,...}}
```

## Why

`ResumeUntil` (`application/usecase/search_projection_usecase.go:163`) is the only caller that halves `Rows` on `SearchProjectionNoProgressLockDurationCap`. Two call sites bypass it:

- `CatchUp` → `u.Resume` at `:312` and `:374` — the automatic path, on every store open
- `store search-projection resume` without `--until-complete` → `Resume` at `presentation/cli/store_search_projection.go:85`

So the case #1793 was filed for — "the projection cannot advance on a real store" — is still true for anyone who does not know to pass `--until-complete`, which is everyone who never runs the command at all. The catch-up is the mechanism the release relies on to build the projection without operator action.

## What to decide

The obvious repair is for the catch-up to use the same adaptive path rather than a single fixed-size attempt. Worth confirming first:

- Whether a shrink loop belongs inside `Resume` itself rather than in `ResumeUntil`, so every caller inherits it and there is one behaviour to reason about instead of two. The single-batch CLI `resume` is the case against — an operator asking for one batch at a stated size may want to be told the size does not fit, not to be silently given a smaller one.
- The cost ceiling on the automatic path. Halving 128 rows down to 1 is at most eight attempts at the 250 ms lock cap, so roughly 2 s added to a store open in the worst case. That is acceptable next to today's 15 s, but it should be a deliberate bound rather than a consequence.
- Whether the terminal `single_row_lock_duration_cap_exceeded` should park the generation rather than retry the same failure on every subsequent open.

## Related

- #1793 — the shrink this is missing.
- #1798 — the store-open cost that hid this; the catch-up failure only became visible once opens got cheap enough to run repeatedly.
- #1794 — the other reason a generation can fail permanently.
- #1620 — the release gate. The projection cannot be demonstrated as self-building until this is fixed.

duck8823:
Closes #1801.

#1793 fixed the permanent projection stall for the operator, but not for the user. The adaptive shrink lived in `ResumeUntil`, and the automatic per-open catch-up called `Resume` directly — so the generation still never advanced on its own.

## What was wrong

On the 7.16 GB rehearsal store, with #1793 and #1798 merged, every single invocation:

```
$ traceary list --limit 1
{"level":"ERROR","msg":"search projection catch-up incomplete; retrying on next initialization",
 "action":"resume","phase":"source","batches":0,"selected":0,"written":0,
 "error":"search projection catch-up: search projection made no progress: projection lock duration cap exceeded"}
```

`batches: 0`, forever. `store search-projection resume --until-complete` worked, but that is not the mechanism this release relies on to build the projection — the catch-up is, precisely because it needs no one to run a command.

## What changed

`CatchUp` now goes through the same adaptive loop as the operator path, via a single-batch `ResumeUntil`. One implementation of "halve on lock cap": a second one that could drift from the first would be worse than the bug.

The wall ceiling is derived rather than assumed — `bits.Len(uint(b.Rows)) * b.LockTime`, the halving floor expressed in wall time. A hard-coded 8 would have been correct only because `defaultSearchProjectionCatchUpBudget` happens to use 128 rows; a caller with a different budget would have been cut off mid-shrink by the wall clock and reported the wrong error.

**The generation is deliberately not parked on `single_row_lock_duration_cap_exceeded`.** Parking is right for an oversized row, which fails deterministically. A lock-duration cap can also come from another process holding the write lock, and permanently disabling the projection over a moment of contention is worse than retrying on the next open. There is a comment saying so, because the opposite change looks like an improvement.

The single-batch `store search-projection resume` stays strict: an operator asking for one batch at a stated size should be told the size does not fit, not silently given a smaller one.

## Verified on the store that reproduced it

Three consecutive `traceary list --limit 1` runs on `identity.db`, no operator command involved:

| | before | after |
|---|---|---|
| catch-up result | `batches: 0` + ERROR, every run | advances, no error |
| checkpoint | 2688, unchanged | 2688 → 2816 |
| wall time | — | 1.09 s / 1.80 s / 2.61 s |

## Tests

- `TestSearchProjectionCatchUpShrinksFailedBatch` — the catch-up plans `[4, 2]` and commits the smaller batch.
- `TestSearchProjectionCatchUpBoundsAdaptiveAttempts` — `[128, 64, 32, 16, 8, 4, 2, 1]` then the distinct single-row code. This assertion is what holds the bound; the derived ceiling has nothing else keeping it honest.
- A non-default `Rows: 256` budget still admits its full shrink sequence, so the derivation cannot silently stop matching the floor.
- `TestSearchProjectionResumeDoesNotShrinkSingleBatch` — the operator's single-batch path still surfaces the failure.

Implementation by gpt-5.6-luna; review and the store verification by Claude Opus 5.

### DIFF
--- a/application/usecase/search_projection_plan_test.go
+++ b/application/usecase/search_projection_plan_test.go
@@ -163,11 +163,12 @@ type multiBatchProjectionStore struct {
 }
 
 type adaptiveProjectionStore struct {
-	budget       apptypes.SearchProjectionBudget
-	maxRows      int
-	checkpoints  []int64
-	plannedRows  []int
-	alwaysTooBig bool
+	budget          apptypes.SearchProjectionBudget
+	maxRows         int
+	checkpoints     []int64
+	plannedRows     []int
+	alwaysTooBig    bool
+	catchUpDeadline time.Time
 }
 
 func (s *adaptiveProjectionStore) Start(context.Context, apptypes.SearchProjectionBudget, time.Time) (apptypes.SearchProjectionGeneration, error) {
@@ -187,7 +188,12 @@ func (s *adaptiveProjectionStore) SelectSnapshot(_ context.Context, b apptypes.S
 	}
 	return apptypes.ProjectionSnapshot{Generation: apptypes.SearchProjectionGeneration{GenerationID: "g", ConfigHash: s.budget.ConfigHash(), SourceRevision: 1, HighWater: 3, Checkpoint: s.checkpoint()}, Phase: "source", Documents: documents, SourceDone: len(documents) > 0 && documents[len(documents)-1].Sequence == 3, Now: time.Now().UTC()}, nil
 }
-func (s *adaptiveProjectionStore) ApplyBatch(_ context.Context, p apptypes.ProjectionBatchPlan, _ time.Duration, _ time.Time) (apptypes.SearchProjectionProgress, error) {
+func (s *adaptiveProjectionStore) ApplyBatch(ctx context.Context, p apptypes.ProjectionBatchPlan, _ time.Duration, _ time.Time) (apptypes.SearchProjectionProgress, error) {
+	if s.catchUpDeadline.IsZero() {
+		if deadline, ok := ctx.Deadline(); ok {
+			s.catchUpDeadline = deadline
+		}
+	}
 	if len(p.Writes) > s.maxRows || s.alwaysTooBig {
 		return apptypes.SearchProjectionProgress{}, &apptypes.SearchProjectionNoProgressError{Code: apptypes.SearchProjectionNoProgressLockDurationCap, Reason: "projection lock duration cap exceeded"}
 	}
@@ -309,8 +315,9 @@ func TestSearchProjectionCatchUpShrinksFailedBatch(t *testing.T) {
 
 func TestSearchProjectionCatchUpBoundsAdaptiveAttempts(t *testing.T) {
 	t.Parallel()
-	b := apptypes.SearchProjectionBudget{Rows: 128, WallTime: time.Second, LockTime: 250 * time.Millisecond, StoredBytes: 100, DecodedBytes: 100, WriteBytes: 1000, RecentAge: time.Hour, IndexFamilyBytes: 100}
+	b := apptypes.SearchProjectionBudget{Rows: 256, WallTime: time.Second, LockTime: 50 * time.Millisecond, StoredBytes: 100, DecodedBytes: 100, WriteBytes: 1000, RecentAge: time.Hour, IndexFamilyBytes: 100}
 	store := &adaptiveProjectionStore{budget: b, alwaysTooBig: true}
+	started := time.Now()
 	_, err := usecase.NewSearchProjectionUsecase(store).CatchUp(context.Background(), b, time.Now())
 	var noProgress *apptypes.SearchProjectionNoProgressError
 	if !errors.As(err, &noProgress) {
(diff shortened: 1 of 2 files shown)


duck8823:

No CRITICAL or HIGH findings. Three MEDIUM observations; one adopted, two recorded.

### Adopted — the catch-up could report a batch that never happened (`0532cdc6`)

The best of the three, and a bug this PR introduced. `ResumeUntil` returns `StopReason = "total_wall_time"` with a **nil error and zero progress** when its wall bound expires. `resumeCatchUpBatch` discarded the stop reason and returned only `result.Progress`, so `CatchUp` saw success and `finishCatchUpProgress` set `Batches = 1` unconditionally. A catch-up that committed nothing was logged as one completed batch. Before this PR the same path returned an error.

sol also showed it was reachable with the production budget, not just in theory: the ceiling was `bits.Len(128) * LockTime` = 2 s, but a failed attempt costs more than its lock wait — status read, snapshot selection, hydration, planning, transaction setup all sit outside `LockTime`. Seven failed attempts could consume the whole ceiling before the one-row attempt ran, which also swallows the `single_row_lock_duration_cap_exceeded` diagnosis that attempt exists to produce.

Two changes:

- No committed batch and no error now returns a `SearchProjectionNoProgressError` naming the exhausted bound. The caller already logs a returned error as "catch-up incomplete; retrying on next initialization", which is the accurate thing to say.
- The ceiling's unit is now `b.WallTime`, not `b.LockTime`. `Resume` bounds the whole attempt with `WallTime`, so a total built from `LockTime` was smaller than one attempt is allowed to take — incoherent, and the reason the sequence could be truncated before reaching the floor it exists to reach. The comment says why, so the tighter-looking value does not come back.

Existing tests used a fake that failed instantly and so covered neither path; both are now covered.

Re-verified on `identity.db` after the change: checkpoint 2816 → 2944 over two `traceary list --limit 1` runs, no error, 1.12 s and 2.14 s.

### Recorded, not fixed — a deterministically slow row retries forever (#1804)

sol is right that the typed error cannot tell "lost a contention race" from "this row's own work exceeds the cap every time", and that the second case makes every store open repeat the whole shrink sequence indefinitely.

Fixing it here would mean either parking on the first single-row failure — which reintroduces exactly the fragility the no-parking decision avoids, one contended moment permanently disabling the projection — or adding state across opens, or separating lock-wait time from work time. Those are design decisions, and they overlap #1794, which already needs an answer to "one row cannot be projected under the current budget, and the operator needs to be told which and why". Filed as **#1804** so the two are decided together.

### Not fixed — duration overflow on an extreme budget

`bits.Len(uint(b.Rows)) * duration` can overflow for a budget with an enormous `WallTime`, since `Valid()` sets no upper bound. Not reachable from `defaultSearchProjectionCatchUpBudget` or from any CLI flag path — only by constructing such a budget programmatically. Guarding it would add a branch for a caller that does not exist.

PR #1805, opened 2026-08-11
fix: skip and record oversized search projection rows instead of failing the generation

### LINKED ISSUE #1794: v0.34.0: a single event 38 bytes over the decoded-bytes budget permanently fails the search projection

duck8823:
Closes #1794

## The defect

A single event over a **per-row** budget failed the entire generation. On the 7 GB rehearsal store the walk reached 36.1% and stopped on two transcripts of exactly 8,388,646 bytes — 38 bytes over the 8 MiB `DecodedBytes` default, both already truncated at 8 MiB by the host. No batch size helps: the unit is one row. Recovery meant `abort` + `start` with a wider budget, discarding everything built, and the next 8 MiB transcript hits the same wall at the new number.

## The design — skip, record, advance

Decided after a design consultation with gpt-5.6-sol on the alternative (index a truncated prefix). Truncation was rejected because it breaks three things in this codebase:

- `projectionTokens` would emit tokens cut mid-word that do not exist in the source, and session keyword hits have **no canonical re-verification**, so a truncated row produces false positives nothing catches.
- A partial fingerprint set reads as a complete one. `literal_search_fingerprints` is fail-open only when a row is *absent*; rows present and non-matching are dropped before canonical verification.
- The benefit was illusory: normal event search does not treat the recent FTS as its authority — it walks `events` and verifies candidates against the canonical body. An excluded row stays in `search_projection_source_sequence` and remains findable. Excluding it means "do not build an incomplete accelerator entry", not "remove the row from search".

So the row is detected at metadata admission **before hydration**, skipped, recorded, and the checkpoint advances past it.

- An explicit `excluded` disposition, distinct from `deleted` — an excluded row is present and deliberately unindexed, and conflating the two would lose that in every future reader.
- No document, no summary contribution, no keywords and specifically **no fingerprint** for an excluded row.
- The exclusion record and the checkpoint advance commit in the same transaction (`search_projection_exclusions`, migration `000057`, keyed by generation and source sequence).
- Exclusions are cleaned up with their generation, by both the cleanup phase and `abort`.
- `status` reports `exclusion_count` and up to 20 named rows for the active generation, on the **measured** read only — #1798 exists because measurement leaked onto the control path.
- All three per-row classes are covered: `stored_bytes`, `decoded_bytes`, `write_bytes`. Batch-level oversize errors are unchanged; `markFailed` is still right for those.

## Review findings fixed before this PR opened

Two defects were reproduced during review of the first implementation and fixed in `4ffac7c2`:

- **A `decoded_bytes` exclusion was recorded as `stored_bytes`.** The disposition was the first disjunct of the classifier and `AdmitSource` does not say which budget tripped, so every upstream exclusion was labelled `stored_bytes` and the `decoded_bytes` branch was unreachable. A document `{stored: 5, decoded: 11}` against `{stored: 1000, decoded: 10}` produced `{class: stored_bytes, measured: 5, limit: 1000}` — a durable record asserting 5 bytes exceeded a 1000-byte limit, on exactly the row #1794 exists for. The class is now derived from the measured values, and an `excluded` row matching no class is an error rather than a silent empty document.
- **An excluded row still fed the session summary chain, which cascaded exclusions onto later rows.** The running summary map was written before the `write_bytes` check. A row of `Text: "tiny"` alone is admitted; preceded in the same session by one excluded 148-character row it measured 269 bytes against a 260 limit and was excluded too. The map is now written only where the write is accepted.

Three more, also fixed:

- The `len(p.Writes) == 0` branch of the write-bytes check was **not** dead — `LogicalWriteBytes` is seeded with the checkpoint mutation — and it was the last per-row landmine: a row fitting under `WriteBytes` alone but not with the fixed checkpoint cost still failed the generation deterministically, discarding any exclusions already planned. The per-row test now includes the fixed cost and records an exclusion; what remains is a plain batch-full `break`.
- `source_bytes` was renamed to `measured_bytes`: for the `write_bytes` class the value is a projected write size, not a source byte count. Both the migration and the JSON are unreleased, so this was free now.
- Exclusion INSERTs are charged to the ledger, so `WriteBytes` still bounds what the transaction actually mutates.

## Tests

Boundary (exactly at each limit admitted, one byte over excluded), each of the three classes with its own numbers, no summary chaining from an excluded row, no fingerprint row for an excluded row, exclusion records removed by `abort`, and the generation reaching `complete` with the oversized row recorded.

`go build ./...`, `go vet ./...`, `go tool golangci-lint run` (0 issues) and `go test ./...` pass.

### DIFF
--- a/application/types/search_projection.go
+++ b/application/types/search_projection.go
@@ -152,6 +152,7 @@ const (
 	ProjectionDispositionAdmitted ProjectionDisposition = "admitted"
 	ProjectionDispositionDeleted  ProjectionDisposition = "deleted"
 	ProjectionDispositionExcluded ProjectionDisposition = "excluded"
+	ProjectionDispositionBatchFull ProjectionDisposition = ""
 )
 
 type ProjectionSnapshot struct {
@@ -195,7 +196,7 @@ type ProjectionWrite struct {
 }
 
 type ProjectionExclusion struct {
-	Sequence, SourceBytes, ByteLimit int64
+	Sequence, MeasuredBytes, ByteLimit int64
 	EventID, Class                   string
 }
 
@@ -241,7 +242,7 @@ func (l *BudgetLedger) AdmitSource(b SearchProjectionBudget, stored, decoded int
 		return ProjectionDispositionExcluded, nil
 	}
 	if !l.ReserveSource(b, stored, decoded) {
-		return "", nil
+		return ProjectionDispositionBatchFull, nil
 	}
 	return ProjectionDispositionAdmitted, nil
 }
@@ -354,7 +355,7 @@ type SearchProjectionExclusion struct {
 	Sequence int64 `json:"sequence"`
 	EventID string `json:"event_id"`
 	Class string `json:"class"`
-	SourceBytes int64 `json:"source_bytes"`
+	MeasuredBytes int64 `json:"measured_bytes"`
 	ByteLimit int64 `json:"byte_limit"`
 }
 
--- a/application/types/search_projection.go
+++ b/application/types/search_projection.go
@@ -53,23 +53,17 @@ func (b SearchProjectionBudget) Valid() bool {
 	return b.Rows > 0 && b.WallTime > 0 && b.LockTime > 0 && b.StoredBytes > 0 && b.DecodedBytes > 0 && b.WriteBytes > 0 && b.RecentAge > 0 && b.IndexFamilyBytes > 0
 }
 
-// ConfigHash contains only budgets that define generation contents. RecentAge
-// and IndexFamilyBytes decide what the generation retains, so resuming with a
-// different value would leave an index that no single budget describes.
+// ConfigHash contains only budgets that define generation contents. StoredBytes,
+// DecodedBytes, WriteBytes, RecentAge and IndexFamilyBytes decide what the
+// generation contains, so resuming with a different value would leave an index
+// that no single budget describes.
 //
-// Rows, WallTime, LockTime, StoredBytes and WriteBytes only bound one batch's
-// working set, so they stay out of the generation identity: a batch that cannot
-// fit its lock cap must be allowed to resume with a smaller unit of work rather
-// than discard durable progress. StoredBytes and WriteBytes can reject an
-// oversized row, but rejection fails the generation outright — it never quietly
-// admits a different set of rows — so they change no content either.
-//
-// DecodedBytes is held in by choice, not by that argument. It rejects rows the
-// same way, and #1794 decides what an over-budget row should do instead; until
-// that lands, keeping it here means a widened budget starts a new generation
-// rather than silently changing the meaning of a partial one.
+// Rows, WallTime and LockTime only bound the cost of one batch, so they stay out
+// of the generation identity: a batch that cannot fit its lock cap must be
+// allowed to resume with a smaller unit of work rather than discard durable
+// progress.
 func (b SearchProjectionBudget) ConfigHash() string {
-	return fmt.Sprintf("v3:%d:%d:%d", b.DecodedBytes, b.RecentAge.Nanoseconds(), b.IndexFamilyBytes)
+	return fmt.Sprintf("v4:%d:%d:%d:%d:%d", b.StoredBytes, b.DecodedBytes, b.WriteBytes, b.RecentAge.Nanoseconds(), b.IndexFamilyBytes)
 }
 
 type SearchProjectionGeneration struct {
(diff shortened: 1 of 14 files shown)


duck8823:

Two review rounds. Round 1 (`4af36d2a`): no CRITICAL, three HIGH, all three adopted. Round 2 on the delta (`13a2963a`): 問題なし.

Five findings from my own review of the first implementation are described in the PR body; this covers the three from sol.

### Adopted — `StoredBytes` and `WriteBytes` had to join the generation identity (`62b8b178`)

The best of the three. The comment on `ConfigHash` justified leaving them out like this:

> StoredBytes and WriteBytes can reject an oversized row, but rejection fails the generation outright — it never quietly admits a different set of rows — so they change no content either.

This branch makes that false: rejection now quietly admits a different set of rows. The same comment already named #1794 as the event that would decide it for `DecodedBytes`.

Without the fix, a generation's contents depend on the budget history of its resumes — start with `--stored-bytes 8MiB` and index a 6 MiB row, resume with `--stored-bytes 4MiB` and the next 6 MiB row is excluded permanently. Neither budget applied from the start produces that index.

`ConfigHash` is now `v4` over `StoredBytes`, `DecodedBytes`, `WriteBytes`, `RecentAge` and `IndexFamilyBytes`. `Rows`, `WallTime` and `LockTime` stay out, so #1803's adaptive shrink — which halves `Rows` — still resumes an existing generation instead of discarding durable progress. `v3` was introduced earlier in this same unreleased cycle, so this costs released stores nothing they were not already paying.

### Adopted — an exclusion record that does not fit was a new permanent stall (`62b8b178`)

When the checkpoint fits but the checkpoint plus the exclusion record does not, the planner broke and returned `NoProgressError`, which does not mark the generation failed. The same batch would then be retried on every store open forever, checkpoint never moving, with no error state and one log line per open. sol put numbers on it: about 94 bytes for a source-phase checkpoint and 112 for a `stored_bytes` exclusion, so `--write-bytes 180` with an oversized first row stalls permanently, and neither `Valid()` nor the CLI enforces a floor.

Narrow, but it is a new instance of exactly what #1794 removes: a deterministic condition that silently disables the projection. When no progress of any kind is possible it now returns a batch-level `SearchProjectionOversizeError{Class: "write_bytes"}`, so the operator gets `failed` with the budget to raise instead of an invisible retry loop. When writes or exclusions were already planned, the batch still commits and the row is retried next time.

### Adopted — `status` reported one generation's exclusions beside another's state (`62b8b178`)

Exclusions were keyed on `active_generation_id` while every other status field describes the current `generation_id`. During a rebuild those are different generations, so the operator read `state=rebuilding` with generation B's checkpoint next to generation A's exclusion count — and B's own exclusions stayed invisible until cutover, which is when they matter most. Now keyed on `generation_id`, with `exclusion_generation_id` in the payload so the scope is never inferred.

### Cleared

sol confirmed the exclusion contract is complete (no document, summary, aggregate, keyword or fingerprint survives an excluded row, and nothing propagates into a later batch's summary), that a duplicate exclusion INSERT is unreachable because the checkpoint fence precedes it inside the same transaction, that the classifier's error path is unreachable on the real path and correct as a fail-closed invariant, and that no path orphans exclusion rows.

### Filed, not fixed here — #1806

Widening the hash means the CLI flag defaults and `defaultSearchProjectionCatchUpBudget` must now agree on five fields rather than three; they do today, and nothing pins it. A drift would make `resume` silently refuse the generation the automatic catch-up started. That wants one owner for the defaults rather than a test, so it is its own change.

`go build` / `go vet` / `go tool golangci-lint run` (0 issues) / `go test ./...` all green.

Issue #1796, opened 2026-08-10
v0.34.0: a search projection rebuild peaks at ~7x its index-family budget because eviction only runs after the whole source walk


duck8823:
> **Decided 2026-08-11** ([issuecomment-5248947989](https://github.com/duck8823/traceary/issues/1796#issuecomment-5248947989)). `IndexFamilyBytes` bounds the **evictable recent tier only** — `search_projection_recent_documents`, its indexes, and `search_projection_recent_fts_*`. The session tier and `literal_search_fingerprints` are corpus-proportional by design and must be **documented** as such, not silently counted inside a ceiling that never bounded them (measured: they are already 80.5% of the ceiling at 36% of the walk — [issuecomment-5248363164](https://github.com/duck8823/traceary/issues/1796#issuecomment-5248363164)).
>
> What remains in scope here: **interleave eviction with the source walk**, so the ceiling bounds the peak rather than the post-hoc total, and document the corpus-proportional tiers. Split out: the cutoff derivation's timeout (#1807), the duplicated autoindexes (#1808), the flag rename the narrowed scope forces (#1810), the fallback amplification constant (#1751). The gate this feeds is now stated per-record, not in GiB (#1620).

Measured during the #1620 gate rehearsal, on a copy of the live store (534k events, 24.7 GiB before compaction, 2.87 GiB after every Wave 3 lever). It is a separate failure from #1793 and #1794: those stop the walk, this one is what the walk *costs* when it does run.

## What was measured

The generation was started with defaults — `--index-family-bytes 1535115264` (1464 MiB), `--recent-age 720h`. The derivation recorded:

```
recent_source_ceiling_bytes = 690432000     (0.64 GiB of source text)
recent_amplification_ppm    = 2160000       (2.16x, the fallback)
non_recent_family_bytes     = 43782144
recent_cutoff_norm          = ''            (no byte cutoff)
capacity_evidence_status    = unavailable
capacity_evidence_reason    = amplification sample below minimum; recent cutoff: context deadline exceeded
```

At **checkpoint 147,584 of high_water 408,970 — 36.1% of the source walk** — the store held:

| structure | bytes |
|---|---|
| `search_projection_recent_fts_data` | 1,849,626,624 |
| `search_projection_recent_documents` | 878,751,744 |
| `search_projection_session_keywords` (+autoindex) | 826,871,808 |
| `literal_search_fingerprints` (+2 indexes) | 345,440,256 |
| other projection tables | 116,875,264 |
[...]

PR #1813, opened 2026-08-11
docs: restate the sprint spec's landing targets in gate-safe units

### LINKED ISSUE #1812: v0.34.0: align the sprint spec's landing targets with the restated gate units
Mechanical follow-up to the gate restatement on #1620.

`.ai/spec/1606.md` carries the sprint's 着地目標 table, and it still states the withdrawn units:

```
| ストア総量 | 24.71 GiB | ≤ 4 GiB |
| 月あたり増分 | +9.9 GiB | ≤ +0.5 GiB |
| 索引ファミリ | 15.83 GiB (64.1%) | ≤ 2 GiB |
```

`GiB` and `GiB/月` were withdrawn as gate units on 2026-08-11 ([#1620 issuecomment-5249722984](https://github.com/duck8823/traceary/issues/1620#issuecomment-5249722984)), and per-*event* units were withdrawn immediately after ([issuecomment-5249869562](https://github.com/duck8823/traceary/issues/1620#issuecomment-5249869562)). The spec is the document every remaining v0.34 issue is written against, so leaving the old table in it means the sprint's own source of truth contradicts its gate.

## Done when

- The 着地目標 table in `.ai/spec/1606.md` matches the revised table on #1620: canonical-operation denominators, the event-emission amplification guard, the peak-during-rebuild row, and the statement that thresholds are regression tripwires rather than capacity promises.
- The remaining GiB figures in the same section are labelled as reference-workload illustrations rather than targets.
- No behaviour change and no code change. Spec only. `.ai/` is gitignored, so the file needs `git add -f`.

Parent: #1606

duck8823:
Spec-only. No code, no behaviour change.

`.ai/spec/1606.md` is the document every remaining v0.34 issue is written against, and its 着地目標 table still stated the gate in units that #1620 withdrew:

| withdrawn | why |
|---|---|
| `≤ 4 GiB` store, `≤ +0.5 GiB/month` | `GiB/month = bytes-per-unit-of-work x work-per-month`; only the first factor is a property of Traceary. A heavy user's copy breaks the promise on day one; a light user's satisfies it while carrying a 10x per-record regression. |
| per-*event* units (the first replacement) | the producer decides what an event is. Splitting one operation into ten improves bytes/event while total storage rises — the same defect one level down. |

The table now uses **canonical operations** — one `command_executed` or one `prompt`, units of the operator's work that Traceary does not control — plus the two rows an adversarial review by gpt-5.6-sol found missing:

- **event-emission amplification** (events / canonical operation, measured 1.31, gate ≤ 2.0), which is the guard that makes denominator dilution a visible regression instead of an improvement;
- **peak resident during rebuild** (peak / steady state, measured **2.30x against a 1.5x gate — fails**), which is the row #1796 exists to fix and which the previous table said nothing about.

Thresholds are labelled for what they are: `measured value x declared margin (1.3-1.5x)`, published as regression tripwires rather than capacity promises. No derivation makes a specific byte count correct, and saying so is more useful than dressing a chosen number as a law. Remaining GiB figures are kept as reference-workload illustrations with their corpus attached.

Measurements are from the rehearsal copy at `/private/tmp/traceary-gate/rehearsal.db` on 2026-08-11 (30-day window: 228,595 canonical operations, 300,282 events, 19,237 sessions), never the live store from a repo build.

## Verification

- Spec file only; `go build`, `go test` and lint are unaffected by construction (no `.go`, no schema, no docs shipped in the module).
- `git diff` confirms a single file changed.

Full reasoning: [#1620 issuecomment-5249722984](https://github.com/duck8823/traceary/issues/1620#issuecomment-5249722984) (unit change), [issuecomment-5249793468](https://github.com/duck8823/traceary/issues/1620#issuecomment-5249793468) (threshold justification), [issuecomment-5249869562](https://github.com/duck8823/traceary/issues/1620#issuecomment-5249869562) (the review that broke three rows and the revision).

Closes #1812


duck8823:
## Review

**gpt-5.6-sol (adversarial, read-only)** — 1 HIGH, fixed in `90118036`:

> HIGH — the claim that every threshold equals the measured value times a declared 1.3-1.5x margin is false. `3x / 2.55x = 1.18x` and `4x / 3.37x = 1.19x`; no per-row margins are declared.

Correct and material: it was exactly the kind of tidy-sounding blanket statement this restatement exists to remove. The sentence now states the actual spread — 1.3-1.55x on byte-count rows, 1.18-1.19x on the two already-normalised amplification rows — and calls out that the peak-during-rebuild row is not a margin at all, since its target sits *below* the measurement. The same correction was applied to #1620's body and recorded in [issuecomment-5249980852](https://github.com/duck8823/traceary/issues/1620#issuecomment-5249980852).

The substance of the table itself was reviewed by the same model before it was written down ([#1620 issuecomment-5249869562](https://github.com/duck8823/traceary/issues/1620#issuecomment-5249869562)), which is where the canonical-operation denominator, the event-emission guard and the peak row came from — three findings that broke the first restatement.

**Claude Opus 5** — verified each margin arithmetically against the recorded measurements; the corrected sentence matches. No other claim in the changed hunk contradicts the surrounding spec.

**CI**: Docs / Integrations / Lint / Test all pass on `4525b232`; re-running on `90118036`.

PR #1814, opened 2026-08-11
fix: order sessions and rehearsal runs by normalized timestamps

### LINKED ISSUE #1729: v0.34-52: order sessions by ts_norm(started_at), not raw text
`ts_norm()` is applied to `started_at` **filters** but not to `started_at` **ORDER BY** in the session listing path, so the #1185 hazard is still reachable there.

`started_at` is written as `time.RFC3339Nano`, which is variable width: `2026-01-01T00:00:00Z` and `2026-01-01T00:00:00.5Z` differ in length, and `'.'` (0x2E) sorts below `'Z'` (0x5A). Plain TEXT comparison therefore orders the *later* timestamp *first*, which is why every comparison is supposed to go through `ts_norm()` on both sides.

### Sites

- `infrastructure/sqlite/sql/list_sessions.sql:18` — `ORDER BY s.started_at DESC`, directly followed by `LIMIT ? OFFSET ?`. Lines 16-17 in the same query already use `ts_norm(s.started_at)` for the `from`/`to` filters, so the inconsistency is inside one statement.
- `infrastructure/sqlite/sql/list_sessions.sql:72` — same, in the second query in the file.
- `infrastructure/sqlite/payload_rehearsal.go:685` — `SELECT ... FROM payload_rehearsal_runs WHERE state='completed' ORDER BY started_at DESC LIMIT 1`.
- `infrastructure/sqlite/payload_rehearsal.go:917` — `SELECT ... FROM payload_rehearsal_runs ORDER BY started_at DESC LIMIT 1`.

`payload_rehearsal_runs.started_at` is also written as RFC3339Nano (`payload_rehearsal_store.go:229`), and both queries pick "the latest run" with `LIMIT 1`, so a sub-second timestamp can select the wrong run.

### Reproduction (list_sessions)

Two sessions, the second genuinely later:

| session_id | started_at |
|---|---|
| `sess-a` | `2026-01-01T00:00:00Z` |
| `sess-b` | `2026-01-01T00:00:00.5Z` |

`traceary list sessions --limit 1` returns `sess-a`. The newer session is dropped from the page.

### Fix

`ORDER BY ts_norm(s.started_at) DESC, s.session_id DESC` (adding the id tie-break for a total order, matching the convention used in `search_projection_session_query.go`). Same treatment for the two `payload_rehearsal_runs` queries, with a `run_id` tie-break.

Note that `sessions` has no persisted `_norm` column the way `events` does (migration 031), so this is a function call in the ORDER BY rather than an indexed column. If the ordering cost matters for large stores, a persisted `started_at_norm` column with a maintaining trigger is the alternative — but that is a schema change and should be measured before it is chosen.

### Provenance

Found by gpt-5.6-sol during round-2 review of #1726 (#1717), confirmed by reading the SQL and the write-side timestamp formats. Out of scope for #1726, which only touched the search read path.

Related: #1185, #1717.

duck8823:
`started_at` is written as RFC3339Nano, which is variable width, and `'.'` (0x2E) sorts below `'Z'` (0x5A). A lexical `ORDER BY` therefore puts the *later* timestamp last — the #1185 hazard, still reachable in four places.

| site | consequence |
|---|---|
| `sql/list_sessions.sql:18` (inner CTE, paged) | `list sessions --limit 1` returns the older session and drops the newer one from the page |
| `sql/list_sessions.sql:78` (outer ordering) | the returned page is ordered wrongly even when it contains the right rows |
| `payload_rehearsal.go:685` | "latest completed run" selects the wrong run to scrub against |
| `payload_rehearsal.go:917` | "latest run" selects the wrong run to roll back against |

Lines 16-17 of the same statement already routed the `from`/`to` filters through `ts_norm()`, so the inconsistency sat inside one query.

Each site gains an id tie-break (`session_id` / `run_id`) so the order is total, matching the convention in `search_projection_session_query.go`.

No persisted `started_at_norm` column: `sessions` has none (unlike `events`, migration 031), and adding one is a schema change that should be measured before it is chosen. A function call in `ORDER BY` is the accepted cost here.

## Verification

Red-before-green, verified independently of the implementer by reverting each site in turn:

- revert **both** `list_sessions.sql` sites → `limit 1` returns `sess-a` where `sess-b` (`…00:00:00.5Z`) is genuinely later:
  ```
  session IDs mismatch (-want +got):
  - "sess-b"
  + "sess-a"
  ```
- revert **only** the outer `ORDER BY` → the `limit 2` case fails alone, with the page returned in inverted order. So the two sites are covered independently rather than by one test that happens to pass through both.

`go build ./...`, `go test ./infrastructure/sqlite/`, `go tool golangci-lint run` (0 issues) all clean on the final tree.

Implementation: gpt-5.6-luna. Review and verification: Claude Opus 5.

Closes #1729

### DIFF
--- a/infrastructure/sqlite/list_sessions_test.go
+++ b/infrastructure/sqlite/list_sessions_test.go
@@ -456,6 +456,50 @@ func TestDatasource_ListSummaries(t *testing.T) {
 
 }
 
+func TestDatasource_ListSummariesOrdersSubsecondStartedAtAfterWholeSecond(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name  string
+		limit int
+		want  []string
+	}{
+		{name: "limited listing selects genuinely latest session", limit: 1, want: []string{"sess-b"}},
+		{name: "listing orders all selected sessions by normalized timestamp", limit: 2, want: []string{"sess-b", "sess-a"}},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			fixture := newListSessionsFixture(t, filepath.Join(t.TempDir(), "traceary.db"), listSessionsTestMigrations())
+			ctx := context.Background()
+			if err := fixture.storeManager.Initialize(ctx); err != nil {
+				t.Fatalf("Initialize() error = %v", err)
+			}
+			for _, session := range []struct {
+				id      string
+				started time.Time
+			}{
+				{id: "sess-a", started: time.Date(2026, 1, 1, 0, 0, 0, 0, time.UTC)},
+				{id: "sess-b", started: time.Date(2026, 1, 1, 0, 0, 0, 500_000_000, time.UTC)},
+			} {
+				saveTestSession(ctx, t, fixture.sessionDS, session.id, session.started, types.None[time.Time](), "codex", "workspace")
+			}
+
+			got, err := fixture.sessionDS.ListSummaries(ctx, tt.limit, 0, types.SessionID(""), types.Workspace(""), types.Client(""), types.Agent(""), "", false, types.None[time.Time](), types.None[time.Time]())
+			if err != nil {
+				t.Fatalf("ListSummaries() error = %v", err)
+			}
+			gotIDs := make([]string, 0, len(got))
+			for _, summary := range got {
+				gotIDs = append(gotIDs, summary.SessionID().String())
+			}
+			if diff := cmp.Diff(tt.want, gotIDs); diff != "" {
+				t.Fatalf("session IDs mismatch (-want +got):\n%s", diff)
+			}
+		})
+	}
+}
+
 func TestDatasource_LineageOfCycleDetection(t *testing.T) {
 	t.Parallel()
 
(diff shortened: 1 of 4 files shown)


duck8823:
## Review record

**gpt-5.6-sol** (read-only, full diff): **no findings**.

**Claude Opus 5** (independent verification, not taking the implementer's word):

- Red-before-green reproduced by reverting each site in turn. Reverting **both** `list_sessions.sql` sites makes the `limit 1` case return `sess-a` where `sess-b` (`…00:00:00.5Z`) is genuinely later:
  ```
  session IDs mismatch (-want +got):
  - "sess-b"
  + "sess-a"
  ```
  Reverting **only** the outer `ORDER BY` makes the `limit 2` case fail alone, with the page inverted. So the inner paging CTE and the outer ordering are covered independently, not by one test that happens to pass through both.
- `ts_norm` is registered on the modernc driver in `database.go`'s `init()`, so it is present on every connection — no per-connection registration gap.
- `go build ./...`, `go test ./infrastructure/sqlite/`, `go tool golangci-lint run` (0 issues) clean on the final tree. All CI checks pass.

## Out of scope, filed as follow-up

`payload_backfill_runs.started_at` is written with `time.RFC3339Nano` (`payload_backfill.go:467`) and read back through a bare lexical `ORDER BY started_at DESC` at `payload_backfill.go:930` and `:943` — the same #1185 hazard class, in a table #1729 did not enumerate. Not pulled into this PR: 1 issue = 1 PR. Filed separately.

Issue #1810, opened 2026-08-11
v0.34.0: rename --index-family-bytes before the narrowed scope ships


duck8823:
Follows the decision on #1796. Time-boxed: everything named here is **unreleased**, so the rename is free now and permanent the moment v0.34 ships.

## Why the names are now wrong

`DefaultSearchProjectionIndexFamilyBytes` documents itself as covering "the whole bounded search index family" — documents, trigram index, session tier and literal fingerprints. #1796's decision narrows the ceiling to the **evictable recent tier only**: `search_projection_recent_documents`, its indexes, and `search_projection_recent_fts_*`. The session tier and `literal_search_fingerprints` are corpus-proportional by design and sit outside the ceiling.

Measured at 36.1% of the source walk, the tiers now excluded were 1,235,628,032 B — 80.5% of the 1464 MiB the flag claims to bound. An operator setting `--index-family-bytes 1G` today would reasonably expect the family to stay under 1 GiB; after #1796 it bounds a little over a third of it. The name is the whole contract for a flag, and this one will state the wrong contract.

## Surface to rename

| kind | current | scope |
|---|---|---|
| CLI flag | `--index-family-bytes` | `presentation/cli/store_search_projection.go:102` |
| constant | `apptypes.DefaultSearchProjectionIndexFamilyBytes` | `application/types/search_projection.go:26` |
| budget field | `SearchProjectionBudget.IndexFamilyBytes` | same file, `:49`; appears in `ConfigHash` |
| JSON status fields | `index_family_byte_limit`, `index_family_within_budget` | `:295`, `:306` |
| schema columns | `index_family_byte_limit`, `index_family_within_budget` | migration `000055` |
| docs | `docs/search-projection-rebuild.md` and `.ja.md` | both |

Suggested direction, not decided: `--recent-index-bytes` / `RecentIndexBytes` / `recent_index_byte_limit`. Whatever is chosen has to distinguish it from the pre-existing `recent_byte_limit` column (which holds a *source text* ceiling, and which migration 55's comment already flags as confusable) and from `--recent-age`.

## Constraints

[...]

Issue #1622, opened 2026-08-01
v0.34.0-12: add historical search candidates and continuation


duck8823:
## Objective

Add bounded historical literal-search candidates, canonical verification, and resumable tier-aware continuation without false complete negatives.

## Acceptance criteria

- [ ] The characterized filterable query class has no candidate false negatives across Unicode, case, paths, IDs, and error fragments.
- [ ] Candidate matches are verified against codec-decoded canonical plaintext.
- [ ] Short/unsafe/unfilterable queries use bounded verification and explicit partial coverage plus continuation.
- [ ] CLI and MCP expose tier provenance, coverage, partial reasons, and stable continuation.
- [ ] `--deep`, if present, changes budgets only.
- [ ] Candidate structures respect measured byte budgets and rebuild deterministically.
- [ ] Tests, vet, and lint pass.

## Delivery routing

Kimi K3 max implementation; Claude Fable 5 search-semantics review; GPT-5.6 Sol edge-case verification; Codex orchestration.

## Risk / dependencies

High. Depends on #1611 and #1609. Parent: #1606.

Issue #1727, opened 2026-08-08
v0.34.0-53: give the MCP search tool the session tier the CLI got


duck8823:
## Problem

#1717 makes the bounded search projection authoritative for `traceary search`. History older than the recent window no longer returns events; it returns **sessions**, resolved from summaries and keywords, shown as a labelled SESSIONS group.

The CLI got that group. The MCP `search` tool did not.

`presentation/mcpserver/server.go` resolves search through `s.eventBounded.Search` / `s.event.Search` / `s.eventMetadata.Search` and never calls `ProjectionSessionSearchQuery.SearchSessionHits`. So through MCP:

- a query matching only old history returns an empty event list
- the agent cannot distinguish that from "nothing was ever recorded"
- the session trail that the projection exists to preserve is unreachable

This is the wrong way round. MCP is the agent-facing surface, and an agent searching for something it did months ago is precisely the case the session tier was built for. The human at the CLI gets the trail; the agent gets silence.

## Why it is not part of #1717

The MCP tool response shape is a frozen contract (v0.10 MCP surface reduction, 24 → 8 tools, JSON contract freeze). Adding a `sessions` array to the `search` response is a shape change on that surface and needs its own decision about the compatibility window, exactly as `search --json` did in #1717 — where the array was deliberately preserved for v0.34 and the envelope deferred to v0.35.

Widening #1717 to cover it would have bundled two contract decisions into one PR.

## Options

1. **Add `sessions` to the MCP `search` response** on the same schedule as the CLI `--json` envelope (v0.35), so both agent-facing JSON surfaces change shape in the same release and the docs tell one story.
2. **Return session hits as synthetic entries in the existing event list**, discriminated by a field. Cheaper, but it makes an agent parse two row kinds out of one array, and #1717 rejected exactly this for the CLI.
3. **Leave MCP event-only** and rely on the agent falling back to `session` tools when search comes up empty. Only defensible if the agent can actually tell "empty because old" from "empty because absent" — today it cannot, so this option requires at minimum a signal that says so.

Option 1 is the default. Option 3 without a signal is the current behaviour and is the thing being reported.

## Done when

- A decision is recorded here with its reason
- If the response shape changes: the compatibility window follows `docs/cli-stability.md`, MCP schema snapshots are updated, and the contract test suite pins the new shape
- The behaviour on a store whose projection has never completed is stated — MCP must not be silently empty either

No new MCP tool. v0.34 reduces surface; this is a field on an existing tool or nothing.

[...]

PR #1816, opened 2026-08-11
refactor: remove the never-released tiered search preview

### LINKED ISSUE #1725: v0.34.0-52: resolve the two competing tiering mechanisms on traceary search
## Problem

`traceary search` will carry **two independent tiering mechanisms** once #1717 and #1680 land.

The one being built is the bounded projection: a recent trigram tier plus a session-summary tier, selected automatically, no flags.

The one already there is `--tiered-preview`, added by #1622 (`presentation/cli/search.go:118-120`, `application/queryservice/literal_search_service.go`, `infrastructure/sqlite/literal_search_query.go`):

```go
searchCmd.Flags().BoolVar(&tieredPreview, "tiered-preview", false, "use bounded historical literal preview with explicit coverage")
searchCmd.Flags().BoolVar(&deep, "deep", false, "increase tiered preview resource budgets without changing search semantics")
searchCmd.Flags().StringVar(&continuation, "continuation", "", "resume an incomplete tiered preview")
```

Three public flags with:

- **no documentation** — `rg -l "tiered-preview" docs/` returns nothing
- **no localisation** — the usage strings are raw English while every other flag on the same command uses `Localize(en, ja)`
- **a hard dependency on `--json`** — `--tiered-preview` without `--json` returns `--tiered-preview requires --json`
- **its own result shape** — `{events, tier, coverage, partial_reason, continuation}`, unrelated to the shape the rest of `search` returns

They are visible in `traceary search --help` today.

## Why this is a reduction issue, not a cleanup nicety

v0.34's stated direction is that surface goes down. Shipping a release where one command exposes two different answers to "what happens when the query reaches past the indexed window" is the opposite, and the undocumented one is the one users will hit by accident from `--help`.

The two mechanisms also answer different questions:

| | `--tiered-preview` | #1717 projection tiering |
|---|---|---|
| older-history behaviour | literal scan with an explicit budget and a resumable continuation | session rows from summaries and keywords |
| coverage honesty | explicit `coverage` / `partial_reason` fields | complete within the window, sessions beyond it |
| output | JSON only, own shape | text and JSON, main shape |
| selection | user passes a flag | automatic |

Deciding which one survives is the point of this issue.

## Options

1. **Remove the three flags** once #1680 has cut over, and let the projection tiering be the only answer. The literal-search machinery underneath (`literal_search_query.go`, `literal_search_progress.go`, migration 039 fingerprints) may still be worth keeping for internal use by the projection path — removing the flags is not the same as deleting the code.
2. **Keep them, documented and localised**, and state in the docs exactly when a user should reach for the literal preview instead of the default. This is only defensible if there is a question the projection tiering genuinely cannot answer.

Option 1 is the default unless the evidence for option 2 is produced.

## Done when

- A decision is recorded here with its reason
- If removed: the flags are gone, the deprecation notice follows `docs/cli-stability.md`, and any capability worth keeping is reachable without a flag
- If kept: they are documented in `docs/cli/README.md` and `docs/cli/README.ja.md`, localised like every other flag on the command, and the docs say when to use which

Blocked on #1680 — the comparison is not decidable until the projection tiering is actually the read path.

duck8823:
`traceary search` carried two independent answers to "what happens when the query reaches past the indexed window": the bounded projection tiering that #1717/#1680 made authoritative, and the `--tiered-preview` literal preview added by #1622. The second was undocumented, unlocalised, JSON-only, had its own result shape, and was visible in `--help`. Decision and reasoning: #1725 (issuecomment-5250476681).

## No deprecation window

Both entry points came from `10ab81f7` (2026-08-02). `git tag --contains 10ab81f7` is empty and the latest tag is `v0.33.1`, so no released binary ever exposed them. `docs/cli-stability.md`'s notice-in-N / removal-in-N+1 flow protects surface external callers could have pinned against; there is none here.

## Removed

| surface | removed |
|---|---|
| CLI | `--tiered-preview`, `--deep`, `--continuation`, the `criteriaTo` special case, and the JSON-only branch |
| MCP | `search` input fields `tiered_preview` and `deep` |
| wiring | `WithTieredEventSearch` on both surfaces, `main.go` construction |
| unreachable | `literal_search_service.go`, `literal_search_progress.go`, `literal_search_query.go`, `TieredEventSearchQuery`, and the unreferenced parts of `types/literal_search.go` |

MCP `search`'s `continuation` **stays** — it is the ordinary event-pagination cursor, shared with `get_context`, and was only coincidentally forwarded into the tiered request. The `tool_registry.golden.json` diff is exactly two input properties; no output schema changed.

## Deliberately kept

`literal_search_fingerprints` and `literal_search_projection_state` (migration `000039`) are **not** tiered-preview-only. They are written by the projection rebuild (`search_projection_rebuild.go:887`), read by the projection's own authority path (`event_search_authority.go:226-229`), and accounted for in `select_search_projection_logical_non_recent_bytes.sql`, `measure_search_projection_family_split.sql` and `search_projection_catchup.go:72`. The tables, the migration, and the prepared-migration catalog entry are untouched, so migration 039's bytes are unaffected by this PR — relevant to anyone reading #1810 or #1620.

The implementer also reported `CharacterizeLiteralQuery` and `DeepLiteralSearchBudget` surviving as projection-authority dependencies rather than preview-only code. Confirmed against `event_search_authority.go`.

## Verification

Re-run independently of the implementer on the final tree:

- `go build ./...` clean
- `go test ./presentation/cli/ ./presentation/mcpserver/ ./application/types/ ./infrastructure/sqlite/` — all pass
- the removal is pinned, not just deleted: `TestRootCLI_SearchRejectsRemovedTieredPreviewFlags` asserts each of the three flags now fails with `unknown flag`, so they cannot be silently re-added
- `rg tiered-preview docs/` returns nothing, so there was no documentation to remove; the change is recorded in `CHANGELOG.md` / `CHANGELOG.ja.md`

Net: 61 insertions, 1600 deletions.

Implementation: gpt-5.6-luna. Decision, review and verification: Claude Opus 5.

Closes #1725


duck8823:
## Review record

**gpt-5.6-sol** (read-only, full tree, asked to look for over-deletion, under-deletion, projection-table damage, MCP schema drift, `continuation` asymmetry, and vacuous tests): **one HIGH**, and it was right.

> `presentation/mcpserver/output.go:52` — `eventsOutput.Tier` remains. Nothing sets it any more, yet the shared output schema keeps `search`, `list_events` and `get_context` all declaring a tiered-provenance field none of them can return. The removal is incomplete and schema-driven clients are promised behaviour that does not exist.

**Accepted and fixed in `167482a0`.** My earlier position on #1727 was to leave the declaration alone and remove it in v0.35, on the reasoning that dropping a property from three `additionalProperties: false` schemas is breaking. That reasoning does not apply here: `git log -S` puts the `Tier` field in the same commit as the input fields — `10ab81f7` (2026-08-02) — and `git tag --contains 10ab81f7` is empty against a latest tag of `v0.33.1`. It was never released. And any client that could read `tier` necessarily passed `tiered_preview`, which this branch removes anyway. So the removal is free and belongs here, which also removes the vestigial-field decision from #1727's scope.

Regenerated with the repo's generator, not by hand:

```
go test ./presentation/mcpserver -run 'TestServer_Tool(RegistrySnapshot|AdvertisementBudget)$' -update -update-tool-schema-budget
```

Sol reported nothing on the other five areas.

## Independent verification (Claude Opus 5)

- `go build ./...` clean; `go test ./presentation/cli/ ./presentation/mcpserver/ ./application/types/ ./infrastructure/sqlite/` all pass; `go tool golangci-lint run` reports 0 issues — re-run on the final tree, after the `Tier` removal, not taken from the implementer's report.
- The removal is **pinned**, not merely deleted: `TestRootCLI_SearchRejectsRemovedTieredPreviewFlags` drives each of `--tiered-preview`, `--deep`, `--continuation` and requires `unknown flag` in the error, so none can be silently re-added.
- `rg -n "Tier" presentation/mcpserver/*.go` returns nothing, confirming no writer survived the field.
- `literal_search_fingerprints` and `literal_search_projection_state` are untouched: still written by `search_projection_rebuild.go:887`, still read by `event_search_authority.go:226-229`, still counted in `select_search_projection_logical_non_recent_bytes.sql`, `measure_search_projection_family_split.sql` and `search_projection_catchup.go:72`. Migration 039 and its prepared-migration catalog entry are unchanged, so this PR moves none of #1620's bytes.
- MCP golden diff is now two input properties plus the one output property, and `list_events` / `get_context` change only through the shared `tier` removal — verified by reading the regenerated `tool_registry.golden.json`, not inferred.

Ready to merge.

Issue #1808, opened 2026-08-11
v0.34.0: session-keyword and fingerprint tables store their key twice via implicit autoindexes


duck8823:
Split out of #1796, where it was measured but deliberately left unimplemented. Now that #1796's decision documents the session tier and the fingerprint tier as **corpus-proportional** rather than bounded by the index-family ceiling, their size stops being a rounding error inside a cap and becomes a number the release publishes. Halving it is worth doing on its own terms.

## Measured

`dbstat` on the rehearsal copy at checkpoint 147,584 / 408,970 (36.1% of the source walk):

| structure | bytes |
|---|---|
| `search_projection_session_keywords` table | 414,154,752 |
| `search_projection_session_keywords` autoindex | **412,717,056** |
| `literal_search_fingerprints` table | 233,410,560 |
| `literal_search_fingerprints` autoindex | **112,029,696** |
| `idx_literal_search_fingerprint_candidate` | **116,424,704** |

Cardinality at the same point: 2,228,959 keyword rows over 16,059 sessions, 1,202,818 fingerprint rows.

Both tables are rowid tables carrying a UNIQUE or PRIMARY KEY constraint, so SQLite materialises an implicit index that duplicates the entire key. For `search_projection_session_keywords` the autoindex is **as large as the table it indexes** — the key is essentially the whole row, so the row is stored twice.

## Three changes

1. `WITHOUT ROWID` on `search_projection_session_keywords`. The key (`generation_id`, `session_id`, `keyword`) is the row apart from the count, so the b-tree becomes the table and the duplication disappears.
2. `WITHOUT ROWID` on `literal_search_fingerprints`, same argument.
3. `idx_literal_search_fingerprint_candidate` overlaps the same key material as the autoindex it sits beside. Confirm against the query plans that actually read it (`event_search_authority.go`) before dropping it — if it is a genuine covering index for the candidate lookup, keep it and drop the redundant half instead.

Roughly 640 MB at 36% of the walk, extrapolating to ~1.7 GB at completion on this corpus. Expressed rate-independently: the keyword tier costs ~26 KiB/session of which about half is duplication.

## Constraints

- These are `CREATE TABLE` changes, so they need a migration that rebuilds the tables. Both are **derived** projection tables — a generation can be discarded and rebuilt — so the migration may drop and recreate rather than copy, provided it also clears the generation state so the next store open restarts the build. Confirm that against `AbandonSearchProjection`.
[...]

Issue #1807, opened 2026-08-11
v0.37-12: recent-cutoff derivation times out precisely on the stores that need it


duck8823:
Split out of #1796. With the owner's decision on that issue, this stopped being a correctness blocker and became a build-cost problem, so it needs its own issue rather than being carried by the ceiling work.

## What it does now

`deriveSearchProjectionRecentCutoff` walks the corpus newest-first to find a byte cutoff for the recent tier, under `searchProjectionCutoffTimeout = 2 * time.Second` (`infrastructure/sqlite/search_projection_rebuild.go:83`). On the rehearsal copy of the live store it timed out, persisting an **empty** `recent_cutoff_norm` and `capacity_evidence_status = unavailable` with reason `amplification sample below minimum; recent cutoff: context deadline exceeded`.

## Why the failure mode is backwards

The derivation exists to keep a large store from admitting more recent text than the ceiling allows. It is a wall-clock walk over that same store. So it succeeds on small stores, which do not need it, and times out on large stores, which do — the guard is absent exactly where it is load-bearing.

With #1796's decision (the ceiling scopes to the evictable recent tier, and eviction interleaves with the source walk) the consequence is no longer unbounded growth. It is cost: without a byte cutoff the planner falls back to age-only retention, so a large store admits its whole 30-day window and then evicts most of it back out. The build pays for admit-then-evict churn on rows that were never going to be retained — on this corpus that is the majority of the recent window.

## What to decide

- Whether 2 seconds is a deadline at all, or whether the derivation should be **bounded by work** (N pages / N rows sampled) so that it produces a usable answer on every store instead of a complete answer on small ones. A sampled cutoff that is 20% wrong is worth far more than no cutoff.
- Whether a timed-out derivation should be **retried with a wider budget** on the next batch rather than persisted as a permanent `unavailable` for the generation.
- Whether `status` should distinguish "no cutoff because the store is small" from "no cutoff because the derivation gave up" — today both read as `unavailable`.

## Not in scope

The ceiling's enforcement (#1796) and the 2.16x fallback amplification constant (#1751, measured at 3.37x). This issue is only about making the derivation terminate with an answer.

Parent: #1606

Issue #1751, opened 2026-08-09
v0.37-5: make the search projection's capacity derivation trustworthy enough to bound the store


duck8823:
Follow-up to #1679 / #1750. Found by an adversarial review of that PR and deliberately deferred there — each item needs state-machine work that would have ballooned an already-large change.

#1679 makes the operator's budget mean physical index-family bytes and converts it to a source-text ceiling using a **measured** trigram amplification. The conversion is only as good as that measurement, and today it is not good enough to call the resulting bound trustworthy. Five distinct reasons, all in `infrastructure/sqlite/search_projection_rebuild.go`:

**1. The re-derivation never retries.** The ceiling is derived twice: an estimate at `Start` from the outgoing generation, then a re-derivation from the new generation's own sample at the source→eviction transition. The transition is the *only* trigger. A crash between the phase commit and `recordSearchProjectionCapacityRederivation` leaves the generation running to completion on the Start-time estimate, and nothing ever notices. If the two generations' amplification differs materially, so does the resulting family size.

**2. The measurement is not a consistent snapshot.** `measureSearchProjectionFamilySplit` (dbstat) and the `SUM(decoded_bytes)` sample are separate statements outside a transaction. A concurrent eviction between them yields "physical before the delete / logical after it" and records an amplification far above reality, which then shrinks the ceiling for the rest of the generation. Traceary's writers are shell hooks, so concurrency here is routine, not hypothetical.

**3. Amplification blends both generations.** Numerator and denominator both cover every generation present in the recent tier, which is required for them to be consistent with each other but means the figure is a mixture of the outgoing and incoming layouts. If the old generation indexes at 4x and the new one at 2x, the recorded 3x needlessly shortens the window; reversed, it lets the family exceed the budget. FTS5 deleted postings that survive until a merge (#1740) inflate the numerator only, which biases this toward a ratchet: each rebuild measures higher than the truth and derives a smaller ceiling than the last.

**4. Nothing corrects an over-budget completion.** On completion the family is re-measured and `index_family_within_budget` records 0 when it landed over. That is all that happens: the state goes `complete`, the next `CatchUp` returns `already_complete`, and the store stays over budget until an operator notices the status field and runs an explicit `start` with a smaller `--index-family-bytes`. For a gate that is supposed to hold unattended (#1620), a detector with no actuator is not enough.

[...]

PR #1817, opened 2026-08-11
fix: interleave eviction with the projection source walk

### LINKED ISSUE #1796: v0.34.0: a search projection rebuild peaks at ~7x its index-family budget because eviction only runs after the whole source walk

duck8823:
## What

Eviction now runs **interleaved with the source walk** instead of waiting for it to finish, so the derived source-text ceiling bounds the rebuild **peak** rather than the post-hoc total.

Before this, `PlanProjectionBatch` set `NextPhase = "eviction"` only under `SourceDone`. Everything the source phase admitted stayed resident until the last row was walked. Measured on a copy of the live store (534k events), at checkpoint 147,584 of high-water 408,970 — **36.1% of the walk** — the index family stood at **3.74 GiB against a 1.43 GiB ceiling**, with 63.9% still to go. The walk order makes it worse than it sounds: `search_projection_source_sequence` is filled `ORDER BY e.id` and `events.id` is a content hash, so the recent window is admitted at a uniform rate across the whole walk rather than in a final burst.

## How

- **`recent_source_bytes`** (migration `000058`) — a persisted per-generation counter of source text currently resident in the recent tier. `Start` resets it to 0; `ApplyBatch` increments on admission and decrements by exactly what it deleted, in the same transaction. The bound therefore survives a restart rather than being re-derived from a scan.
- **`selectProjectionEviction`** (new file, so the 800-line `search_projection_rebuild.go` gains no new responsibility) — selects the generation's oldest recent documents, and computes the **`Expired` flag in SQL**, against the same `projectionCutoff(now - RecentAge)` string the snapshot gate uses.
- **`SelectSnapshot`** routes the source phase into eviction when `recent_source_bytes > ceiling` **or** any row has aged past the cutoff.
- **`PlanProjectionBatch`** walks the candidates with a running byte total and stops as soon as the tier is under the ceiling and nothing is expired.

Migration `000058` also restarts any generation an older binary left mid-rebuild: the counter cannot be reconstructed for work already done, so the generation is abandoned in the lifecycle table, `literal_search_projection_state` is marked stale, and the state row is reset to idle so the next store open starts a clean replacement. It is `MigrationConstantInPlace` — the statements are constant, not data-dependent.

## Why the expiry decision is computed in SQL

The first implementation re-derived it in the planner with `time.Parse(time.RFC3339Nano, c.CreatedAtNorm)`, while the snapshot gate decided it with a SQL string comparison. `created_at_norm` can be `''` (migration 031's backfill left NULLs, and the source-row query applies `COALESCE(…,'')`), so the parse fails → not expired → under ceiling → the branch selects nothing → **no cleanup, no checkpoint move, no phase change, repeated identically forever**. That is #1793 and #1794's failure class — a permanent stall — reintroduced by the fix for a resource problem. One decision, one place, one representation.

## What this does *not* bound

The sum across generations. `Start` keeps the previous generation fully resident — that is what keeps search answerable during a rebuild — and it is reclaimed only in the terminal cleanup phase. Free space must cover the old family plus the new ceiling. Now documented rather than implied.

## Docs

`docs/search-projection-rebuild.md` / `.ja.md` gain **what the budget actually bounds**: only `search_projection_recent_documents` + its indexes + `search_projection_recent_fts_*` are evictable and held to a ceiling. The session tier and `literal_search_fingerprints` are corpus-proportional by design — they enter the derivation as the non-recent reserve, so at a fixed budget the recent window *shortens* as the corpus grows (on the reference corpus they were already 80.5% of a 1464 MiB budget at 36% of one walk). Also drops the "residual of the 4 GiB store gate" justification for the default, which no longer describes how the figure is set.

Per [the 2026-08-11 decision](https://github.com/duck8823/traceary/issues/1796#issuecomment-5248947989), the rest of #1796's original findings are split out: the cutoff derivation's timeout (#1807), the duplicated autoindexes (#1808), the flag rename the narrowed scope forces (#1810), the fallback amplification constant (#1751).

## Tests

| behaviour | test |
|---|---|
| peak stays bounded after **every** batch of a full walk, and the counter equals the measured residency each time | `TestSearchProjectionRecentResidencyAndCounterStayBoundedEachBatch` |
| eviction phase, tier under ceiling, nothing expired ⇒ **zero rows deleted**, phase advances | `TestSearchProjectionEvictionUnderCeilingRetainsRows` |
| an unparseable `created_at_norm` does not stall admission | `TestPlanProjectionBatchAdmitsAfterUnparseableCandidateIsNotExpired` |
| `Rows=1` (reachable since #1803's adaptive shrink) still makes progress | `TestPlanProjectionBatchEvictsOneRowWithoutCheckpoint` |
| the eviction query uses the index, no temp b-tree | `TestSelectProjectionEvictionUsesEvictionIndexWithoutTempBTree` |
| migration 58 abandons a mid-rebuild generation and the next open starts a replacement | `TestSearchProjectionMigration58AbandonedRebuildAutoStartsReplacement` |
| row/byte budget boundaries | `TestPlanProjectionBatchCapacityBoundaries` |

`go build ./...`, `go test -count=1 ./...` and `go tool golangci-lint run` (0 issues) all clean on the final tree.

Closes #1796


duck8823:
## Review record

**gpt-5.6-sol** (read-only, full tree, asked about stall reachability, counter correctness, production effectiveness of the ceiling, migration 058, the oversize dead end, query cost, and test integrity): **one CRITICAL**, which I checked and **rejected on evidence**. Nothing on the other six areas.

> `000058_bound_recent_projection_source_bytes.sql:4` — `recent_source_bytes` is maintained only by the application, so rolling back to a supported older binary, admitting source rows, and returning to the new binary leaves the counter under-counting. Eviction then under-fires and the completed recent tier exceeds its ceiling.

The mechanism is right; the premise that such a binary exists is not. The finding requires a binary that **both** rebuilds the projection **and** lacks migration 058. There is none:

```
$ git ls-tree -r v0.33.1 --name-only | grep -c search_projection
0
$ git ls-tree v0.33.1 schema/sqlite/migrations/ | tail -1
000034_create_event_metadata_projection.sql
```

The entire bounded search projection is new in v0.34 — tables, state row, rebuild machinery, all of migrations 035+. A v0.33.1 binary opening a v0.34 store tolerates the higher migration versions and then never touches `search_projection_recent_documents`, because it has no code that knows the table. The only binaries inside the described window are unreleased intra-v0.34 development builds, and migration `000058` is precisely the mitigation for those: it abandons any generation left mid-rebuild and resets the state row so the next open starts a clean replacement.

The **underlying** observation — a cache with no independent verifier is one call site away from silent drift, low (peak unbounded again) or high (over-eviction), neither detectable from outside — is correct and is filed as **#1819** for v0.35, with the three options costed.

## Independent verification (Claude Opus 5)

Re-checked on the final tree rather than taken from the implementer's report or from sol.

**Does it bound the peak in production, or only in the tests?** The tests set `recent_source_ceiling_bytes` directly, and source-phase eviction is gated on `RecentSourceCeilingBytes > 0` (`search_projection_rebuild.go:594`), so a `Start`-time ceiling of 0 would make the change ineffective exactly where it matters. Traced `deriveSearchProjectionCapacity` (`:1224-1293`): on a store with a populated recent tier the ceiling is `(budget − nonRecent) / measured amplification`; on a fresh store the amplification falls back to the constant and the ceiling is still positive. It reaches 0 only when the non-recent reserve alone meets the budget (`:1271-1287`) — the documented empty-recent-tier case, which `search_projection_plan.go:48` handles by admitting nothing rather than by silently skipping eviction. So the gate is closed only when the correct answer is "build no recent tier at all".

**Counter maintenance across every path.** Only `Class == "eviction"` decrements (`:923`), so the terminal cleanup deleting other generations' rows correctly leaves it alone. `cleanup_scope='all'` does delete current-generation rows without decrementing, but that generation ends `drifted` and `Start` resets the counter to 0 (`:227`), so the stale value is never read. Source-phase replay is fenced by `(?=0 OR recent_source_bytes=?)` (`:977`); eviction-phase replay has no checkpoint to fence against but is caught by the delete-must-affect-exactly-one-row check (`:920-922`).

**Stall reachability**, the class that produced #1793 and #1794 and that an earlier revision of this branch reintroduced: the eviction phase now always `return`s and sets `NextPhase = "cleanup"` when it selects nothing (`search_projection_plan.go:38-43`), so it can neither fall through to `PlanProjectionRetention` — which has no ceiling or expiry logic and would delete rows a within-budget tier must retain — nor return a plan that advances nothing. The source phase falls through to admission when eviction selects nothing, so an unparseable `created_at_norm` cannot park the walk.

**Test integrity.** `TestSearchProjectionRecentResidencyAndCounterStayBoundedEachBatch` runs `Rows=1` against a corpus several times the ceiling and asserts residency **and** counter equality after **every** batch, then requires completion within 100 batches — it cannot pass by completing in one. `TestSearchProjectionEvictionUnderCeilingRetainsRows` asserts the row count in `search_projection_recent_documents` is unchanged, not merely that the plan looked empty; it was written red against the previous revision and failed with `recent document count=0, want 1`.

**Build.** `go build ./...`, `go test -count=1 ./...` (all packages ok) and `go tool golangci-lint run` (0 issues) clean on the final tree.

Ready to merge.

PR #1818, opened 2026-08-11
fix: announce when --kind empties the session search group

### LINKED ISSUE #1728: v0.34.0-54: reconcile session-tier filter semantics with the event tier
## Problem

#1717 added the session tier to `traceary search`. Two of its filter semantics differ from the event tier in ways that are defensible per-filter but produce results the user did not ask for. Raised by a gpt-5.6-sol review of #1726 and confirmed against the code; deliberately not fixed inside #1717 because both are design decisions, not slips.

### 1. `--kind` silently removes the session group

`infrastructure/sqlite/search_projection_session_query.go` fails closed when `criteria.Kind()` is set: session summaries and keywords carry no per-kind breakdown, so filtering them by kind is impossible and returning them unfiltered would be wrong.

The consequence is that `traceary search --kind note <query>` over old history returns nothing at all — the event tier cannot reach evicted events and the session tier is skipped. Without `--kind` the same query returns the session. The user sees an empty result with no indication that narrowing by kind is what emptied it.

Options:

- store a per-session kind set (or per-kind counts) in the projection so the filter can be applied honestly — costs projection bytes, which v0.34 is trying to reduce
- keep failing closed but say so: when `--kind` suppressed a non-empty session group, tell the user on stderr, the same way `--json` announces omitted sessions
- decide that `--kind` is an event-tier-only filter and document it

### 2. `--from` / `--to` and `--failures-only` are session-granular, not event-granular

`appendSessionSearchFilters` applies the time range to `sessions.started_at` and `--failures-only` to `search_projection_command_aggregates.failure_count > 0` for the whole session.

Two consequences:

- a session that started before `--from` but whose matching activity falls inside the range is dropped
- `--from ... --to ... --failures-only <query>` can return a session where no single event satisfies all three: the query matched a successful note, and an unrelated command in the same session failed

For the event tier every filter narrows down to one row, so "matched" means one event satisfied everything. For the session tier there is no such row — a session hit means "somewhere in this trail". That may be the right meaning, but it is currently unstated, and the two tiers are presented side by side in the same output.

Options:

- state the semantics in `docs/cli/README.md` and accept them ("session rows mean the trail contains a match; filters apply to the session, not to a single event")
- carry a coarse time range per session summary (first/last event time) so `--from` / `--to` can intersect rather than test the start instant
- suppress the session group entirely when a filter cannot be applied at session granularity, as `--kind` already does

## Done when

- A decision is recorded here for each of the two, with its reason
- Whatever is chosen is reflected in `docs/cli/README.md` and `docs/cli/README.ja.md`, since this is user-visible search behaviour
- Any case where a filter silently empties the session group is either fixed or announced — silent emptiness is what #1717 was careful to avoid for `--json`, and the same standard applies here

No new flags. If a filter cannot be honoured at session granularity, the answer is to say so, not to add a switch that turns the tier off.

duck8823:
## What

`traceary search --kind <k>` silently returned an empty SESSIONS group. Now it says so.

`SearchSessionHits` fails closed when a kind filter is set (`infrastructure/sqlite/search_projection_session_query.go:31-34`), because session summaries and keywords carry no per-kind breakdown — returning them unfiltered would be wrong. Failing closed is correct; being silent about it is not. `traceary search --kind note <query>` over history older than the recent window returned nothing at all, and the user had no way to learn that `--kind` was the reason.

## How

An empty result is indistinguishable from "no session matched", so announcing requires knowing whether there would have been hits. When — and only when — `--kind` is set, the session query runs **once more with the kind filter dropped**; if that probe finds anything, the cause goes to the warn writer, localised, in both text and `--json` mode. No other search pays for it.

The probe is derived with `EventSearchCriteria.WithoutKind()`, not rebuilt field by field through the builder. It must differ from the user's criteria in exactly one field: a filter added to `EventSearchCriteria` later would otherwise silently not reach the probe, and the notice would then claim sessions were suppressed by `--kind` when the user's own filter had excluded them. `TestRootCLI_SearchKindProbeKeepsEveryOtherFilter` drives every flag `search` accepts and asserts the two criteria differ only in kind.

### The notice reports presence, not a count

The first implementation reported *how many* sessions were suppressed. gpt-5.6-sol found that the probe reused the exclusion list built from the **kind-filtered** event page, so it did not reproduce the search it claimed to describe: it could count a session whose other-kind events would have filled the event group, and miss a session that `--limit` pushed out of it.

The exclusion list is not the thing to fix — the count is. Producing an honest number means re-running the whole kind-less search including the event page, and the number does not change what the operator does about it; the remedy is the same for one hidden session or forty. So the probe now runs with **no exclusion list**, asking exactly "does the session tier match this query when kind is not applied", and the message states presence.

The two `--json` notices are also emitted **independently** rather than as an if/else. They answer different questions — "sessions matched but `--json` cannot carry them" and "sessions matched but `--kind` hid them" — and neither may swallow the other. Each is a no-op when its own condition does not hold, so this function needs no ordering rule and no knowledge that today they cannot both fire.

## What is deliberately *not* changed

Two session-tier filter semantics differ from the event tier. Both are documented in `docs/cli/README.md` / `.ja.md` rather than "fixed", and the reasons are worth recording because the fix is tempting:

**`--from` / `--to` test `started_at`, not an interval intersection.** An intersection was designed and withdrawn. `search_projection_session_keywords` and `search_projection_session_summaries` are accumulated with `ON CONFLICT … DO UPDATE` across every event in the session (`search_projection_rebuild.go:860-883`) and carry **no per-term timestamp**. Intersecting `[started_at, ended_at)` with the requested range proves the *session* overlapped it, never that the *match* did — so it would return sessions where the term appears only outside the window while claiming to fix a false negative. `ended_at` is also unreliable (Codex and Antigravity have no host-level termination signal, and the stale GC writes the GC run time). And it would give `[from,to)` a third meaning: `session list` applies it to `started_at`, `timeline` to event `created_at`.

**`--failures` is session-granular**, resolving through `search_projection_command_aggregates.failure_count > 0` for the whole session. Unlike `--kind` it widens the group rather than emptying it, so there is nothing to announce per search — but a session can appear when the query, the time range and `--failures` are each satisfied by different activity within it. That is what a session hit means: the tier deliberately keeps no per-event row.

Rejected as out of scope: storing a per-session kind set or per-kind counts in the projection. That buys an honest filter at the cost of projection bytes, in the release whose point (#1620, #1796) is bounding them.

## Tests

| behaviour | assertion |
|---|---|
| suppression announced | old history matches with `--kind` set: stderr matches the notice **exactly** in both locales — which is what pins the absence of a count and of a spurious JSON notice — while stdout still shows the event-tier result |
| no false announcement | `--kind` set, probe finds nothing: stderr is silent |
| no probe without `--kind` | the session query runs exactly once |
| probe carries no exclusions | the real call excludes the session already shown as an event; the probe receives `nil` |
| probe fidelity | probe criteria differ from the real ones in exactly the kind field, across every search flag |

`go build ./...`, `go test -count=1 ./presentation/cli/ ./application/types/` and `go tool golangci-lint run` (0 issues) clean.

Closes #1728


duck8823:

### The one HIGH, and why it changed the design rather than the code

sol, on `presentation/cli/search.go:198,219,280`:

> probeが`--kind`付きのevent/metadata結果から作った除外リストを再利用するため、実際の「`--kind`なし」検索を再現しません。別kindのeventがあるsessionを誤って抑制件数に含める一方、`--limit`によってevent枠から外れるsessionを見逃します。後者のJSON出力ではkind警告もJSON omission警告も出ません。

Confirmed, both halves. The exclusion list is `sessionIDsFromEvents(events)` / `sessionIDsFromMetadata(metadata)` — session ids drawn from a page the `--kind` filter shaped. A search *without* `--kind` returns a different page, so it would exclude a different set. The probe was therefore answering a question nobody asked, in both directions.

The fix is not a more faithful exclusion list. **The count is what made exclusion-list fidelity matter**, and the count adds nothing the operator acts on — the remedy is the same whether one session or forty were hidden. Producing an honest count would mean re-running the whole kind-less search including the event page, to print a number whose only use is deciding how surprised to be.

So the notice became a presence signal:

- the probe runs with **no exclusion list**, asking exactly "does the session tier match this query when kind is not applied" (`search.go:276-289`);
- `searchProjectionSessions` returns `bool` instead of `int`;
- the message drops the count in both locales.

sol's second half — the JSON case emitting neither notice — is gone for a different reason: the two notices are now emitted **independently** rather than as an if/else (`search.go:247-256`). They answer different questions ("sessions matched but `--json` cannot carry them" / "sessions matched but `--kind` hid them") and neither is allowed to swallow the other. Each is a no-op when its own condition does not hold, so no ordering rule is needed. Today they cannot both fire — `SearchSessionHits` returns empty whenever kind is set (`search_projection_session_query.go:33-35`) — but this function no longer has to know that.

- **The fail-closed premise is real, not assumed.** `search_projection_session_query.go:31-35` returns an empty page rather than unfiltered session rows when kind is set. The notice is therefore reporting a genuine suppression, not a display quirk.
- **The probe cannot silently drop a filter added later.** `criteria.WithoutKind()` (`application/types/event_search_criteria.go`) clears one field on a copy rather than rebuilding the criteria field by field through the builder. A filter added to `EventSearchCriteria` next quarter narrows the probe automatically. `TestRootCLI_SearchKindProbeKeepsEveryOtherFilter` drives nine flags and asserts `cmp.Diff(criteria[0].WithoutKind(), criteria[1])` is empty, so this cannot regress unnoticed.
- **Cost is confined to `--kind`.** `TestRootCLI_SearchWithoutKindDoesNotProbe` pins one query, not two, when the flag is absent.
- **Page two does not repeat the notice.** The probe inherits `Offset` / `PageAnchor` from the real criteria, and the session query already returns empty for those (`:40-42`), so the notice belongs to page one — consistent with the SESSIONS group itself.
- **Test integrity.** The announcement test now asserts stderr **exactly**, not by substring. That is what pins the absence of a count and the absence of a spurious JSON notice; a `strings.Contains` on `"--kind"` would have passed against the old wording too. `TestRootCLI_SearchKindProbeIgnoresKindFilteredExclusions` asserts the real call excludes `session-kind` and the probe carries `nil` — sol's finding cannot come back.
- **The stub is faithful to the datasource.** `projectionSessionSearchStub` returns nil when kind is set, mirroring the real fail-closed branch, so the tests are not passing against a fiction.

### Verification

`go build ./...`, `go test -count=1 ./presentation/cli/ ./application/types/`, `go tool golangci-lint run` (0 issues) — all clean locally. CI green on `d039f7a` (all six checks).

Approved.

PR #1820, opened 2026-08-11
feat: expose session hits through MCP search

### LINKED ISSUE #1727: v0.34.0-53: give the MCP search tool the session tier the CLI got

duck8823:
## What

#1717 made the bounded search projection authoritative for search. History older than the recent window no longer returns events; it returns **sessions**, resolved from summaries and keywords. The CLI got that group. MCP did not.

Through MCP, a query matching only old history returned an empty event list. An agent could not tell that from "nothing was ever recorded", and the session trail the projection exists to preserve was unreachable — on the agent-facing surface, which is the case the session tier was built for.

## Scoped schema change, not a free field

`eventsOutput` is shared by **three** tools — `search`, `list_events`, `get_context` — and all three output schemas declare `additionalProperties: false`. Adding a field to it would change three published schemas and break any consumer validating against a cached copy.

So `search` gets its own output type. `searchOutput` **embeds** `eventsOutput` rather than duplicating it, so the two cannot drift on the event side, and adds `sessions`. The golden diff is confined to the `search` entry:

```
changed tools: ['search']
list_events unchanged: True
get_context unchanged: True
```

`searchOutput.Reasons` deliberately **shadows** `eventsOutput.Reasons`. That is what confines the description change — the new reason values — to the `search` schema while the other two keep theirs. It reads like a mistake, so it is commented and pinned by a test that marshals both an event-level and a session-level reason and asserts they arrive in one array.

## Machine-readable reasons, so un-updated consumers still get the signal

`SearchSessionHits` returns an empty page **without error** both when nothing matched and when the projection is not read-ready. An empty `sessions` inherits that ambiguity. Two values are added to the existing `reasons` array — a new *value*, not a new field, so a client still pinned to the old schema and ignoring `sessions` entirely still receives the signal:

| value | meaning |
|---|---|
| `session_matches` | older history matched and is reported in `sessions` |
| `session_projection_not_ready` | the projection has never completed, so older history could not be consulted at all |

The readiness check runs only when there are no hits, since that is the only branch that can use the answer.

Sessions resolve through the **same** query service the CLI uses (`queryservice.ProjectionSessionSearchQuery`), so the two surfaces cannot disagree about which sessions match, and session ids already present in `events` are excluded exactly as the CLI does.

## Tests

Behaviour tests run through a real in-memory MCP client against a real store, not against the converters:

| behaviour | assertion |
|---|---|
| the issue's own case | a match outside the recent event window returns the session in `sessions` with `session_matches` |
| no double reporting | a session already in `events` is **not** repeated in `sessions` — and `events` is asserted too, so the empty `sessions` cannot pass because the search matched nothing |
| never-completed projection | empty `sessions` **and** `session_projection_not_ready` — routed through the real datasource, so the optional-interface wiring is what is under test |
| the other two tools | `list_events` and `get_context` response shapes unchanged |
| the `Reasons` shadow | an event reason and a session reason serialize together, once each |

`go build ./...`, `go test -count=1 ./...` and `go tool golangci-lint run` (0 issues) clean.

Out of scope, unchanged: the CLI `search --json` envelope (still a bare array in v0.34, per #1717's deprecation notice), session-tier filter semantics (#1728), and any new MCP tool — v0.34 reduces surface, so this is a field on an existing tool or nothing.

Closes #1727


duck8823:

Reviewed against the failure class that matters on this surface: a **silent false negative** — MCP reporting "nothing found" for history that exists.

### sol HIGH 1 — readiness was optional, and its absence read as "ready". Fixed here.

> `presentation/mcpserver/server.go:943-949` — `projectionSessionSearch` が `ProjectionSessionSearchReadiness` を実装しない場合、`ready := true` のまま成功します。（…）`WithProjectionSessionSearch` と既存のcompile-time assertionは検索interfaceしか要求しないため、この退行はビルドでもテストでも検出されません。

Correct, and this is the dangerous shape: **permanent, and undetectable by build or test**. Drop `SearchSessionProjectionReady` from the datasource and MCP keeps compiling, keeps passing, and quietly stops emitting `session_projection_not_ready` forever — the exact signal this PR exists to add.

Fixed by making readiness part of the port MCP holds rather than something it hopes for:

- `queryservice.ProjectionSessionSearch` composes query + readiness;
- `Server.projectionSessionSearch` and `WithProjectionSessionSearch` take it, so the type assertion and its `ready := true` default are gone;
- `var _ queryservice.ProjectionSessionSearch = (*EventDatasource)(nil)` at the implementation.

The regression is now a compile error at `main.go:254`.

### sol HIGH 2 — hits and readiness come from two transactions. Real; filed as #1822, not fixed here.

> `SearchSessionHits` が空だった後、別のDB transactionで readinessを再確認しています。（…）検索結果とreadinessを同じsnapshotで取得する必要があります。

Confirmed. Both directions are reachable: a generation publishing between the two reads yields an empty `sessions` with **no reason at all**; a rebuild starting between them yields `session_projection_not_ready` for a search that was in fact served.

Not fixed in this PR, for reasons of shape rather than effort. The honest fix returns hits and state from one call, and it needs a **third** state — `SearchSessionHits` returns before opening the database at all when the query is empty, when `--kind` is set (#1728's fail-closed path) and when the caller paged past the first page, so neither "ready" nor "not ready" describes those. That retires `ProjectionSessionSearchReadiness` as a separate port and touches the CLI, which holds the same port. Widening #1727 into a query-service refactor at RC is the wrong trade for a window that opens once per generation and self-heals on the next call. Filed with the full design and the call sites as **#1822**.

Worth recording: the datasource already computes the answer in the right snapshot (`search_projection_session_query.go:58-67`) and discards it. That is why the caller has to ask twice, and why the fix is a return value rather than a lock.

- **The scoped schema claim holds.** `list_events` and `get_context` goldens are byte-identical; only the `search` entry changes. This was the constraint that shaped the design — three tools share `eventsOutput` and all three declare `additionalProperties: false`.
- **The `Reasons` shadow does not duplicate a JSON key.** Go's promotion rules give the shallower field the `reasons` tag; the embedded one is unreachable in the marshalled output. Pinned by a test that marshals one event-level and one session-level reason and asserts they arrive in a single array — which is also what stops the shadow being read as a mistake and "fixed".
- **Event-level reasons are not dropped.** `result.Reasons` is seeded from `output.Reasons` at construction and every later append targets `result`; nothing appends to `output` after that point.
- **CLI and MCP cannot disagree about which sessions match.** Both resolve through `queryservice.ProjectionSessionSearchQuery`, and both exclude session ids already present in the event page.
- **No error is swallowed.** The readiness error is now its own variable rather than reassigning `searchErr`, so a failed readiness read cannot be mistaken for a successful empty search. Both paths return the error rather than degrading to "no sessions".
- **Test integrity.** The double-report test asserts `events` as well as an empty `sessions`, so it cannot pass because the search matched nothing — this was added after luna's first delivery, which could have passed vacuously. The never-completed-projection case routes through the real datasource, so what is under test is the wiring, not a stub agreeing with itself.

### Delivery notes

luna's first delivery landed two pure-helper unit tests against four required behaviour tests, leaving `session_projection_not_ready` — the path through the optional-interface assertion — with no coverage at all, and reported `golangci-lint ✅` when the real run failed on an unchecked `db.Close` (`search_behavior_test.go:135`). Both corrected before this PR opened; recorded because every delegated verification claim here is re-run rather than believed.

### Verification

`go build ./...`, `go test -count=1 ./presentation/mcpserver/ ./infrastructure/sqlite/`, `go tool golangci-lint run` (0 issues) clean after the readiness change.

Approved once CI is green on `1b29f2c`.

PR #1823, opened 2026-08-11
refactor: give the search projection default budget one owner

### LINKED ISSUE #1806: unify the search projection default budget so the CLI and catch-up cannot drift apart
Raised while reviewing #1805 (#1794) and deliberately left out of it.

The bounded search projection's default budget exists twice: as cobra flag defaults in `presentation/cli/store_search_projection.go` and as `defaultSearchProjectionCatchUpBudget()` in `infrastructure/sqlite/search_projection_catchup.go`. They agree today on all eight fields, and the catch-up function's comment says why they must:

> It matches the CLI search-projection defaults so an operator-started generation with default flags and the auto path share a config hash and can resume each other.

Nothing enforces it.

## Why it matters more after #1805

Before #1805, `ConfigHash` covered `DecodedBytes`, `RecentAge` and `IndexFamilyBytes`. #1805 widened it to `v4` and added `StoredBytes` and `WriteBytes`, because an over-budget row no longer fails the generation — it is excluded, so those two budgets now decide contents.

So the number of fields that must stay in lockstep went from three to five, and the failure mode is quiet: change `--stored-bytes`'s default without changing the catch-up budget and `traceary store search-projection resume` refuses the generation the automatic catch-up started, reporting only "budget does not match generation configuration". The operator has no way to see that two defaults disagree.

## What to decide

- Whether the defaults should have **one owner** — a `DefaultSearchProjectionBudget()` in `application/types` that both the flag defaults and the catch-up path read — so parity holds by construction and no test is needed.
- Or whether the two are allowed to diverge deliberately, in which case the invariant belongs in a test that pins both config hashes, and the catch-up comment should stop claiming they match.

The first is the smaller change and removes the hazard rather than detecting it. Note that the flag *help text* should keep showing concrete numbers either way.

## Related

- #1805 / #1794 — widened the hash to `v4`.
- #1803 — the adaptive shrink that depends on `Rows` staying **out** of the hash; whatever owns the defaults must not quietly pull it in.

duck8823:
## What

The bounded search projection's default budget existed twice — cobra flag defaults in `presentation/cli/store_search_projection.go` and `defaultSearchProjectionCatchUpBudget()` in `infrastructure/sqlite/search_projection_catchup.go`. They agreed on all eight fields, by convention, and the catch-up comment said they had to. Nothing made them.

Five of those fields feed `ConfigHash` (`v4:<storedBytes>:<decodedBytes>:<writeBytes>:<recentAge>:<indexFamilyBytes>`), so the failure mode is quiet and the diagnostic is useless: change one default on one side and `traceary store search-projection resume` with default flags is refused the generation automatic catch-up started, reporting only *"budget does not match generation configuration"*. Both budgets look like the default to the operator, so there is nothing to compare.

## How

`apptypes.DefaultSearchProjectionBudget()` owns the eight values. The flag defaults read its fields; the catch-up budget returns it. Parity is structural — the two cannot disagree without someone editing the owner, which changes both.

`--help` still prints concrete numbers, because cobra derives `(default N)` from the value passed to `Int64Var` / `DurationVar` / `IntVar`. Verified by diffing `store search-projection resume --help` before and after: the only difference is #1810's wording change, every default is byte-identical.

`ConfigHash()` is untouched — no new field, no version bump. `Rows` stays out of it, as #1803's adaptive shrink requires: the shrink varies `Rows` within a generation, so it cannot identify one. That is now written down at the owner rather than being an unstated property of two literal lists.

The explanation of *why* the two must agree moved to the owner as well. Making the hazard impossible removed the place where the reason was recorded, which is how a constraint gets re-broken later.

## Test

`TestSearchProjectionDefaultFlagsMatchCatchUpBudget` builds the real root command, resolves `store search-projection start` and `resume`, reads the eight flag values through cobra, and compares them against `defaultSearchProjectionCatchUpBudget()` — value and `ConfigHash()`. It reaches the CLI wiring, not the shared constant, because the wiring is where the hazard lives; a test calling `DefaultSearchProjectionBudget()` twice would prove nothing.

Verified it fails on drift rather than merely passing: reverting one flag default to a literal `4<<20` produces

```
StoredBytes: 8388608 -> 4194304
default CLI config hash = "v4:4194304:…", catch-up config hash = "v4:8388608:…"
```

The test lives in `package sqlite` so it can reach the unexported catch-up budget, which means an infrastructure test importing `presentation/cli`. Accepted deliberately: the invariant spans both layers, and pinning it against the real catch-up function detects drift on either side in one assertion. There is existing precedent for the direction in tests (`infrastructure/sqlite/memory_datasource_test.go`).

No operator-visible behaviour change. `go build ./...`, `go test -count=1 ./...`, `go tool golangci-lint run` (0 issues) clean.

Closes #1806


duck8823:

The failure this PR is about is silent: a drifted default refuses a resume with an error that names neither default. So the review question is the mirror image — **did the refactor itself change a default?** If it did, every in-progress generation becomes unresumable on upgrade, for exactly the reason the PR exists to prevent.

### What I verified

- **No default changed.** Compared the eight pre-change literals against the new shared value field by field, and diffed `store search-projection resume --help` built from `main` against the same command built from this branch. The only difference in the whole help output is #1810's wording change; every `(default N)` is byte-identical, including `1535115264`.
- **The help text still carries concrete numbers.** Cobra derives `(default N)` from the value handed to `Int64Var` / `DurationVar` / `IntVar`, so reading fields off the shared value preserves it. Confirmed from the built binary, not from the code.
- **`ConfigHash()` is untouched** — same `v4` prefix, same five fields, no version bump. `Rows` stays out of it, which #1803's adaptive shrink requires: the shrink varies `Rows` within a generation, so it cannot identify one. That property was previously only implicit in two literal lists; it is now stated at the owner.
- **The test can fail.** A test that cannot fail is worse than no test, so I checked rather than assumed: reverting one flag default to a literal `4<<20` produces

  ```
  StoredBytes: 8388608 -> 4194304
  default CLI config hash = "v4:4194304:…", catch-up config hash = "v4:8388608:…"
  ```

  It reads the real cobra wiring through `root.Find(...)` and `Flags().GetInt64(...)`, which is the point — the hazard is the wiring, not the constant, and a test calling `DefaultSearchProjectionBudget()` twice would prove nothing.
- **No third copy exists.** The remaining occurrences of `8 << 20`, `30 * 24 * time.Hour` and `128` in projection-adjacent code are either unrelated or test-local values.

### gpt-5.6-sol

Independent read of the same five questions: no CRITICAL or HIGH findings, and no design objection. It confirmed the defaults are unchanged, the hash is untouched, the help keeps concrete values, the test exercises the real wiring and would fail on drift, the `package sqlite` → `presentation/cli` test import is cycle-safe, and no third copy of the defaults exists.

### One thing I changed after luna's delivery

Making parity structural also deleted the only place that recorded **why** it mattered — the catch-up comment carried it. A constraint whose reason is not written down is a constraint that gets broken again. Moved to the owner, together with why `Rows` belongs in the value but not in the hash.

### On the test's location

`TestSearchProjectionDefaultFlagsMatchCatchUpBudget` lives in `package sqlite` so it can reach the unexported catch-up budget, which means an infrastructure test importing `presentation/cli`. Accepted deliberately: the invariant spans both layers, and asserting against the real catch-up function detects drift on either side in one place. Splitting it into "CLI equals owner" and "catch-up equals owner" would be logically equivalent and avoid the import, but the second half is a tautology against a one-line delegation. Precedent exists for the direction in tests (`infrastructure/sqlite/memory_datasource_test.go`).

### Verification

`go build ./...`, `go test -count=1 ./application/types/ ./presentation/cli/ ./infrastructure/sqlite/`, `go tool golangci-lint run` (0 issues). CI green on all six checks.

Approved.

PR #1824, opened 2026-08-11
fix: make the hygiene scan duration tests deterministic

### LINKED ISSUE #1802: v0.34-follow-up: TestMemoryHygieneScan_RevisionChurnReturnsPartialOnlyAfterDurationExhausts is flaky on CI
`TestMemoryHygieneScan_RevisionChurnReturnsPartialOnlyAfterDurationExhausts` fails intermittently on CI. Observed on PR #1800, whose diff does not touch memory hygiene:

```
--- FAIL: TestMemoryHygieneScan_RevisionChurnReturnsPartialOnlyAfterDurationExhausts (0.00s)
    memory_usecase_hygiene_test.go:656: revision-churn Scan() error =
      memory hygiene continuation cannot progress:
      time_limit budget stopped before advancing the cursor:
      memory hygiene continuation cannot progress
```

Re-running the same commit passed. Locally it passes 20/20 (`go test ./application/usecase/ -run ... -count=20`).

## Why it is unstable

The test pins two competing budgets a few milliseconds apart (`application/usecase/memory_usecase_hygiene_test.go:639-651`):

```go
MaxDuration: time.Millisecond,
...
query.scanDelay = 5 * time.Millisecond
```

and then asserts the scan stops with `MemoryHygieneStopReasonRevisionChanged` rather than the time limit. The outcome depends on the goroutine being scheduled between those two deadlines. A shared CI runner does not guarantee that, and when the time budget expires before the cursor advances the scan reports the non-progressing continuation error instead.

The behaviour under test — that a revision change wins over an exhausted duration — is real and worth holding. The wall clock is the wrong instrument for holding it: 1 ms is below the scheduling noise floor of the machines this runs on.

## What to decide

Make the ordering deterministic rather than probable. The scan already receives its clock and its delay through the test double, so the two events can be sequenced explicitly instead of raced — e.g. drive the elapsed time from an injected clock the stub advances at a known point, rather than sleeping and hoping.

Widening the margins (10 ms vs 100 ms) would lower the failure rate without removing the race, so it is a worse answer than removing the timing dependency.

## Impact

A red `Test` job on an unrelated PR, requiring a manual re-run before merge. Merging on red is not acceptable in this repository, so every occurrence costs a full CI cycle.

duck8823:
## What

`TestMemoryHygieneScan_RevisionChurnReturnsPartialOnlyAfterDurationExhausts` failed intermittently on CI — on PR #1800, whose diff does not touch memory hygiene:

```
memory hygiene continuation cannot progress:
time_limit budget stopped before advancing the cursor
```

It raced two independent wall-clock deadlines: a budget of `MaxDuration: time.Millisecond` against a stub that slept `5 * time.Millisecond`. The scan checks elapsed time at the top of every iteration, *before* fetching the first page (`memory_hygiene.go:131`). On a loaded runner more than 1 ms can pass between `startedAt` and that check, so the scan stopped with `time_limit` before the cursor moved — and a partial result that made no progress is a typed error. Re-running the same commit passed; locally it passes 20/20. That is the signature of a scheduling race, not a logic bug.

The behaviour under test — a revision change wins over an exhausted duration — is real and worth holding. The wall clock is the wrong instrument for holding it: 1 ms is below the scheduling noise floor of the machines this runs on.

## How

Not by widening the margins. 10 ms against 100 ms would lower the failure rate without removing the race, and it would come back the first time CI got slower.

The elapsed-time reads in the hygiene scan now go through an injected clock. `MemoryUsecaseDependencies` — already a variadic optional struct — gains `NowFunc func() time.Time`, defaulting to `time.Now`, so **no existing call site changes**. The two tests inject a mutex-guarded fake clock that the scan stub advances by a known amount at a known point, instead of sleeping and hoping the scheduler cooperates. Production behaviour is unchanged: `nowFunc` is `time.Now` everywhere outside these tests.

Two details worth recording:

- `finishPartialMemoryHygieneResult` used to take `startedAt` and call `time.Since` itself, which is a second, hidden reading of the clock. It now takes the already-computed `elapsed`, so there is exactly one place elapsed time is measured.
- `context.WithTimeout(ctx, budget.MaxDuration())` deliberately still runs on the real clock, and the tests do not depend on it firing. Faking a context deadline would have meant faking the runtime, and the assertions do not need it: the fake clock reaches the budget on the first stub call, so the branch under test is chosen before the real deadline is relevant either way.

## Verification

Determinism is the whole point, so it was measured rather than asserted:

| run | result |
|---|---|
| `go test -race -count=200` on the two tests | ok |
| 8 concurrent `go test -race -count=30 -run TestMemoryHygieneScan` on this machine | 8/8 ok |

`go build ./...`, `go test -count=1 ./...` and `go tool golangci-lint run` (0 issues) clean.

No CHANGELOG entry: no operator-visible behaviour changes.

Closes #1802

### DIFF
--- a/application/usecase/memory_hygiene.go
+++ b/application/usecase/memory_hygiene.go
@@ -20,6 +20,14 @@ type memoryHygieneUsecase struct {
 	memory              memoryHygieneWriter
 	memoryQuery         queryservice.MemoryQueryService
 	extraRedactPatterns []string
+	nowFunc             func() time.Time
+}
+
+func (u *memoryHygieneUsecase) now() time.Time {
+	if u.nowFunc == nil {
+		return time.Now()
+	}
+	return u.nowFunc()
 }
 
 // defaultStalenessThreshold controls the expiry-suggestion window when
@@ -101,7 +109,7 @@ func (u *memoryHygieneUsecase) Scan(ctx context.Context, criteria apptypes.Memor
 		Consistency:       consistency,
 		ConsistencyReason: consistencyReason,
 	}
-	startedAt := time.Now()
+	startedAt := u.now()
 	finishPartial := func(reason apptypes.MemoryHygieneStopReason) (apptypes.MemoryHygieneScanResult, error) {
 		return finishPartialMemoryHygieneResult(
 			result,
@@ -113,7 +121,7 @@ func (u *memoryHygieneUsecase) Scan(ctx context.Context, criteria apptypes.Memor
 			consistencyReason,
 			digest,
 			now,
-			startedAt,
+			u.now().Sub(startedAt),
 			initialPhase,
 			initialKeyset,
 			initialRevision,
@@ -123,7 +131,7 @@ func (u *memoryHygieneUsecase) Scan(ctx context.Context, criteria apptypes.Memor
 	defer cancel()
 
 	for {
-		if time.Since(startedAt) >= budget.MaxDuration() {
+		if u.now().Sub(startedAt) >= budget.MaxDuration() {
 			if _, hasRevision := revision.Value(); hasRevision {
 				return finishPartial(apptypes.MemoryHygieneStopReasonTimeLimit)
 			}
@@ -173,7 +181,7 @@ func (u *memoryHygieneUsecase) Scan(ctx context.Context, criteria apptypes.Memor
 				result.Consistency = consistency
 				result.ConsistencyReason = consistencyReason
 				revisionChangedWhileRetrying = true
-				if errors.Is(scanCtx.Err(), context.DeadlineExceeded) || time.Since(startedAt) >= budget.MaxDuration() {
+				if errors.Is(scanCtx.Err(), context.DeadlineExceeded) || u.now().Sub(startedAt) >= budget.MaxDuration() {
 					return finishPartial(apptypes.MemoryHygieneStopReasonRevisionChanged)
 				}
 				continue
@@ -194,7 +202,7 @@ func (u *memoryHygieneUsecase) Scan(ctx context.Context, criteria apptypes.Memor
 		result.Usage.Comparisons += page.Comparisons
 
 		for _, unit := range page.Units {
-			if time.Since(startedAt) >= budget.MaxDuration() {
+			if u.now().Sub(startedAt) >= budget.MaxDuration() {
 				return finishPartial(apptypes.MemoryHygieneStopReasonTimeLimit)
 			}
 			matches := u.matchesForScanUnit(phase, unit, now, staleness, similarity)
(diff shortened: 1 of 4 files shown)


duck8823:

A flaky test has two bad fixes: widen the margins until it stops failing, or delete the assertion. Both leave a green suite that no longer tests the thing. So the review question is not "is it still green" but **does the test still fail when the behaviour it names is broken**, and **did fixing the test change production**.

### What I verified

- **The clock never reaches production.** `NowFunc` defaults to `time.Now` in the constructor, and every elapsed read goes through the `now()` helper, which falls back to `time.Now` when the field is nil. There is no path that calls the field directly, so a zero-valued `MemoryUsecaseDependencies` — which is what every existing call site passes — behaves exactly as before. No call site changed.
- **The margins were not widened.** This is the fix I was most alert to, because it is the easy one. The budget is still 1 ms; what changed is that 1 ms is now measured on a clock the test controls, so the branch under test is reached deterministically rather than probabilistically.
- **Determinism was measured, not asserted.** `-race -count=200` on the two tests, plus eight concurrent `-race -count=30` runs to reproduce the loaded-runner conditions the original flake needed. All green. sol independently ran 400 iterations of the pair and 875 of the whole `application/usecase` package under `-race`.

### One real difference in production, and why it is acceptable

`finishPartialMemoryHygieneResult` used to call `time.Since(startedAt)` itself; it now receives the already-computed `elapsed`. sol correctly noticed that this moves the sampling instant *earlier* — before the cursor is encoded (`memory_hygiene.go:525-555`) — so `elapsed_ms` can under-report by the encode duration.

I checked what that duration is: a JSON marshal and base64 of a seven-field struct. It is sub-millisecond, and `elapsed_ms` has millisecond resolution, so in practice it does not appear at all. More to the point, the new sampling point is the more correct one: elapsed should measure the scan, not the serialization of its result. The value feeds no budget decision, no stop reason and no cursor-progress check — it is reported, not consumed.

### On the context deadline

`context.WithTimeout(ctx, budget.MaxDuration())` deliberately still runs on the real clock, and I wanted to be sure that did not leave the race in place, only rarer. It does not: the stub does not observe the context, it advances the fake clock and returns the revision error synchronously, and the revision branch is evaluated before the time-limit branch (`memory_hygiene.go:173-185`). Even if the real 1 ms deadline fires first, the outcome is `revision_changed`. Faking the deadline as well would have meant faking the runtime for no additional coverage.

### Coverage is not weaker

The assertion I was worried about losing is the ordering one — "a revision change wins over an exhausted duration" degenerating into "the fake clock was advanced". It survives: the stub advances the clock past the deadline *and then* changes the revision, and the test requires `revision_changed`, the best-effort consistency, and the unconsumed rows separately. If the duration won, the stop reason would be `time_limit` and the test fails.

### Verification

`go build ./...`, `go test -count=1 ./...`, `go tool golangci-lint run` (0 issues). No CHANGELOG entry, correctly — nothing operator-visible changed.

sol reports no CRITICAL or HIGH findings.

Approved.

Issue #1825, opened 2026-08-11
v0.37-14: Rework projection cleanup addressing before any WITHOUT ROWID conversion


duck8823:
Split out from #1808, which proposed three space reclamations. The third — dropping the dead `idx_literal_search_fingerprint_candidate` — is done. The two `WITHOUT ROWID` conversions are not, because they are blocked by a design coupling that has to be removed first.

## The blocker

Projection cleanup addresses rows by `rowid`.

`selectProjectionCleanup` (`infrastructure/sqlite/search_projection_rebuild.go:680,683`) UNION-ALLs six projection tables, selecting `(class, rowid, logicalBytes)` from each, and `applyProjectionPlan` deletes the chosen rows with `DELETE FROM <table> WHERE rowid = ?`. It treats `RowsAffected() != 1` as `SearchProjectionDriftError`.

A `WITHOUT ROWID` table has no `rowid`. Converting either of the candidate tables makes the `SELECT` fail to compile, and there is no smaller fix available: the plan carries a single integer per row, so the delete path cannot address a composite primary key without changing `ProjectionCleanupCandidate` and every producer and consumer of it.

That matters more than the disk it would save. The cleanup path is on the eviction phase of the bounded rebuild, and its failure mode is a **permanent stall**: a rebuild that cannot evict cannot complete, and every retry repeats identically. Reworking it to carry a per-class composite key is a real design change to the machinery whose stalling is the most severe failure this project has — not something to do alongside a `DROP INDEX` at RC.

## What this issue is

Decide whether the conversions are worth it, and if so do the addressing rework first.

Two parts, in order:

1. **Make cleanup addressing table-shaped rather than rowid-shaped.** The candidate needs to carry whatever identifies a row in its own table, and the delete needs to be built per class rather than shared. The `RowsAffected() != 1` drift check must survive the change — it is what turns a silently wrong delete into a loud one, and losing it would trade a stall for a corruption.

2. **Then measure whether the conversions actually pay.** #1808 estimated the saving from the duplicate autoindexes, but the estimate predates the projection changes that landed in v0.34 (#1796's bounded peak, #1803's adaptive shrink, and the index-family measurement now used to derive the source ceiling). Re-measure on a current store before spending the rework: if the saving is small relative to the family budget, the honest answer is to close this and keep the rowid path.

## Not urgent

[...]

PR #1826, opened 2026-08-11
perf: drop the unused literal fingerprint candidate index

### LINKED ISSUE #1808: v0.34.0: session-keyword and fingerprint tables store their key twice via implicit autoindexes

duck8823:
## What

Migration `000059` drops `idx_literal_search_fingerprint_candidate` from `literal_search_fingerprints`.

The index is dead. The table's primary key is `(generation_id, event_id, fingerprint)`, and both fingerprint readers filter on a prefix of it, so SQLite resolves them through the implicit `sqlite_autoindex_literal_search_fingerprints_1` and never consults the candidate index. Every other statement touching the table is an `INSERT`, a full-table aggregate, or the cleanup `rowid` scan — none of which can use it either.

It cost **118,489,088 bytes** on a dogfood store, against a table of 118,796,288 and an autoindex of 114,008,064. It was as large as the data it indexed, and nothing read it.

## Why this is safe to do as a migration

Dropping an unused index changes no row and no query result, only the physical layout. Unlike `000058`, which had to abandon in-flight generations because it changed an invariant, this one invalidates nothing — the SQL says so in a leading comment so the next person does not have to re-derive it.

No `VACUUM`. The freed pages become free pages inside the file and are reused by the store; they are not returned to the filesystem until `traceary store compact`. Adding a `VACUUM` to the migration would put unbounded work at store-open time, which is precisely the class of thing the prepared-migration catalog exists to keep out. The CHANGELOG states the reuse-not-return behaviour instead, because that is the part an operator watching a file size would otherwise misread.

There is a second-order effect worth naming: `literal_search_fingerprints` is inside the measured projection index family, so `--index-family-bytes` was spending 118 MB of its budget on an index nobody reads. After this migration the same budget buys that much more real index, which raises the derived source ceiling. That is the intended direction — the budget is still enforced against the measured family; it simply stops being consumed by dead weight.

## Scope

#1808 proposed three changes. **This PR implements only the third.** The two `WITHOUT ROWID` conversions are deliberately not here: projection cleanup addresses rows by `rowid` (`search_projection_rebuild.go:680,683` select `(class, rowid, logicalBytes)`; `applyProjectionPlan` deletes with `WHERE rowid = ?` and treats `RowsAffected() != 1` as drift), and a `WITHOUT ROWID` table has no rowid. Converting either table therefore requires reworking the candidate type and the per-class delete path — the machinery whose failure mode is a permanent rebuild stall. Filed as #1825, with the addressing rework ordered before any conversion and a re-measurement gate after it.

## Verification

| check | result |
|---|---|
| `go build ./...` | ok |
| `go test -count=1 ./...` | ok (whole tree) |
| `go tool golangci-lint run` | 0 issues |
| catalog SHA-256 vs the file on disk | `c1403cd2…` matches |

The new test was falsified rather than trusted. Neutering the migration to `SELECT 1;` (with the catalog digest updated to match, so the failure could not be a digest artefact) makes it fail on the assertion it is supposed to own:

```
--- FAIL: TestTieredAuthorityFingerprintReadersUsePrimaryKeyAfterCandidateIndexDrop
    candidate index still exists (-want +got):
      int(
    - 	0,
    + 	1,
      )
```

Its query-plan half is a guard, not a regression detector, and it is worth being explicit about that: inverting the existence assertion to expect the index *present* leaves the plan assertions passing. The planner already preferred the autoindex before this change. That is the evidence the drop is safe — the readers were never on this index — but it does mean the plan assertions can only catch a *future* reader that starts depending on it, which is what they are there for.

Closes #1808 — the remaining two changes it proposed are carried by #1825, which states the blocker and orders the rework before them, so nothing is lost by closing the original.


duck8823:

Dropping an index is either a no-op with a smaller file or the beginning of a permanent stall, and nothing in between. sol returned two HIGH findings, both arguing for the second. Both were checkable, so both were checked against the real 7.1 GB dogfood store rather than reasoned about.

### sol HIGH 1 — "the index is not fully dead: the ON DELETE CASCADE child lookup uses it"

**The premise is right and the consequence is not.** The foreign key is real (`000039:23`), `source_sequence` is the third column of the candidate index and appears in no other index, and the primary key cannot substitute for it. So this is the one plan that named the index, and I had not checked it before delegating — my earlier evidence covered only the two reader subqueries. That gap was real.

What the store actually says:

```
before: SEARCH search_projection_source_sequence USING INTEGER PRIMARY KEY (rowid=?)
        SCAN literal_search_fingerprints USING COVERING INDEX idx_literal_search_fingerprint_candidate
after:  SEARCH search_projection_source_sequence USING INTEGER PRIMARY KEY (rowid=?)
        SCAN literal_search_fingerprints
```

A full **SCAN** both times. sol's O(M×N) scenario requires the seek form it quoted — `(ANY(generation_id) AND ANY(fingerprint) AND source_sequence=?)` — which is a skip-scan, and skip-scan requires `ANALYZE`. `ANALYZE` appears nowhere in this repository, and `sqlite_stat1` does not exist on the real store. So the cascade was already a full scan; the drop changes which comparable structure it scans, not its order of growth.

It also has no caller. `search_projection_source_sequence` is insert-only in production (`INSERT OR IGNORE`, `search_projection_rebuild.go:502`), it is absent from the six-table cleanup union, and no trigger deletes from it. The only deletes are in tests.

What sol's finding did earn is a test. Performance was never the risk here; **correctness** is — a foreign key that quietly stops cascading leaves orphaned fingerprints that no reader can distinguish from real ones. `TestLiteralFingerprintCascadeSurvivesCandidateIndexDrop` now pins it, and it fails when the cascade does not run (verified by flipping the pragma: `orphaned 0 -> 1`).

### sol HIGH 2 — "a 118 MB DROP INDEX is misclassified as MigrationConstantInPlace"

**Measured: 0.096 s.**

Dropping an index moves its pages to the freelist rather than rewriting them — 31,148 pages freed on the real store. The stall scenario needs the DROP to exceed the migration deadline at store open; 96 ms does not come close, and there is no retry loop to enter.

sol is right that migration `000052` states the policy, and right to hold this change to it. But that comment forbids DROPs that would *block store open* — "a multi-GiB DROP" — not DROPs of large objects as such. The distinction is the measurement, so the measurement is now in the catalog comment rather than the verdict alone. The next person classifying a DROP should see the test that was applied.

- **The readers do not depend on it, on the real store.** After the drop: `SEARCH known USING COVERING INDEX sqlite_autoindex_literal_search_fingerprints_1 (generation_id=? AND event_id=?)` and the same for `matched` with the fingerprint term.
- **The new test can fail.** Neutering the migration to `SELECT 1;` — with the catalog digest updated so the failure could not be a digest artefact — produces `candidate index still exists (-want +got): - 0 + 1`.
- **Its query-plan half is a guard, not a regression detector**, and the PR says so: inverting the existence assertion to expect the index *present* leaves the plan assertions passing, because the planner already preferred the autoindex. That is the evidence the drop is safe; it also means those assertions can only catch a future reader that starts depending on the index.
- **The plan assertion was brittle and is no longer.** It matched the literal phrase `using index`, while the real store says `USING COVERING INDEX` — the fixture happened to say the other one. It now matches per row without the word in between, because which form SQLite picks is not the property under test.
- **The catalog SHA matches the file on disk**, recomputed after the comment rewrite.
- **Family accounting.** `literal_search_fingerprints` is inside the measured projection index family, so `--index-family-bytes` was spending 118 MB on an index nobody reads. The budget now buys that much more real index and the derived source ceiling rises accordingly — the intended direction, and still enforced against the measured family. `ConfigHash` contains no measured quantity, so no generation is invalidated mid-rebuild.

### Verification

`go build ./...`, `go test -count=1 ./...`, `go tool golangci-lint run` (0 issues), all green after the amendments. CI green on all six checks.

Both HIGH findings are answered on evidence rather than dismissed, and both left something behind: a cascade test and a measurement in the classification comment.

Approved.

PR #1828, opened 2026-08-11
feat: show what a retention plan discards, not just what it reclaims

### LINKED ISSUE #1786: retention plan selects candidates whose pruning reclaims nothing
Found while reviewing #1759 (PR #1784), which fixes the reported figure. This is about **selection**, which that PR deliberately leaves alone.

## Background

`store retention plan` reported the decoded plaintext length of each candidate body as the space applying the plan would reclaim. #1759 makes it the encoded extent, and then the net extent: applying the plan does not empty the body, it replaces it with `EventBodyUnavailableRetentionMarker` (`"[traceary:body-unavailable:retention]"`), identity-encoded. So the real change to the body column is the candidate's encoded extent minus the marker's.

## What is left

Selection does not consider that arithmetic. A candidate whose encoded extent is at or below the marker's contributes zero or negative bytes, and pruning it still:

- discards the body **irreversibly**,
- writes `body_pruned_at`, `body_pruned_plan_id` and, on legacy rows, codec metadata columns that were previously NULL,
- consumes a ledger entry and a recovery obligation.

The operation costs storage and buys nothing. With bodies now compressed by default after #1618 and #1685, small encoded extents are ordinary rather than exceptional, so this is not a corner case.

## What to decide

Whether the planner should exclude candidates whose net contribution is not positive, and if so with what reason string, so the operator can see they were considered and skipped rather than silently missing.

The alternative — keep selecting them and let the signed total speak for itself — is defensible if the ledger entry has value beyond the bytes, for instance as a record that a body was reviewed and deliberately discarded. That is a question about what the retention ledger is for, which is why it is not being decided inside a bug-fix PR.

## Not urgent

#1759's fix removes the misleading number, which was the dangerous part. This is a policy refinement on top of an already-honest figure.

duck8823:
## The decision

#1786 asks whether the planner should **exclude** candidates whose net contribution is zero or negative. The answer is no, and selection is unchanged.

1. `--keep-days` is documented as "number of days of raw bodies to retain". It is a policy about how long bodies persist, not a space optimizer. A body that outlived its window because it happened to be small would be a silent exception to the policy the operator set, and nothing in the output would say which bodies were left behind.
2. Pruning is not as irreversible as the issue implies. `--recovery` is required and must cover every candidate, so the body remains recoverable from the pinned archive.
3. Per-candidate disclosure through `reasons` is not available at any price worth paying: `application/usecase/retention_plan_codec.go:253` rejects a candidate whose reasons are not exactly `["age"]`, so adding a second reason makes every new plan unappliable. A `retention-plan/v3` bump to carry a disclosure string is not a trade worth making.

## What was actually missing

`store retention apply` prints `Candidates: %d`. `store retention plan` — the command the operator reads *before* deciding — printed only the net byte figure. So the number that decides "is this worth doing" was the one the plan withheld, which is the whole premise of #1786: an operator cannot see that a plan destroys many bodies for almost nothing.

```
Plan: /path/plan.json
Plan ID: …
Net body-column change after retention markers are written (encoded bytes): -1
Bodies discarded: 2
Of those, 2 reclaim nothing: their bodies are no larger than the marker that replaces them, and they are discarded anyway.
```

Both figures are derived from the plan JSON this command already parses. The marker's own encoded size is not carried in the plan, but the projected extent is exactly marker × candidate count, so it divides out — the same arithmetic `retentionProjectedBytes` uses to build it.

The second line appears only when the count is non-zero. A line printed on every healthy plan is a line the operator learns to skip, and this one exists precisely for the plan where it is not zero. A payload the reader does not recognise yields no line rather than a guess, because an approximate count here would be read as exact.

`Candidates` became `Bodies discarded`: the cost is measured in bodies that stop being readable, and calling them candidates invites the reading that something might still filter them out. Nothing does — that is the decision this PR records.

## Scope

No change to `CreatePlan`, to `RetentionPlanCandidate`, to the canonical payload, or to `retentionPlanSchemaVersion`. No new flag and no new subcommand. `apply` and `restore` output unchanged.

## Verification

`go build ./...`, `go test -count=1 ./...`, `go tool golangci-lint run` (0 issues).

One note on the delivered tests: as committed they asserted the whole of stdout including the `Plan: <path>` line, which carries `t.TempDir()`, so they could not pass on any machine. They now strip that line — after asserting it is present — and compare the rest exactly. The four cases are every candidate reclaiming bytes, some reclaiming nothing, all reclaiming nothing, and no candidates at all (which also pins that the count is never divided by zero).

Closes #1786


duck8823:

**sol: no CRITICAL or HIGH findings.**

The risk in this PR is narrow and specific: it prints a number the plan does not store. `zeroReclaimCount` recovers the retention marker's encoded size by dividing the projected extent by the candidate count, and a derived figure that is *plausible but wrong* on an operator surface deciding an irreversible discard is worse than no figure at all. That is what I asked sol to attack.

### On the division being sound

sol confirmed it and produced the citation I had not: `raw_body_retention_usecase.go:52-56,242-246` builds `projected = markerEncodedBytes × candidateCount`, and `:227-237` **recomputes the same product at apply time and rejects a plan whose figures disagree**. So the identity the reader inverts is not an incidental property of how the plan happens to be written — it is enforced on the way back in. A plan whose projection did not equal marker × count could not be applied at all.

sol's one reservation is that the guard checks shape (one class result, one ceiling, integral division) rather than class or ceiling *names*, so a future schema with the same shape and a different projection meaning would pass it. That is true and I am leaving it: this command reads a plan the current writer produced moments earlier, and an older or foreign schema is rejected by apply's strict decoder before it can be acted on. There is no plan shape reachable today that clears the guard and yields a wrong exact count.

- **The failure mode is absence, not error.** Every unrecognised shape — no candidates, multiple class results, multiple ceilings, an unparseable or negative byte string, a non-integral division — returns 0, and 0 suppresses the line entirely. There is no path where a partially-understood payload produces a partial number.
- **No division by zero.** `len(p.Candidates) == 0` is the first guard, and the four-case table includes the empty plan explicitly to pin it.
- **`Bodies discarded` is accurate.** Apply processes the candidate list as written; nothing re-filters it, which is exactly the decision this PR records — that is why the label changed from `Candidates`, which invites the reading that something still might.
- **The tests can fail.** They strip only the `Plan: <path>` line, after asserting it is present, and compare the remainder exactly — including the absence of the disclosure line in the all-reclaiming case.

### On the delivered tests

As committed by luna they asserted the whole of stdout including a `Plan: <path>` line carrying `t.TempDir()`, so they could not have passed on any machine. Confirmed by stashing my changes and re-running at that commit. This is the second delivered `go test ✅` in this sprint that did not hold, so I am re-running every delegated verification claim rather than reading it.

### Verification

`go build ./...`, `go test -count=1 ./...`, `go tool golangci-lint run` (0 issues).

Approved.

Issue #1724, opened 2026-08-08
v0.40-6: long-lived active sessions are never consolidated but their old events are deleted


duck8823:
## Problem

Orphan discovery only sees two kinds of session:

1. rows recorded in `session_orphan_ranges` (written at a compact boundary), and
2. sessions that have **ended**, or that the 24h stale rule treats as ended

A session that stays continuously active — activity at least every 24 hours, never an `ended_at` — matches neither, unless a compact boundary happened to record a marker for it. It is therefore never a consolidation candidate.

Deletion has no such restriction. `delete_old_events.sql` is `DELETE FROM events WHERE created_at < ?` against the retention cutoff (default 90 days). So a session that has been active for longer than the retention window has old events that deletion removes and consolidation never folds.

## Why this surfaces now

#1719 added a deletion gate on `OrphanConsolidationResult.Complete()`. That gate closes the hazard it was built for — a bounded or partly skipped pass — but it cannot close this one, because these sessions are never *discovered*, so `HasMore` is false and `Complete()` reports a clean pass. `gc` then deletes with a clear conscience.

The gap predates #1719: discovery on `main` had the same ended-or-stale predicate and deletion had the same cutoff-only rule. #1719 only made it visible by making the safe cases explicit. The `Complete()` doc comment names the gap and points here.

## Scope

Decide what "unfolded" means for a session that has not ended, and make discovery see it:

- the natural rule is coverage-relative rather than lifecycle-relative: a session with events past `covers_to` that are older than the retention cutoff needs folding regardless of whether it is still active
- folding an active session's tail must not claim to cover events that arrive afterwards — the range has to end at a definite event, and later events remain a new orphan range
- the stale-session and marker paths should keep working unchanged; this adds a third source, it does not replace them

## Alternative worth weighing first

Making deletion itself coverage-aware (`DELETE ... WHERE created_at < ? AND <range is folded>`) would close this and #1721 at once, and would make the whole class of "did we fold before deleting" gates unnecessary. It is a change to what `gc` deletes for every user, so it needs its own design pass — but it may be the smaller total change.

## Done when

- A session active for longer than the retention window has its pre-cutoff events folded before `gc` deletes them
- Folding an active session does not claim coverage of events that had not arrived
- A test pins the failing case: a never-ending session with events older than the cutoff, where `gc` currently deletes unfolded

Found during review of #1723 (gpt-5.6-sol, adversarial pass).

PR #1829, opened 2026-08-11
fix: stop gc asking for a re-run it cannot act on

### LINKED ISSUE #1795: store gc asks to be re-run forever for an orphan range it can never process
Found in the #1620 gate rehearsal. `store gc` reaches a state where it consolidates nothing, changes nothing, and still tells the operator to run it again — forever.

## What happens

Running `store gc` repeatedly on the rehearsal copy produced refinements as designed, reaching 26,575 of 27,263 sessions. From then on every further run reported:

```
整理しました: 0
orphan 機械要約: 0
orphan range のスキップ: 1
orphan range が残っています。機械要約を続けるには gc を再実行してください
```

Identical across every subsequent invocation. `session_orphan_ranges` is 0, the refinement count does not move, and the message still asks for another run. An operator following the instruction loops indefinitely.

## Why

The reason only reaches `slog.Warn`, not the operator-facing output:

```json
{"level":"WARN","msg":"skipped orphan range",
 "session_id":"session_db869d4f-92bc-4a64-8261-df0cb4846a7d",
 "error":"failed to load orphan material for session ...: invalid orphan event created_at \"\":
          parsing time \"\" as \"2006-01-02T15:04:05.999999999Z07:00\": cannot parse \"\" as \"2006\""}
```

`orphan_consolidation_usecase_impl.go:122` increments `skipped` and moves on, which is the right call for an isolated bad candidate — the comment says so explicitly, separating "one candidate is bad" from "the mechanism is broken". But nothing records that this candidate is unprocessable, so the next selection returns it again, and `hasMore` stays true. The skip is per-run; the candidacy is permanent.

## The data underneath

Three events carry an empty `created_at`:

```
rowid   id                                session_id                                    kind        source_hook  agent  created_at
595377  32a47a458090ad64cdee4e627274efec  session_db869d4f-92bc-4a64-8261-df0cb4846a7d  transcript  (empty)      kimi   ''
595378  daf34d1da228581ea0697ad24460a020  session_db869d4f-92bc-4a64-8261-df0cb4846a7d  transcript  (empty)      kimi   ''
595379  bad7b9ac7935084ba508ec7fe66c003c  session_db869d4f-92bc-4a64-8261-df0cb4846a7d  transcript  (empty)      kimi   ''
```

Three of 408,893 rows, all in one kimi session, all `transcript`, all with an empty `source_hook` too. The column is `TEXT NOT NULL`, which an empty string satisfies, so the schema does not catch it.

The session itself has valid timestamps (`started_at 2026-07-24T05:48:03.210445Z`, `ended_at 2026-07-25T19:45:15.77799Z`), so this is not a wholly corrupt session — some write path persisted events without a timestamp.

## Two separate things to decide

1. **How were they written?** An event with no `created_at` should not be reachable from any recording path. Whether that is a missing validation in the domain constructor, a hook payload with an absent field, or a datasource that binds an empty default, it needs finding — three rows out of 408,893 is rare enough to be a specific path rather than a systemic one. A `CHECK (created_at <> '')` would make it unwritable, but only after the existing rows are dealt with.

2. **A candidate that can never succeed must stop being a candidate.** Options: mark the range as unprocessable so selection excludes it, or keep re-selecting but stop claiming there is work to do when every remaining candidate failed last time. Either way the operator-facing line has to become false eventually, and the failure reason belongs in the command output rather than only in a warning log — right now nothing on screen says which session is stuck or why.

## Note on the coverage gate

This did not prevent the gate row from passing: 26,575 / 27,263 = **97.5%** against a >= 95% target, and the 688 sessions without a refinement are mostly ones with nothing to fold rather than failures. The defect is the liveness of the command, not the coverage it achieves.

## Related

- #1724 (long-lived active sessions are never consolidated but their old events are deleted) — also about a session that consolidation never completes, from the other direction.
- #1722 (orphan consolidation opens one read-only connection per candidate) — same code path.

duck8823:
## What

`traceary store gc` reached a state where it consolidated nothing, changed nothing, and still asked to be run again — forever:

```
orphan 機械要約: 0
orphan range のスキップ: 1
orphan range が残っています。機械要約を続けるには gc を再実行してください
```

Identical on every subsequent invocation.

## Why

One line. `OrphanConsolidationResult.Complete()` was `!hasMore && skipped == 0`, and the CLI printed "re-run gc to continue consolidation" whenever the result was not complete. So a single skipped candidate produced the re-run prompt on every pass — and re-running selected the same candidate and skipped it for the same reason, because nothing records a candidate as unprocessable and `DiscoverCandidates` is a plain keyset scan.

The skip itself is right: `orphan_consolidation_usecase_impl.go` treats one bad candidate as skippable rather than fatal, and that is the correct call. What was wrong is that the failure reached only `slog.Warn`, and that "something failed" was conflated with "there is more work to do".

In the observed case the candidate can never succeed — three events in one session carry an empty `created_at`, so loading the orphan material fails at parse time every time.

## How

**The completion signal answers one question.** `HasMore()` is now the only predicate, and `Complete()` is gone. It used to mean "no more candidates *and* nothing was skipped"; now that it would mean exactly `!HasMore()`, a second name for the same bit would only invite the same conflation back. Its comment — including the #1724 note about sessions that stay continuously active — moved to `HasMore()`.

**The result carries the failures, not just a count.** `OrphanConsolidationFailures` retains up to ten `(session, reason)` pairs and the true total; `Skipped()` is derived from it, so the count and the list cannot disagree. The cap is not exported — a caller reporting truncation already holds the retained failures, and `len(Items())` is the same number by construction.

**The command says what happened, and does not predict.** Skipped ranges are listed with their session and reason. The listing deliberately does *not* say re-running will skip them again — see below. The "more ranges remain" line is driven by `HasMore()` alone, so it and the failure listing can appear together without contradicting each other: a pass can genuinely have failures *and* remaining work.

### Why the listing does not claim the skips will recur

An earlier revision of this PR had the header say `re-running gc will skip these again`. That is true of the case in the issue and false in general. `produceDegraded` returns `Refine`'s error the same way it returns a load failure, so a CAS retry exhaustion or a lock arrives as the same wrapped error as an unparseable `created_at` — and the command cannot tell them apart.

Shipping that claim would have replaced #1795 with its mirror image. A pass whose only candidate failed on a lock has `HasMore() == false`, so no re-run line appears, and the header would tell the operator not to bother with a range that would have consolidated next pass. A loop that never ends is at least visible; a stall that ends too early is not.

Classifying the two would mean typed errors through the load and refine chain — a real change, and not one to make at RC. So the header states what happened, a line after the listing says re-running retries them and that unreadable data fails the same way each time, and the reason is printed because it is what distinguishes the two cases.

### Terminal safety

Both printed fields come back out of the store. A reason is a wrapped error chain quoting row values; a session id is only trimmed by `SessionIDFrom`. Either can carry an ESC sequence, and elision at a fixed rune count can cut one off from its reset — which would conceal the lines this listing exists to show. Both now go through `normalizeTabularColumn`, the helper `presentation/cli/event_output.go` already uses for the same reason in the event listing.

Reasons are still collapsed to one line and elided at 200 runes: a newline would make a continuation read as another session, and an unbounded reason makes the listing unreadable exactly when there is most to read. The full text still goes to the log.

## Not in this PR

The issue's first question — how events with an empty `created_at` were written, and whether a `CHECK (created_at <> '')` belongs on the column — is untouched. It needs the three existing rows dealt with first, which means a data migration, and that is a different change from making the command honest about what it cannot do.

## Verification

`go build ./...`, `go test -count=1 ./...`, `go tool golangci-lint run` (0 issues).

The exact-stdout table covers a pass that both skipped a candidate and left more to find — the case where the two messages have to coexist — plus a multi-line reason, an over-long reason, control sequences in both fields at once, an escape positioned so elision would sever it from its reset, and the cap with its truncation notice. The two control-character cases were falsified by reverting the strip: both fail, and pass again when it is restored.

The cap is written in the test as a literal rather than read from the type: it is a promise about how much output one run can produce, so the test has to fail when the promise changes instead of agreeing with the implementation.

Closes #1795


duck8823:

sol returned **two HIGH findings and both hold**. Neither is about the fix's logic; both are about the new output, which is the entire operator-facing surface this PR adds. Both are now fixed in `213387bc`.

### sol HIGH 1 — "the listing asserts a permanence the command cannot establish"

**Confirmed, and it is the more serious of the two, because it is #1795 inverted.**

The header read `Skipped orphan range failures (re-running gc will skip these again)`. That is true of the observed case — three events with an empty `created_at`, which fails at parse time on every pass — and it is what the issue is about, so it was easy to write. But `processCandidate` does not only return load failures: `produceDegraded` returns `Refine`'s error the same way (`orphan_consolidation_usecase_impl.go:196`), and a CAS retry exhaustion or a SQLite lock is transient. Both arrive as one wrapped error, and nothing in the result distinguishes them.

The consequence is exactly the failure class this PR exists to remove, pointing the other way. A pass whose only candidate failed on a lock has `HasMore() == false`, so there is no re-run line, and the header tells the operator not to bother. The range would have consolidated on the next run. #1795 was a loop that never terminated; this would have been a stall that terminated too early — and the second is worse, because nothing on screen ever contradicts it.

I did not add error classification. Distinguishing transient from permanent means typed errors through the load and refine chain, which is a real change and not an RC one. The command instead stops predicting: the header states what happened, and a line after the listing says re-running retries them and that unreadable data fails the same way each time. The reason is printed because it is the thing that tells them apart — which is also the argument for printing it at all.

### sol HIGH 2 — "`strings.Fields` does not remove ANSI escapes"

**Confirmed, and the repository already had the answer.**

`normalizeTabularColumn` / `stripTerminalControlChars` (`presentation/cli/event_output.go:94-121`) exist for precisely this, with a comment about ESC / CSI / OSC sequences hijacking the terminal in tabular rows. My `singleLineFailureReason` collapsed whitespace and skipped the strip; ESC is not whitespace, so it passed straight through. The session id was printed with no treatment at all, and `SessionIDFrom` only trims — a control character in one is storable.

sol also noted that elision at a fixed rune count can cut a sequence off from its reset, which turns a shortening into a display that conceals what follows. Stripping first removes the possibility rather than bounding it.

Both fields now go through `normalizeTabularColumn`. Two table cases cover it: control sequences in both fields at once, and an escape positioned so that elision would have severed it from its reset. Falsified by reverting the strip — both fail, and pass again when restored.

- **The abort path is unchanged.** `consecutive` still increments on every failure and still returns an error at the limit; collecting failures did not touch the loop's termination.
- **Removing `Complete()` loses nothing.** It has no remaining consumer — no JSON surface, no MCP tool, no other formatter. It would now return exactly `!HasMore()`, and keeping a second name for one bit is the shape of the bug being fixed.
- **The count and the list cannot disagree.** `Skipped()` derives from `failures.Count()`, `Items()` copies, and `Add` returns a value with a fresh backing array, so two results built from one collection cannot alias.
- **The re-run guidance is genuinely gated.** A pass that skipped everything it found prints no "more ranges remain" line; that property is what #1795 asked for and it survives the wording change.

### Verification

`go build ./...`, `go test -count=1 ./...`, `go tool golangci-lint run` (0 issues), after the fixes.

Both findings were real. The first one mattered: the delivered wording would have shipped a new way to stall, in the PR whose purpose was to remove one.

Issue #1827, opened 2026-08-11
v0.36-12: Decide when the compaction rollback copy is released


duck8823:
Split out from the comment on #1790, which documents the cost but deliberately does not decide the policy.

## What happens today

After `store compact apply` reaches `phase: committed`, the directory holds both the new store and a full copy of the old one:

```
 3,094,962,176  rehearsal.db
33,217,593,344  rehearsal.db.rollback-e59b9aca6ac60852432f077621f09001
```

`PublishRollback` (`infrastructure/sqlite/compaction_files.go:766`) creates it. Nothing removes it. `store compact` offers `plan`, `apply`, `resume`, `status` and `rollback` — none of which says "I am satisfied, release the safety copy".

So the operator who ran `store compact` to reclaim disk has, at the moment it reports success, used **more** disk than before they started, and the only way out is `rm` on a file they were never told the name of.

#1790 makes that cost visible in the refusal message and in `docs/`. This issue is about whether it should exist at all.

## Why it is not obvious

The rollback copy is the recovery route. A defect in the compacted store that only shows up in use — a corrupted index, a lost row, a migration that behaved differently against the compacted layout — is recoverable only while that file exists. Deleting it automatically trades a disk problem for a data-loss risk, and the whole point of the compaction machinery is that the replacement is reversible.

So the question is not "should it be deleted" but **what event proves it is no longer needed**, and that is a judgement nobody has made yet.

## The three candidate answers

1. **An explicit subcommand.** Honest and unambiguous — the operator says when they are satisfied. Costs one more command on a surface this project is actively shrinking, and does nothing for the operator who never learns the command exists.
2. **Release on the next successful open.** Requires no new surface and happens without being asked. But "opened successfully" is a very weak proof that the store is correct, and it puts filesystem work on the open path, which is where unbounded work is least welcome.
3. **Release after a verification step at the end of `apply`.** Strongest proof, and it keeps the decision inside the run that created the copy. Requires deciding what verification is sufficient — and if the answer is "a full integrity check", that is a second pass over a store the operator just spent four minutes compacting.

There is a fourth option worth stating rather than assuming away: **keep it forever and consider the documentation in #1790 sufficient.** That is defensible if the operators who run compact are the same ones who read the docs, and indefensible if the command is ever run unattended.

## What would settle it

[...]

PR #1830, opened 2026-08-11
docs: state the disk cost of store compact instead of discovering it

### LINKED ISSUE #1790: v0.34.0 follow-up: store compact needs 1.1x the store size in free disk, so the escape hatch is unavailable when the store is largest
Found while rehearsing the #1620 release gate. Not a bug — the preflight is doing what it was designed to do — but the design has a consequence worth deciding on deliberately before v0.34 promises a bounded store.

## What happens

`compactionRequiredBytes` (`infrastructure/sqlite/compaction_files.go:606`) asks for the source size plus a 10% margin:

```go
base := uint64(destination)
margin := base / 10
if margin < 64<<20 {
    margin = 64 << 20
}
return base + temporary + margin, margin, nil
```

`Plan` refuses when that exceeds free space. The candidate is assumed to be as large as the source, because `VACUUM INTO` has not run yet and nothing cheaper can bound its output.

So `store compact` needs **1.1x the store size in free disk**, and the store it is asked to shrink is the reason the disk is full. The larger the problem, the less likely the fix is available.

## Measured

30 GB rehearsal copy of the live store:

```
$ traceary store compact plan --db-path .../rehearsal.db
Error: failed to execute CLI command: insufficient free space:
       need 36539352678 bytes, have 31627071488
```

`36539352678 / 33217593344 = 1.10`. After freeing 7 GB by hand the same plan succeeded and the compaction produced:

```
33,217,593,344 -> 3,094,962,176 bytes   (30.9 GiB -> 2.88 GiB, 4m05s)
```

The candidate was **9.3%** of the source. The preflight reserved 1100% of what was actually needed.

The live store is 24.71 GiB, so an operator in the situation this release is about needs **27.2 GiB free** to run the one command that fixes it.

## Why it is not obviously wrong

`VACUUM INTO` cannot report its output size in advance, and a candidate that runs out of disk mid-write is worse than a refusal. Reserving the source size is the only bound that is certainly safe. The rollback copy is also real: the committed run left a 33 GB `rehearsal.db.rollback-<id>` beside the new file, and that is what makes the replacement recoverable.

## What to decide

1. Is a cheaper upper bound available? `dbstat` can measure live page bytes without writing anything, and the candidate cannot exceed that plus overhead. On the rehearsal store that would have estimated a few GB rather than 33.
2. If not, the error should tell the operator what to do rather than only what is missing. `need N, have M` does not say that most of N is a worst-case reservation, that the rollback copy needs room too, or that freeing a few GB may be enough.
3. Either way `docs/` should state the headroom requirement where the store-size promise is made, so it is a known cost rather than a surprise at the moment of need.

Not urgent for the v0.34 gate itself — the gate measurement completed after freeing space by hand — but the gate is exactly the scenario where an operator would hit it.

duck8823:
## The decision

`store compact` is the command an operator runs when the store has grown too large. It needs 1.1x the store size in free disk to start, and it leaves a full copy of the pre-compaction store behind forever. Measured on a 30 GB rehearsal: the preflight demanded 36,539,352,678 bytes, the result was 3,094,962,176 — 9.3% of the source — and a 33 GB `.rollback-<id>` remained after the run committed.

**The preflight arithmetic is unchanged.** `VACUUM INTO` cannot report its output size before writing it, and a candidate that runs out of disk mid-write is worse than a refusal. A cheaper bound from `dbstat` or `page_count - freelist_count` is not *certainly* safe, and the cost of being wrong is a half-written candidate on a full disk. Reserving less is not worth that.

**The rollback copy is not auto-deleted either.** It is the recovery route, and deciding what event proves it is no longer needed is a real design question — filed as #1827 rather than answered in a paragraph here.

What this PR changes is that both costs are **stated instead of discovered**.

## The refusal

Before:

```
insufficient free space: need 36539352678 bytes, have 31627071488
```

That gives the operator two eleven-digit numbers to subtract and implies the whole requirement is real. Now:

```
insufficient free space: free 4912281190 more bytes to proceed (need 36539352678, have 31627071488)
	most of that is a worst-case reservation: VACUUM INTO cannot report the compacted size until it has written it, so the requirement assumes the result could be as large as the 33217593344-byte source. It is usually far smaller, but reserving less risks a half-written candidate on a full disk
	the space is not all returned when the run succeeds: a rollback copy of the source is kept as <db>.rollback-<run id>, and nothing removes it. Deleting it is your decision, and gives up "store compact rollback" for that run
```

The shortfall comes first because it is the only number the operator acts on. It is written across lines because it is read in a terminal, at the moment the store is already too large to compact.

## Documentation

New bilingual pair `docs/operations/store-compact-disk-cost{,.ja}.md`, linked from `docs/operations/README.md`, and referenced from `docs/storage/safe-compaction.md` where the compaction guarantee is described. It states the 1.1x requirement and why, the retained rollback copy and its name pattern, what deleting it gives up, and the observed figures as a worked example so the ratio is concrete.

## Scope

No change to `compactionRequiredBytes`, the margin, or any threshold. No new subcommand and no new flag. No change to the compaction run JSON. No automatic deletion.

## Verification

`go build ./...`, `go test -count=1 ./...`, `go tool golangci-lint run` (0 issues). `TestInsufficientCompactionSpaceErrorStatesOperatorCosts` pins the message exactly, including that the shortfall is the arithmetic difference and not a restatement of either input.

Closes #1790


duck8823:

**sol: no CRITICAL or HIGH findings.**

This PR changes no arithmetic and no threshold. Its entire risk surface is whether the words are true of the code, and whether they talk an operator into something harmful — so that is what the review checked, claim by claim.

### On the claims being true

sol verified each one against `compaction_files.go` and its callers: the worst-case reservation assumes the result could be as large as the source; the shortfall is the arithmetic difference; the rollback copy is named `<db>.rollback-<run id>`; nothing in the codebase removes it.

**No underflow.** `required - available` is `uint64`, and sol confirmed the subtraction is reachable only from inside the `required > available` branch. I checked the same call site: there is no other caller.

### On whether the message invites harm

This was the finding I most expected, and sol's answer is the useful form of "no": deleting the rollback copy **mid-run** would leave recovery without an orientation point and could wedge a resume or a rollback. That is a genuinely destructive action, and the question is only whether the text scopes it away from an operator who is in a hurry.

It does. The message says the space is not returned *when the run succeeds*, and the doc holds the copy until the compacted store has been *inspected and accepted*. Neither describes a window that is open during a run. I considered strengthening this to name the mid-run case explicitly and decided against it: the message is read at a refusal, before any run exists to be mid-way through, and adding a warning about a state the reader is not in costs attention that the shortfall needs.

- **The doc matches the code.** The 1.1x multiplier and the `<db>.rollback-<run id>` pattern both trace to source, not to the PR description.
- **EN and JA state the same facts.** A divergence in a bilingual pair is a real defect here, not a nit, since operators read one or the other and never both.
- **The exact-string test fails on a wrong shortfall**, not merely on a rewording — the expectation carries the computed difference, so an off-by-one in the subtraction breaks it.

### On the message shape

luna delivered the message as a single run-on sentence carrying four separate facts. It is read in a terminal at the moment the store is already too large to compact, so it is now written across lines with the shortfall first — it is the only number the operator acts on, and it was previously the one number the message did not contain.

### Verification

`go build ./...`, `go test -count=1 ./...`, `go tool golangci-lint run` (0 issues). sol independently re-ran `go test ./... -count=1` outside its sandbox after an initial sandbox-constrained failure; it passes.

Approved.

PR #1832, opened 2026-08-11
docs: correct what v0.34 says about the legacy search index

### LINKED ISSUE #1831: v0.34 release notes misdescribe the retired legacy search index, and omit the retirement entirely
Found while consulting on #1804. The v0.34 release notes describe the legacy search index wrongly in two ways, and both push the reader toward the same wrong conclusion: that they need do nothing.

## 1. The notes say search falls back to an index v0.34 retired

`CHANGELOG.md` / `CHANGELOG.ja.md`, the #1717 entry in `[Unreleased]`:

> Stores that have never rebuilt a generation keep using the legacy index silently, so no store has to rebuild to keep searching.

> a tail larger than the bounded candidate limit falls back to the legacy index instead of returning an incomplete page.

Neither is true in v0.34. `infrastructure/sqlite/event_search_authority.go:36-38`:

```go
// The tiered path is the only reader; the migration-032 family is never consulted.
```

A store without a complete generation still searches, and still searches **correctly** — `usableLiteralFingerprintGeneration` returns `""` and the tiered walk decides every candidate by decoding it. What it loses is the ability to skip decoding, so a broad query on a large store exhausts `DeepLiteralSearchBudget` and returns `EventSearchUnavailableIndexIncomplete` (`:172-174`). From the function's own comment (`:117-118`): *"a wide query on a large store exhausts the budget and reports index_incomplete"*.

The tail claim is obsolete for the same reason: `searchTieredTopKMetadataTx:146-151` now visits candidates in public order across the whole source sequence and retains only `offset+limit`, precisely so broad queries do not become unavailable. There is nothing to fall back to.

**This is not a wording nit.** "No store has to rebuild to keep searching" answers an operational question, and answers it wrongly. On a large store, rebuilding is exactly what keeps broad search available.

## 2. The retirement itself has no entry at all

`traceary store search-retire` is a v0.34 command that drops the largest object in the store — 16.15 GiB of a 24.7 GiB store on the maintainer's machine, about 65%. Migration `000052` stops maintaining the family at first startup. `traceary doctor` recommends the command (`presentation/cli/doctor_store_size.go:137-140`). `docs/operations/search-retirement.md` documents the whole procedure.

There is no `[Unreleased]` CHANGELOG entry for any of it (#1718).

The omission matters most where the release notes are least forgiving: the retirement is **roll-forward only**. Downgrading does not restore the index, and an older binary would query one frozen at the moment of upgrade and silently return incomplete results. A user who wants a backup first has to know before running the command, and the release notes are where they would look.

Together the two defects read as: "nothing changed about the legacy index, and you can keep relying on it." The truth is that it stopped being maintained, can be deleted for a large one-time saving, and cannot be brought back.

## Also stale

`infrastructure/sqlite/search_projection_catchup.go:26` carries the same retired claim in a comment:

> The legacy index remains authoritative until a generation reaches complete (#1717).

## Scope

One correction, one PR — the two halves are the same defect in the same files, and the corrected #1717 sentence has to point at the retirement entry that this also adds.

- Rewrite the #1717 fallback sentences (EN + JA) to state what actually happens on an un-rebuilt store, including `index_incomplete` on broad queries.
- Add the missing #1718 entry (EN + JA), covering the split between migration `000052` and the explicit `store search-retire` step, the required `search-retire` → `compact plan`/`apply` order, and the roll-forward-only warning.
- Fix the stale comment.

No code change. The behaviour is intended; only its description is wrong.

## Related

- #1717 — the entry being corrected.
- #1718 — the work with no entry.
- #1804 — where this was found; its harm assessment rests on the same fact.

duck8823:
## What was wrong

Two statements in the v0.34 release notes stopped being true inside this same unreleased cycle, and both told the reader that nothing about the legacy search index needed their attention.

### The #1717 entry described a fallback that no longer exists

> Stores that have never rebuilt a generation keep using the legacy index silently, so no store has to rebuild to keep searching.

> a tail larger than the bounded candidate limit falls back to the legacy index instead of returning an incomplete page.

`event_search_authority.go:36-38` — *"The tiered path is the only reader; the migration-032 family is never consulted."*

An un-rebuilt store does still search, and still searches **correctly**: `usableLiteralFingerprintGeneration` returns `""` and the walk decides each candidate by decoding it. What it loses is the ability to skip decoding, so a broad query on a large store exhausts `DeepLiteralSearchBudget` and returns `index_incomplete` (`:172-174`) — the function's own comment says so at `:117-118`. The tail claim is obsolete for the same reason: candidates are now visited in public order across the whole source sequence with only `offset+limit` retained (`:146-151`).

So the entry answered an operational question backwards. Rebuilding is precisely how a large store keeps broad search available.

### The retirement had no entry at all

`traceary store search-retire` drops the largest object in the store — 16.15 GiB of a 24.7 GiB store, about 65%. Migration `000052` stops maintaining the family at first startup. `doctor` recommends the command. `docs/operations/search-retirement.md` documents the procedure. None of it appeared in `[Unreleased]` (#1718).

That omission matters most where the notes are least forgiving. The retirement is **roll-forward only**: downgrading does not restore the index, and an older binary would query one frozen at the moment of upgrade and silently return incomplete results. A user who wants a backup first has to know before running the command.

## What this changes

- The #1717 fallback sentences (EN + JA) now state what an un-rebuilt store actually does, including `index_incomplete` on broad queries, and that there is no fallback because v0.34 retires the family.
- A new #1718 entry (EN + JA) covering why migration `000052` drops only constant-cost objects, why removing the tables is a separate explicit step, the fixed `search-retire` → `compact plan`/`apply` order, the idempotent `already_removed` exit, and the roll-forward-only warning with its backup advice.
- `search_projection_catchup.go:26` carried the same retired claim in a comment; it now says what failing open actually costs.

## On sourcing

The figures and mechanisms in the new entry are taken from source, not from the operations doc, so the notes cannot inherit a doc error:

- eight `DROP TRIGGER` statements in `000052`, plus the view and `search_maintenance_control` — counted in the migration file;
- `already_removed` / exit 0 — `application/types/legacy_search_retire.go:7`, `presentation/cli/store_search_retire.go:18`;
- the refusal to compact while the family is resident — `infrastructure/sqlite/compaction_sqlite.go:345`.

## Scope

One PR for both halves: they are the same defect in the same two files, and the corrected #1717 sentence has to point at the retirement entry this also adds. No code change — the behaviour is intended, only its description was wrong.

## Verification

`go build ./...`, `go test -count=1 ./infrastructure/sqlite/`, `go tool golangci-lint run` (0 issues).

Closes #1831


duck8823:
## Review record

**Claude Opus 5 (main reviewer)** and **gpt-5.6-sol (read-only consultation)**, on `fdb40fed` → `dec9e8bc`.

### gpt-5.6-sol HIGH — the rollback warning argued for the more dangerous state ✅ fixed

The original text said a downgraded binary would query an index frozen at the upgrade and silently return incomplete results. That is true only *while the migration-032 family is still resident* — and the sentence sat under a warning about deleting it, so it read as an argument for keeping it. It argued for the worse of the two states.

Verified against the tag rather than from memory (`git show v0.33.1:infrastructure/sqlite/event_search_query.go`): `eventSearchSchemaAvailable` returns true only when `count == 3` across `event_search_documents` / `event_search_fts` / `event_search_backfill_state`. So while all three are resident, v0.33.1 trusts an index that stopped receiving writes the moment v0.34 was opened — silently incomplete. Once they are gone it detects their absence and falls back to the literal scan, which is correct and merely refuses queries too broad to scan.

Rewritten in both `CHANGELOG.md` and `CHANGELOG.ja.md`: removing the family is what makes a downgrade honest, not what breaks it. The backup advice stands on its own — a backup is how you get the removed data back, not how you avoid a wrong answer.

### Two errors I caught in my own text ✅ fixed

- **Both cross-references pointed the wrong way.** They were written by direction ("see the entry below", "the compressed-payload entry above"), and #1718 sits at line 13, #1717 at line 20, #1618 at line 18 — so both were inverted. They now cite issue numbers, which reordering cannot invalidate.
- **A referenced heading did not exist.** The text pointed at a "Retire the legacy search index" entry that had never been written; #1718 had no changelog entry at all. Caught by grepping for the heading. Adding that entry is now part of this PR, which is what made #1831 two halves rather than one.

### Verification

`Docs` / `Lint` / `Test` / `Integrations` all green on `dec9e8bc`. The stale comment correction in `search_projection_catchup.go` is comment-only; no behaviour changes in this PR.

Issue #1833, opened 2026-08-11
v0.36-13: separate lock-wait from row work so a deterministically slow row can be skipped safely


duck8823:
Deferred from #1804. v0.34 makes the stall legible; it does not fix it. This is the fix.

## The problem, restated

`ResumeUntil` halves the row budget `128 → 64 → … → 1` on a write-lock duration cap and gives up at the one-row floor (`application/usecase/search_projection_usecase.go:162-200`). Nothing is persisted, so the next store open repeats the whole sequence from the same checkpoint. For a row whose own indexing work exceeds the cap every time, the projection never advances — and until it does, broad queries on a large store report `index_incomplete` (`infrastructure/sqlite/event_search_authority.go:172-174`) instead of returning results.

## Why neither cheap option works

Both were evaluated during #1804 and rejected on evidence.

**Persistence** — park or exclude after N consecutive single-row cap failures at the same checkpoint — cannot tell the two causes apart. The checkpoint advances in the same transaction as the work (`infrastructure/sqlite/search_projection_rebuild.go:977-985`), so a store under sustained writer contention sits at the same checkpoint for exactly the same reason a slow row does. Excluding on that evidence would permanently drop good events from the search index on a busy store, which is a silent false negative — no better than the stall it replaces, and harder to notice.

**Reading an existing signal** does not work either. `applyProjectionPlanWithRetry` (`infrastructure/sqlite/search_projection_rebuild.go:732-753`) does distinguish `isSearchProjectionSQLiteBusy` from `isSearchProjectionDeadline`, which looks like it should separate contention from row cost. It does not, because `database.go:410-421` sets `busy_timeout(1000)` — four times the 250 ms lock cap. SQLite blocks inside the driver waiting for the lock and the context deadline fires before SQLITE_BUSY is ever returned. Under contention the observable signature is identical to a slow row.

## What to build

Separate lock **acquisition** from lock **hold** and time them independently:

- Take the write lock explicitly (`BEGIN IMMEDIATE`) with its own budget, distinct from the row-work budget.
- Time acquisition and hold separately. Only elapsed time under a *held* lock is the row's own cost.
- A row that exceeds the cap on its own work is then identifiable on a single observation, with no counter and no cross-open state — which is what makes it safe to act on.

[...]

PR #1834, opened 2026-08-11
fix: name the checkpoint a stalled search projection cannot pass

### LINKED ISSUE #1804: v0.34-follow-up: a deterministically slow row makes the catch-up retry its whole shrink sequence on every store open
Raised by gpt-5.6-sol while reviewing #1803, and deliberately left out of it.

#1803 made the automatic catch-up shrink its batch on a write-lock duration cap, and deliberately does **not** park the generation when even a single row exceeds the cap. The reasoning: a lock-duration cap can be produced by another process holding the write lock, and permanently disabling the projection over a moment of contention would be worse than retrying on the next open.

That reasoning holds for contention. It does not hold for a row whose indexing work simply takes longer than the cap every time.

## The case

The typed error does not distinguish "lost a contention race" from "this row's own work exceeds 250 ms". For a row that is admissible by every byte budget — up to the 8 MiB logical limits — but consistently too slow to index inside the lock window, every store open repeats the full `128 → 64 → … → 1` sequence, fails, and advances nothing. The checkpoint never moves, so the next open starts from the same row.

With the production budget that is up to eight attempts per open, indefinitely, on a path that runs on every write-capable command. Nothing parks, nothing escalates, and the only signal is a repeated ERROR log.

## Why it was not fixed in #1803

Parking on the first single-row cap failure would reintroduce exactly the fragility the non-parking decision avoids: one contended moment permanently disables the projection. Distinguishing the two cases needs either state across opens (a consecutive-failure counter on the generation) or a measurement that separates lock-wait time from work time. Both are design decisions, not a tweak to a retry loop.

## What to decide

- Whether the distinction should come from **persistence** (park after N consecutive single-row cap failures at the same checkpoint, which is what makes it deterministic rather than transient) or from **measurement** (attribute the elapsed time between waiting for the lock and holding it, and treat only the latter as the row's own cost).
- What a parked generation should say to the operator, given the same row will reappear on any replacement generation. This overlaps #1794: both end at "one row cannot be projected under the current budget, and the operator needs to be told which row and why".
- Whether the two should share one mechanism. #1794's design skips an oversized row, records the exclusion and advances past it. A deterministically slow row wants the same outcome — skip, record, advance — reached by a different test.

## Related

- #1803 — the non-parking decision this qualifies.
- #1794 — the oversized-row case, which wants the same skip-and-record outcome.
- #1620 — the release gate.

duck8823:
Closes #1804

## What this changes

When `ResumeUntil` reaches the one-row floor it returns `SearchProjectionNoProgressSingleRowLockDurationCap`. That error reached `logSearchProjectionCatchUp`, which logged the generic line:

> `search projection catch-up incomplete; retrying on next initialization`

That is the same line an ordinary bounded batch produces when it stops short, so the one condition that can keep the projection from ever advancing was indistinguishable from routine progress. While it holds, broad searches on a large store report `index_incomplete` (`event_search_authority.go:172-174`) rather than returning results, and nothing on screen says why.

That code now gets its own ERROR, naming the phase, the source checkpoint where the walk stopped, and the recovery: `store search-projection resume --until-complete --lock-time <larger>`, which the existing generation accepts because `LockTime` is deliberately outside its config hash (`application/types/search_projection.go:85-96`) — no rebuild, nothing durable discarded. Reporting the checkpoint meant carrying it on `SearchProjectionCatchUpResult`.

The message does not claim the condition is permanent. A writer holding the lock past the cap is retried inside the cap and converted to the same lock-duration error (`search_projection_rebuild.go:733-764`), so at the one-row floor a transient stall is indistinguishable from a deterministic one and the next open may well succeed. Predicting recurrence here would be the #1795 defect in mirror image.

## What this deliberately does not change

Nothing is skipped and no new state is persisted. #1804 offers two routes to *fix* the stall; both were rejected for this release on evidence:

- **Persistence** (park/exclude after N consecutive failures at the same checkpoint) cannot discriminate. The checkpoint advances in the same transaction as the work (`search_projection_rebuild.go:977-985`), so a store under sustained writer contention sits at the same checkpoint for exactly the reason a slow row does. Excluding on that evidence would permanently drop good events from the search index on a busy store — a silent false negative, no better than the stall it replaces and harder to notice.
- **Measurement** cannot read an existing signal either. `applyProjectionPlanWithRetry` does distinguish busy from deadline, but `database.go:410-421` sets `busy_timeout(1000)` — four times the 250 ms lock cap — so SQLite blocks inside the driver and the context deadline fires before `SQLITE_BUSY` ever surfaces. Separating the two needs a restructured `BEGIN IMMEDIATE` timed apart from the held-lock work, which is a change to lock acquisition on the path that runs on every write-capable store open. Not an RC change.

Filed as #1833.

## Review findings fixed in this PR

Four, all of the class this change exists to remove — a surface asserting something the code cannot establish. Full record in the review comment below.

- **The recovery was wrong and destructive** (sol, HIGH). It said to abort and start a new generation, on my premise that `resume` refuses a differing budget. `LockTime` is explicitly excluded from `ConfigHash`, so `resume --lock-time <larger>` works on the same generation; aborting would have discarded every durable batch and forced a full rebuild to raise a lock cap.
- **The permanence claim is not established** (sol, HIGH) — see above.
- **The checkpoint is not the stop position in every phase** (sol, HIGH). Inventory advances a string cursor and cleanup consumes its own scope; neither moves the source checkpoint. It is now reported only in the source phase, with `phase` always present.
- **The reported position could be stale** (mine, then sol MEDIUM for a second cause). Catch-up may replace the generation before it resumes, and `StartGeneration` writes `checkpoint=0`; separately, a concurrent opener may commit between the entry read and the failure. The position is now re-read when the failure is reported, covering both.

## Verification

```
go build ./...                 ok
go tool golangci-lint run      0 issues.
go test -count=1 ./...         ok (all packages)
```

Falsification, both re-run independently rather than taken from the delegated report:

- Removing the new `errors.As` branch → the log test fails on the message (`committed no row at the minimum batch size…` → `incomplete; retrying on next initialization`).
- Removing the source-phase guard on the checkpoint attribute → the inventory case fails (`+int64(1842)` where nothing should be reported).
- Removing the failure-time position re-read → the replaced-generation case fails `-0 +1842` and the concurrent-opener case fails `-1900 +1842`.

`select_search_projection_control_status.sql` gains a column mid-`SELECT`; it has exactly one `Scan` site (`search_projection_status.go:45`, updated) and one `EXPLAIN QUERY PLAN` test that asserts referenced objects, not column positions.


duck8823:
## Review record

**Claude Opus 5 (main reviewer)**, **gpt-5.6-luna (implementation)**, **gpt-5.6-sol (read-only adversarial review)**, through `a5a9d782`.

### Delegated verification, re-run rather than believed

`go build ./...`, `go test -count=1 ./...` (all packages), `go tool golangci-lint run` (0 issues) — run independently after each round. Both falsifications re-done rather than taken from the report: removing the `errors.As` branch flips the message back to the generic line, and removing each guard fails its own case (see below).

### Found in my own review, before sol

Catch-up may replace the generation before it resumes (`idle` / `drifted`, or an obsolete capacity-semantics version), and `StartGeneration` writes `checkpoint=0`. The checkpoint read at entry therefore belongs to the generation that was just replaced, and the error path returned it unchanged — the new message would have named a row the new generation never reached. Superseded by the broader fix below.

I also checked what the shared `select_search_projection_control_status.sql` change could break: one `Scan` site (updated) and one `EXPLAIN QUERY PLAN` test that asserts referenced objects rather than column positions. sol independently found no other reader, and no frozen JSON contract carrying `SearchProjectionCatchUpResult`.

### gpt-5.6-sol HIGH — the recovery instruction was wrong, and destructive ✅ fixed

The message said to abort the generation and start a new one with a larger lock duration, on my premise that `resume` refuses a differing budget. **That premise is false**, and the source says so explicitly (`application/types/search_projection.go:85-96`):

> Rows, WallTime and LockTime only bound the cost of one batch, so they stay out of the generation identity: a batch that cannot fit its lock cap must be allowed to resume with a smaller unit of work rather than discard durable progress.

So `store search-projection resume --until-complete --lock-time <larger>` is accepted by the *existing* generation. Aborting would have discarded every durable batch and forced a full rebuild — on the store where this error fires, hours of work and a rebuild peak, to raise a lock cap. This was my error in the task packet, propagated into the code and the changelog.

### gpt-5.6-sol HIGH — "re-running will not change this on its own" is not established ✅ fixed

A writer holding the lock past the cap is retried inside the cap and converted to the same lock-duration error (`search_projection_rebuild.go:733-764`); at the one-row floor it becomes this code. The next open can then succeed with no operator action.

This is precisely the defect #1795 was fixed for — predicting recurrence the code cannot establish — reintroduced two PRs later in the mirror image. The message now reports what happened (the shrink loop reached one row and it did not commit; search stays incomplete until it advances) and leaves recurrence to the operator's own observation.

### gpt-5.6-sol HIGH — the checkpoint is not the stop position in every phase ✅ fixed

Inventory advances a string cursor (`ApplyInventoryBatch` returns the same lock-cap code) and cleanup consumes its own scope; neither moves the source checkpoint. Printing it there names a position the work never used — a wrong number reported as exact. The attribute is now emitted only in the source phase, with `phase` always present. Falsified: dropping the guard fails the inventory case.

Exposing the inventory cursor as a stop position is out of scope here and noted on #1833.

### gpt-5.6-sol MEDIUM — a concurrent opener's committed batch ✅ fixed

`MaxBatches: 1` only rules out this call committing before it fails; another opener can commit between the entry read and the failure. The position is now re-read when the failure is reported, which also subsumes the replaced-generation case I had fixed narrowly. Falsified: removing the re-read fails both the replaced-generation case (`-0 +1842`) and the concurrent case (`-1900 +1842`).

### Not changed

No exclusion, no persisted counter, no new command or flag. Both routes to actually fixing the stall were rejected on evidence and deferred to #1833; the PR body carries the reasoning.

Issue #1839, opened 2026-08-11
v0.37-17: search projection status mixes generations across its statements


duck8823:
## What

`SearchProjectionStatus` (`infrastructure/sqlite/search_projection_rebuild.go`) builds one status object out of several independent statements, and each statement that needs the published generation re-resolves `active_generation_id` on its own, outside any transaction. A cutover committing between two of those statements produces a status object whose fields describe two different generations.

This is pre-existing. The counters already affected are `recent_documents`, `summary_sessions`, `keyword_rows` and `fingerprint_rows`. #1838 added `recent_oldest_norm` / `recent_newest_norm`, which resolve the generation inside their own single statement — so the range and its own readiness check are consistent with each other, but not necessarily with the counters beside them.

Found by gpt-5.6-sol reviewing #1838; deliberately not fixed there because the fix is a change to how the whole status object is read, not to the field being added.

## Why it matters

The failure is a **wrong number reported as exact**. An operator reading `status` during a cutover can see, for example, `recent_documents` from the newly published generation next to a range from the one it replaced, with nothing in the output marking the mixture. Every value is individually true and the object as a whole is not.

The window is narrow — a cutover is a single commit and `status` is otherwise slow for unrelated reasons — but it is not zero, and automatic catch-up on store open means cutovers happen without an operator doing anything.

## Sketch

Read the whole status object inside one read transaction (`BEGIN DEFERRED`), or resolve `active_generation_id` once and pass it to every statement that needs it. The second is the smaller change and makes the scoping explicit at each call site; the first also protects against rows changing under statements that do not name a generation at all.

Note that `status` also walks `dbstat` for capacity evidence, measured at 184 s on a 408,893-event store. Holding a read transaction across that walk is not obviously acceptable — whichever approach is taken, the generation-scoped reads and the diagnostic measurement probably should not share one transaction scope.

## Acceptance

- [ ] Every generation-scoped field in one `SearchProjectionStatus` result describes the same generation, and a test proves it by committing a cutover between two of the reads
- [ ] The `dbstat` walk is not held inside whatever scope is introduced, or the cost of holding it is measured and stated
- [ ] The stale reference in the `selectSearchProjectionRecentRangeSQL` comment ("the rest of the status object still re-resolves the generation per statement (#1839)") is updated

Follow-up from v0.34.

duck8823:
v0.34 release triage: **deferred to v0.35, not a release blocker.**

[...]

PR #1840, opened 2026-08-11
feat: report the range full-text search actually covers

### LINKED ISSUE #1838: v0.34.0-12: report the recent tier's actual searchable range
#1621 carries this acceptance criterion for v0.34:

> - [ ] The searchable range is reportable: the actual oldest and newest indexed timestamps

It is not met. `store search-projection status` reports 30-plus fields and none of them answers it. `recent_cutoff_norm` is the *source-phase prefilter* boundary derived at `Start`, not the range actually retained: eviction runs after it and drops the oldest documents, so the real oldest is newer than the cutoff by an amount nothing reports. On the store copies measured for #1620 the field is empty, which tells an operator nothing at all.

What that costs: `traceary search` answers recent history from the full-text tier and older history from session summaries and keywords, and the boundary between them moves with the corpus — the derived ceiling shrinks as the permanent tiers grow, so a window that was three weeks becomes one. An operator who searches for a phrase they remember typing, and gets a session-tier hit or nothing, has no way to tell whether the phrase is absent or merely older than the full-text window.

## Scope

Report the recent tier's actual retained range on `store search-projection status`, for the generation the status describes:

- `recent_oldest_norm` / `recent_newest_norm`, from `MIN` / `MAX` of `search_projection_recent_documents.created_at_norm` scoped to `generation_id`.
- Both resolve through the existing `idx_search_projection_recent_eviction(generation_id, created_at_norm, event_rowid)`, so this is an index seek, not a scan — it must not add a measurable cost to `status`.
- Omit both when the recent tier is empty. An empty tier is a real state (derived ceiling 0) and must not be reported as a zero-width range at some default timestamp.
- Additive JSON fields only. No new command, no new flag, no new subcommand.

The fields describe the **full-text** range. Older history remains answerable through the session tier, and the field names and the documentation must not imply that events outside this range are unsearchable.

## Note for the implementer

`created_at_norm` is compared lexically here, which is the same ordering `idx_search_projection_recent_eviction` and the eviction predicate already rely on. If #1185's RFC3339Nano lexical hazard applies, it applies to retention itself and not merely to this report — do not fix it locally in a way that makes the reported range disagree with what eviction actually did. Report what eviction's own ordering says.

## Acceptance

- `status` on a store with a non-empty recent tier reports both timestamps, and they match a direct `MIN`/`MAX` query against the table.
- `status` on a store with an empty recent tier omits them rather than reporting a placeholder.
- A test that fails if the range is computed across generations instead of scoped to the one `status` describes.
- `docs/search-projection-rebuild.md` and its Japanese pair state what the range covers and what falls outside it.

duck8823:
Closes #1838

## What

`store search-projection status` reports thirty-odd fields and, before this, none of them answered the one question an operator asks of a windowed index: *what period can `traceary search` actually find things in?* The closest field, `recent_cutoff_norm`, is the source-phase prefilter boundary derived at `Start` — eviction runs afterwards and moves the real boundary — and on the maintainer's store it is empty anyway.

Two additive JSON fields, `recent_oldest_norm` and `recent_newest_norm`, report `MIN`/`MAX` of `created_at_norm` over the documents actually retained in the recent full-text tier.

## The condition they carry

The fields report the range **search reads**, not the range that exists. The statement repeats `searchProjectionReadReady`'s condition verbatim — `state='complete'` with a non-empty `active_generation_id` — so the range is omitted whenever search is not reading the projection at all:

| state | rows present | reported |
|---|---|---|
| complete, active generation | yes | the active generation's range |
| rebuilding | yes, from the previous generation *and* the one being built | omitted |
| drifted (`active_generation_id` cleared) | yes | omitted |
| complete, tier empty | no | omitted |

Two ways of getting this wrong were caught during review, both of which would have made the field claim coverage that is not there:

1. The original implementation scoped to `generation_id` — the generation being *built*. During a rebuild that reports a wider range than search can answer from, in the same object whose `recent_documents` counts the active one.
2. Scoping to `active_generation_id` alone is still not enough. `searchProjectionReadReady` (`event_search_query.go:49`) also requires `state='complete'`, so during a rebuild the previous generation's rows are present and named but never consulted — broad queries decode every candidate and can return `index_incomplete`. Reporting their range would advertise full-text coverage for precisely the period it is absent. (gpt-5.6-sol)

## Verification

- `EXPLAIN QUERY PLAN` resolves through the covering index `idx_search_projection_recent_eviction (generation_id=?)`: no scan, no temp b-tree.
- Falsification, run by me rather than reported: dropping `AND state='complete'` from the statement fails `omits_the_range_while_a_rebuild_is_in_flight`; scoping back to `generation_id` fails the read-ready case with `-2026-08-01T00:00:00Z +2020-01-01T00:00:00Z`.
- `go build ./...`, `go tool golangci-lint run` (0 issues), `go test -count=1 ./...` — all pass.

## Not fixed here

`SearchProjectionStatus` re-resolves `active_generation_id` once per statement, outside a transaction, so a cutover landing mid-read mixes generations across the object. Pre-existing for `recent_documents`, `summary_sessions`, `keyword_rows` and `fingerprint_rows`; filed as #1839 rather than fixed here, because it is a change to how the whole object is read.

## Attribution

Implementation by gpt-5.6-luna from a packet; scoping fix, readiness gate, test restructure and docs by Claude Opus 5. Reviewed by gpt-5.6-sol (one HIGH, fixed; one HIGH, filed as #1839).


duck8823:
Closing unmerged: the premise is void — nothing reads the recent full-text tier. See the explanation on #1838, and #1841 / #1842 for the work this splits into.

PR #1846, opened 2026-08-12
v0.34.0-57: recover stale WAL sidecars instead of deadlocking store compact

### LINKED ISSUE #1845: v0.34.0-57: a stale zero-byte WAL sidecar deadlocks store compact
## Problem

`store compact` — a headline v0.34 feature — deterministically deadlocks its own documented operator sequence. Once the store carries a stale zero-byte `-wal`/`-shm` sidecar, every compaction command fails with the same error, and no retry can change that.

Found while cutting the v0.34.0 RC (#1621). **This blocks the tag.**

## Reproduction (deterministic, RC binary at `e2daed60`)

```
$ traceary log --kind note "seed"                 # clean: no -wal
$ traceary store search-retire                    # clean
$ traceary store compact plan                     # succeeds, clean
$ traceary store search-projection status         # -wal and -shm now present, 0 bytes
$ traceary store compact plan
Error: failed to execute CLI command: SQLite sidecar exists: /path/store.db-wal
$ traceary store compact plan                     # identical
$ traceary store compact apply RUN_ID             # identical
$ traceary store compact resume RUN_ID            # identical
```

This is the documented sequence in `docs/operations/search-retirement.md`:

```sh
traceary store search-retire
traceary store compact plan
traceary store compact apply RUN_ID
```

Checking projection status between `plan` and `apply` — a natural thing to do, since projection state is usually what the operator is deciding about — puts the run into a state where every compaction command refuses.

Measured, not assumed: with no prior `compact plan`, `store capacity`, `store gc --dry-run` and `store search-projection status` each leave a freshly-seeded store clean. The sidecar appears when `search-projection status` runs *after* a `compact plan`. Why the prior plan changes the outcome is not yet root-caused — see task 5.

## Why it is a permanent stall, not an inconvenience

Every documented compaction transition repeats identically, and retry cannot change the blocking state. Recovery requires an undisclosed, unrelated mutation: running any *writing* command (e.g. `traceary log ...`) clears the sidecars, after which `compact plan` succeeds immediately. The error message names no remedy, and neither `docs/storage/safe-compaction.md` nor `docs/operations/search-retirement.md` mentions it — and an operator running maintenance is typically not writing to the store. A recoverable filesystem state is not a self-progressing workflow.

## Current behaviour

`rejectSQLiteSidecars` (`infrastructure/sqlite/compaction_files.go:940`) `os.Lstat`s `-wal` / `-shm` / `-journal` and fails on mere existence, inspecting neither type nor size. It gates `Plan` (line 375), two further preflights (473, 487), the exchange path (738) and the prepared-migration recipe (`prepared_migration_recipe.go:217`). Its rationale is the comment at line 381: *"After the sidecar rejection the store has no live WAL/SHM, so it is safe to open."*

`search-projection status` reads through `openReadOnly`, whose own contract already says WAL reads may create sidecars (`database.go:269-278`), and it does close its connection (`search_projection_rebuild.go:1084-1090`).

## Fix

Reviewed with gpt-5.6-sol. **Do not** weaken `rejectSQLiteSidecars` globally, and do not simply accept zero-byte files: `Plan` does not currently hold the exclusive lease, and zero length alone is not proof against a creation race.

1. **`Plan` acquires the exclusive store lease**, as `Apply` / `Resume` already do (`store_compaction.go:51-64,157-172`).
2. **While exclusively leased**, treat only *regular, zero-byte* `-wal` and `-shm` files as stale artifacts: remove them, sync the directory, then re-run the strict absence check before opening, digesting or replacing anything.
3. **Keep rejecting**: any non-zero `-wal`/`-shm`; every `-journal` regardless of size; symlinks and non-regular files; any stat or cleanup error.
4. **The remaining refusal must name a remedy** — identify the sidecar's type and size and tell the operator to stop other Traceary processes and retry — and the recovery path goes in `docs/storage/safe-compaction.md` (+ `.ja.md`).

The lease argument holds only because every Traceary physical SQLite connection takes a shared lease (`store_lease.go:71-115`). It does not make the store safe against non-cooperating SQLite clients; the replacement protocol already depends on that same coordination boundary, so this adds no new assumption.

Rejected alternative: making `search-projection status` avoid or checkpoint the WAL. Opening it immutable risks ignoring committed WAL state — a silent false negative in an observational command — and checkpointing puts a mutation inside a read-only surface.

## Acceptance criteria

- [ ] The reproduction above completes: `plan` -> `search-projection status` -> `apply RUN_ID` commits, with no manual intervention and no writing command in between
- [ ] `Plan` holds the exclusive store lease for the whole preflight, including sidecar cleanup and the re-check
- [ ] A non-zero `-wal` is still rejected; prove it with a committed, non-empty WAL
- [ ] A `-journal` of any size is still rejected, including zero-byte
- [ ] Symlinked, FIFO and other non-regular sidecar paths are still rejected
- [ ] A stat or unlink error fails closed, never silently proceeds
- [ ] With a coordinated reader and a coordinated writer connection held open, `plan` / `apply` wait or fail *before* any cleanup — never remove a sidecar out from under a live opener
- [ ] The refusal message names the sidecar type and size and tells the operator what to do
- [ ] `docs/storage/safe-compaction.md` and `.ja.md` document the recovery path
- [ ] After the fix, `apply` and `resume` both reach `committed` from the previously-stuck state, with `PRAGMA integrity_check = ok` and an unchanged logical digest (events / sessions / command_audits counts and total body length)
- [ ] Table-driven subtests cover zero-byte vs non-zero `-wal`/`-shm`, all journal forms, non-regular paths, and the leased-cleanup race
- [ ] `go test ./...` and `go tool golangci-lint run` pass

## Follow-up, not in scope here

Root-cause why a prior `compact plan` changes whether `search-projection status` leaves a sidecar — trace `PreparedStoreUpgradeFiles.Plan`, `RejectRetiredSearchIndex`, `openDirectReadOnly` and `Database.openReadOnly`. The fix above makes the stall unreachable either way; the mechanism is still worth knowing. File separately if it is not answered while implementing.

Parent: #1606

duck8823:
Closes #1845.

`store compact` deadlocked its own documented operator sequence. After `plan`, running `store search-projection status` left a zero-byte `-wal` and a 32 KB `-shm` behind, and every subsequent compaction command refused on their mere existence — identically, forever. The only escape was an undocumented, unrelated mutation: any writing command cleared the sidecars. This is a v0.34.0 release blocker, found while cutting the RC.

## What changed

`plan` now takes the exclusive store lease for its whole preflight, as `apply` and `resume` already did. While it is held, a stale WAL artifact is recovered rather than refused:

- `-journal` present in any form -> refuse, unchanged.
- Any sidecar that is not a regular file (symlink, FIFO, directory) -> refuse, unchanged.
- `-wal` present and non-zero -> refuse, unchanged. This is the case that may hold committed frames.
- Otherwise remove both `-wal` and `-shm`, whatever the `-shm`'s size is, sync the directory, and re-run the strict check.

`Recheck` — the apply/resume preflight — does the same, because the sidecar in the reproduction appears *after* `plan` returns. `RecheckForPublish` deliberately stays strict: cleanup has already happened under the lease, so a sidecar appearing there means an opener arrived mid-run and the run must abort. Candidate-path guards stay strict too; a sidecar beside a file we built ourselves is a surprise, not an artifact.

The refusal message now names the sidecar's type and size and tells the operator what to do, instead of `SQLite sidecar exists: <path>`.

### Recovery is gated on the lease, not assumed

`PreparedStoreUpgradeFiles` is shared by two use cases, and only `storeCompactionUsecase` holds the exclusive lease across `Plan`/`Recheck`. `preparedStoreUpgradeUsecase` — the `store payload-rehearsal` path — acquires it only at the publish steps. So the adapter now carries an explicit `CallerHoldsExclusiveLease`, set only where the compaction use case is wired. With it, sidecars are recovered; without it, they are refused exactly as before this PR. `lease_capability` is derived from the same field rather than hardcoded, so the `!LeaseCapability` guard stays meaningful for the unleased caller and for a run persisted by an older binary.

### Why the `-shm` size is not a criterion

The first attempt required both sidecars to be zero-byte and could therefore never fire on the reproduction it was written for: a cleanly-closed reader always leaves a 32,768-byte `-shm`. Per the [SQLite WAL format specification](https://www.sqlite.org/walformat.html), the shm "does not contain any database content and is not required to recover the database following a crash", is "never fsync()-ed to disk", and is normally truncated by the first client to connect to a quiescent database; and when the WAL's `mxFrame` is zero "all content should be obtained directly from the database file". The `-wal` carries the recoverable state, so its length is the evidence and the `-shm`'s is not.

### Why this cannot race a live opener

Cleanup runs only once the exclusive lease is held, and every live cooperating connection holds the shared form of the same lease, so the acquisition waits rather than deleting anything. Non-cooperating SQLite clients remain outside this advisory-lock boundary — the same boundary the replacement protocol already depends on, so no new assumption.

## Verification

Measured on binaries built from this branch, not asserted.

**The reported reproduction, no manual intervention:**

```
after log                        clean
after search-retire              clean
after compact plan               clean
after search-projection status   -wal 0 bytes, -shm 32,768 bytes
apply RUN_ID                     phase: committed
PRAGMA integrity_check           ok
```

with the logical digest (events / sessions / command_audits counts and total body length) unchanged. Also run through `resume` instead of `apply`.

**Refusals still hold** (each measured, each a separate scratch store):

| case | result |
|---|---|
| `-wal` of 819,912 bytes, live writer connected | refused, size reported |
| zero-byte `-journal` | refused |
| `-wal` symlinked to `/etc/hosts` | refused, type reported |
| zero-byte `-wal` + 32 KB `-shm` on the unleased rehearsal path | refused, both files untouched |

`store payload-rehearsal preview` and `run` still succeed on a clean copy.

**Lease exclusion, measured:**

- With another process holding the **exclusive** lease, `plan` waited 3.61 s and then succeeded.
- With another process holding the **shared** lease for 8 s, `plan` blocked 7.15 s — its full remaining lifetime — and removed the `-wal` only after it released. It did not delete a sidecar out from under the holder.

`go build ./...`, `go test ./...`, `go tool golangci-lint run`, `go run ./cmd/repo-tooling docs verify-i18n` all pass.

## Review

Implementation by gpt-5.6-luna over four rounds; review and the final verification by Claude Opus 5. Round 1 cleaned up in `plan` only, so the reproduction still failed at `apply` — caught by running it. Round 2 added `Recheck` but kept the zero-byte rule for `-shm`, which cannot fire on the real reproduction — caught by luna and corrected against the SQLite specification. Round 4 fixed a HIGH from the gpt-5.6-sol review: the cleanup was reachable from the unleased prepared-upgrade path, where its safety argument does not hold. gpt-5.6-sol also reviewed the blocker-vs-defer call and the shape of the fix before any code was written.

### DIFF
--- a/application/usecase/store_compaction.go
+++ b/application/usecase/store_compaction.go
@@ -31,6 +31,11 @@ func (u *storeCompactionUsecase) Plan(ctx context.Context, source string) (domai
 	if filepath.Clean(source) != u.expectedStore {
 		return domain.CompactionRun{}, fmt.Errorf("compaction store binding mismatch")
 	}
+	release, err := u.lease.AcquireExclusive(ctx, u.expectedStore)
+	if err != nil {
+		return domain.CompactionRun{}, err
+	}
+	defer release()
 	idBytes := make([]byte, 16)
 	if _, err := rand.Read(idBytes); err != nil {
 		return domain.CompactionRun{}, fmt.Errorf("create compaction run id: %w", err)
--- a/docs/storage/safe-compaction.ja.md
+++ b/docs/storage/safe-compaction.ja.md
@@ -39,8 +39,10 @@ advisory lockを保持します。`apply`、`resume`、`rollback` はjournalや�
 排他を継続します。取得はcontext cancellationに従い、プロセス終了時はOSが
 lockを解放します。lock file自体は意図的に残します。既存databaseと親directoryの
 symlinkは同じlease namespaceへ解決し、安全にfenceできないhardlink databaseは
-拒否します。非対応platformまたは
-probe失敗時は `false` としてfail closedします。旧版や非協調processは引き続き
+拒否します。`plan`の実行前にはlease取得が必須なので、非対応platformでは
+planがlease取得時点で失敗します。したがって永続化されたrunの
+`lease_capability`は常に`true`です。このfieldは後続のcapability probeではなく、
+planがlease前提を満たしたことを記録します。旧版や非協調processは引き続き
 事前停止が必要です。
 filesystem安全性は協調モデルです。参加するlive openerはすべて隣接leaseを使い、
 破壊的境界ではsource、candidate、rollbackのhardlinkを拒否します。権限を持つ
--- a/docs/storage/safe-compaction.ja.md
+++ b/docs/storage/safe-compaction.ja.md
@@ -17,6 +17,14 @@ source store size の 1.1 倍の空き容量を要求します。成功後も so
 大きさの元 database rollback copy が残ります。運用者向けの詳細は
 [`store compact` のディスク容量](../operations/store-compact-disk-cost.ja.md) を参照してください。
 
+`plan`はpreflight全体でstoreのexclusive leaseを保持します。regularかつ0 byteの
+`-wal`または`-shm`だけをstale artifactとして削除し、directoryをsyncしてから、
+storeをopenする前に再検査します。non-zeroのWAL/SHM、サイズにかかわらずすべての
+`-journal`、symlink、FIFO、その他のnon-regular pathは削除しません。sidecarが残って
+いる場合は、すべてのTraceary process（projection/status reader、旧版、非協調版を
+含む）を停止して同じcommandをretryしてください。sidecarを手動削除してはいけません。
+non-zeroまたはnon-regularのsidecarにはliveなSQLite stateが含まれる可能性があります。
+
 レガシー検索インデックスの検査はsource digestより前に走るため、数GiBのstore
 全体をハッシュした後ではなく数秒で失敗します。先にcompactすると死んだindexを
 新しいファイルへコピーして固定化してしまうので、`traceary store search-retire`
--- a/docs/storage/safe-compaction.ja.md
+++ b/docs/storage/safe-compaction.ja.md
@@ -17,13 +17,16 @@ source store size の 1.1 倍の空き容量を要求します。成功後も so
 大きさの元 database rollback copy が残ります。運用者向けの詳細は
 [`store compact` のディスク容量](../operations/store-compact-disk-cost.ja.md) を参照してください。
 
-`plan`はpreflight全体でstoreのexclusive leaseを保持します。regularかつ0 byteの
-`-wal`または`-shm`だけをstale artifactとして削除し、directoryをsyncしてから、
-storeをopenする前に再検査します。non-zeroのWAL/SHM、サイズにかかわらずすべての
-`-journal`、symlink、FIFO、その他のnon-regular pathは削除しません。sidecarが残って
-いる場合は、すべてのTraceary process（projection/status reader、旧版、非協調版を
-含む）を停止して同じcommandをretryしてください。sidecarを手動削除してはいけません。
-non-zeroまたはnon-regularのsidecarにはliveなSQLite stateが含まれる可能性があります。
+`plan`はpreflight全体でstoreのexclusive leaseを保持します。`-journal`がなく、
+non-zeroの`-wal`もない場合は、存在するregularな`-wal`と`-shm`を、`-shm`が
+non-zeroでも両方stale artifactとして削除します。directoryをsyncしてから、storeを
+openする前に再検査します。shm fileはdatabase contentを含まず、fsyncもされません。
+empty WALではすべてのcontentをmain database fileから取得します。non-zeroのWAL、
+サイズにかかわらずすべての`-journal`、symlink、FIFO、その他のnon-regular pathは
+削除しません。sidecarが残っている場合は、すべてのTraceary process（projection/status
+reader、旧版、非協調版を含む）を停止して同じcommandをretryしてください。sidecarを
+手動削除してはいけません。non-zero WALまたはnon-regular sidecarにはliveなSQLite
+stateが含まれる可能性があります。
 
 レガシー検索インデックスの検査はsource digestより前に走るため、数GiBのstore
 全体をハッシュした後ではなく数秒で失敗します。先にcompactすると死んだindexを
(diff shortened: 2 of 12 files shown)


duck8823:
## Review record

**gpt-5.6-sol (round 1, `5e39c6a7`)** — one HIGH:

> `PreparedStoreUpgradeFiles` is also called by `preparedStoreUpgradeUsecase` without acquiring an exclusive lease (`application/usecase/prepared_store_upgrade.go:54`, `:81`). Consequently, these paths can delete an arbitrary-size `-shm` or zero-length `-wal` while another connection still uses it. The literal `LeaseCapability: true` also silently asserts a lease that this caller never acquired, bypassing the guard.

Verified against the source before acting on it: `preparedStoreUpgradeUsecase` acquires the exclusive lease only at the publish steps (`:269`, `:317`), so `Plan` and `Recheck` do run unleased on that path. Fixed in `7bae6f83` by making the lease an explicit property of the adapter rather than an assumption of the code.

**gpt-5.6-sol (round 2, `7bae6f83`)** — no CRITICAL or HIGH. It was asked specifically to check that every call site setting the flag really holds the lease, that no other non-test constructor needs it, that the restored `probeStoreLeaseCapability` behaves as it did pre-PR at its new position, that the `!LeaseCapability` guard is still reachable for a v0.33 run resumed under this binary, and whether the new dispatch changes any outcome on the prepared-upgrade path.

**Claude Opus 5** — verification re-run on a binary built from `7bae6f83`, not inherited from the previous round: the reproduction reaches `committed` with `integrity_check = ok`; the non-zero `-wal`, `-journal`, and symlink refusals all still fire with type and size named; `store payload-rehearsal preview` and `run` succeed on a clean independent copy and refuse a copy carrying a zero-byte `-wal` plus a 32 KB `-shm`, leaving both files byte-for-byte intact. `go build`, `go test ./...`, `golangci-lint`, `docs verify-i18n` green locally; CI green on `7bae6f83`.

Issue #1847, opened 2026-08-12
v0.34 follow-up: a fresh store requires store search-retire before store compact


duck8823:
A brand-new v0.34 store still carries the legacy migration-032 search family, so a first-time user who runs `store compact` is refused and told to run `store search-retire` — a command whose documentation frames it entirely as an upgrade concern.

## What happens

On a store created by this version, with no prior data and no upgrade from any earlier release:

```
$ traceary log --kind note "seed"
$ traceary store compact plan
Error: ... legacy search index present; run `traceary store search-retire` first
```

`store search-retire` then succeeds and `compact plan` works. So the escape exists, but the operator has to find it, and the document they would reach for tells them it does not apply to them.

## Why

Migration 052 deliberately drops only the triggers, the view, and the control table. The three tables `event_search_documents`, `event_search_fts`, and `event_search_backfill_state` remain in the schema that a fresh store is built from, so the compaction preflight — which refuses to bake a dead index into a new file — sees them on day one.

That preflight behaviour is correct and should stay. The problem is that the schema a fresh store starts from contains a family that is dead the moment it is created.

## Options

1. Stop creating the family in a fresh store: have the migration chain leave no trace of it for a store that starts at the current head. This is the honest fix — a new store should not be born needing a retirement command.
2. Keep creating it, and document the fresh-install case in `docs/operations/search-retirement.md` and its Japanese pair, plus the refusal message.

Option 1 is preferable. Option 2 is what the code currently implies but does not say.

Either way, `docs/operations/search-retirement.{md,ja.md}` currently reads as if retirement is only for upgraded stores ("After the upgrade the family is inert"), and that framing is wrong for anyone installing v0.34 for the first time.

## Out of scope

The compaction preflight's refusal itself. It is doing its job.

Found while cutting the v0.34.0 RC; deferred out of v0.34 because the workaround is a single documented command and the fix touches the migration chain.

duck8823:
Fixed by #1872 (PR #1897), shipped in v0.35.0.

A source store that still carries the migration-032 family is no longer a compact-plan refusal. Compact drops the family on the work copy (`compaction_files.go` Plan comment; `compaction_copy_filter.go` `dropLegacySearchFamilyOn`; `legacy_search_retire_compaction_test.go`). Doctor now points at `store compact`, not `store search-retire`.

The titled sequence (`compact` refuses until `store search-retire`) cannot happen: `search-retire` is removed, and compact proceeds.

[...]

PR #1850, opened 2026-08-12
fix: correct broken documented commands and gate them in CI

### LINKED ISSUE #1849: Documented operator commands that do not exist or do nothing, including the downgrade backup step
Three commands the v0.34 documentation tells operators to run either do not exist or do nothing. One of them is the downgrade safety instruction.

Found by gpt-5.6-sol during the pre-tag release-readiness review, verified against the CLI.

## `traceary store backup` — the worst one

`CHANGELOG.md` and `docs/operations/payload-retention.{md,ja.md}` all say to take `traceary store backup` before first running v0.34, because after v0.34 writes compressed rows, v0.33.1 and earlier read those bodies as garbage and a backup is the only way back.

`store backup` is a parent command (`presentation/cli/backup.go:33`); the subcommand that makes a backup is `backup create` (`:49`). Running the documented form prints help and exits 0. An operator following the instruction believes they have a downgrade fallback for an irreversible change, and does not.

## `traceary store search-projection rebuild`

`docs/storage/payload-backfill.{md,ja.md}` step 6 tells the operator to rebuild the projection after a backfill with this command. It does not exist — the subcommands are `start`, `resume`, `status`, `abort`, `probe` (`presentation/cli/store_search_projection.go:16-53`). The step fails every time it is followed.

## bare `traceary store compact`

Step 7 of the same document. `store compact` is a parent command; it prints help and exits 0, having compacted nothing. The real sequence needs `search-retire`, then `compact plan`, then `compact apply RUN_ID`.

## Fix

Correct all three in both languages, and add a smoke check so a command spelled in the documentation is a command that exists. The last part is the actual fix — these three were introduced independently, and nothing catches the next one.

duck8823:
Closes #1849

Three commands the v0.34 documentation told operators to run either did not exist or did nothing. One of them was the downgrade safety instruction.

## The three

| Documented | Reality | Fixed to |
|---|---|---|
| `traceary store backup` | parent command; prints help, exits 0, makes no backup | `traceary store backup create` |
| `traceary store search-projection rebuild` | does not exist | `start` → `resume --until-complete` → `status` |
| bare `traceary store compact` | parent command; prints help, compacts nothing | `search-retire` → `compact plan` → `compact apply RUN_ID` → `compact status RUN_ID` |

`store backup` is the one that matters. `CHANGELOG.md`, `CHANGELOG.ja.md` and `docs/operations/payload-retention.{md,ja.md}` all told the operator to take it before first running v0.34, because after v0.34 writes compressed rows, v0.33.1 reads those bodies as garbage and a backup is the only way back. Someone following the instruction gets a help screen and a zero exit status, and believes they have a downgrade fallback for an irreversible change.

All three are corrected in both languages. The `payload-backfill` step 7 note also records that `compact plan` refuses while the legacy migration-032 index is resident (#1847), which is what actually happens when you follow the corrected sequence on a fresh store.

## The gate

`docs verify-commands`, alongside the existing `verify-i18n` / `verify-landing` / `verify-host-coverage`, wired into CI. It walks the live cobra tree from `cli.NewRootCLI().Command()` rather than a hardcoded list, and fails when documentation tells you to run a path that does not resolve, or one that resolves to a group with no action of its own.

**Scope is deliberately narrow: shell fences only** — `sh`, `bash`, `shell`, `console`. Everything else is out.

This took three attempts to get right, and the two rejected rules are worth recording because they look reasonable:

- *Any backticked mention.* 198 findings, all false. Prose legitimately names a command family: "`store compact plan` refuses while the family is present" is a correct sentence, not an instruction.
- *Any fenced block.* 121 findings, all false. A fence is whatever the fence says it is — `docs/integrations/antigravity.md` fences a **json** permissions file, `docs/lifecycle.md` fences an ASCII diagram, `docs/cli/README.md` fences sample **output**, `docs/architecture/host-native-memory-activation.md` fences the contents of a managed memory file, and `docs/release/v0.30.0-…` fences an archived transcript of a command that was a leaf at v0.30.

The unresolvable-in-prose class is the same problem in a different place: `docs/cli-stability.md` has a "Historical removal log" that names `traceary init`, `traceary backup`, `traceary gc` and sixty others on purpose. A rule that flags an instruction flags the removal log identically.

So the gate catches two of the three bugs in this issue. It does not catch `store backup`, which was inline prose. That limitation is stated in the subcommand's own help text so the next person does not widen the rule and rediscover the 198 false positives. A narrow check that is always right beats a wide one CI teaches everyone to ignore.

Archived release documents (`docs/release/**`) are excluded for the same reason released changelog sections are: they record what was run at the time.

## Verification

Passes on the current tree, and still fails on the content it was written for:

```
$ go run ./cmd/repo-tooling docs verify-commands
documentation command check passed
Summary: class 1 unresolved=0, class 2 fenced groups=0, files scanned=134, shell-fence invocations checked=410

$ git show cdda73f5^:docs/storage/payload-backfill.md > docs/storage/payload-backfill.md
$ go run ./cmd/repo-tooling docs verify-commands
Unresolvable command paths:
- docs/storage/payload-backfill.md:67: traceary store search-projection rebuild does not resolve to a command
Group commands in fenced code blocks:
- docs/storage/payload-backfill.md:74: traceary store compact is a group command and does not execute an action; use one of its subcommands: apply, plan, resume, rollback, status
documentation command check failed
```

`go build ./...`, `go test ./...`, `go tool golangci-lint run` (0 issues), `docs verify-i18n` all green.

Implementation delegated to gpt-5.6-luna; rule design, false-positive triage and review by Claude Opus 5.


duck8823:
gpt-5.6-sol review: 6 MUST, all verified against the code and all fixed in `03fa666c`.

The important one is #5 — **the fix for the worst bug was itself broken.** `store backup create` takes `maximumNArgsLocalized(1)` and returns `output path is required` when given neither a positional path nor `--output` (`presentation/cli/backup.go:48-65`). All six corrected call sites left off the destination, so the corrected downgrade-safety instruction exited nonzero and still made no backup. Same class as the original defect, one layer down. Now shows `traceary store backup create ~/traceary-pre-v0.34.db`.

| # | Finding | Fix |
|---|---|---|
| 1 | `README.md` / `README.ja.md` not scanned — 27 `traceary` lines in `sh` fences each, the most-read documents in the repo | added to the scan |
| 2 | any token after a leaf accepted as an argument without checking the leaf's contract; `traceary doctor nonexistent-subcommand` passed while cobra rejects it (`doctor` is `noArgsLocalized()`) | validate remaining tokens with `cobra.Command.ValidateArgs` |
| 3 | `traceary "nonexistent-command"` tokenized to an empty path and read as the bare root | strip quotes wrapping a whole token |
| 4 | unterminated `\` continuation at a fence close silently discarded — zero invocations reported | flush the accumulation at fence close |
| 5 | `store backup create` with no destination (above) | six call sites, both languages |
| 6 | step 6 presented `resume --until-complete` as driving the rebuild to completion; it is bounded by `--max-batches` (100) and `--total-wall-time` (10m) and exits 0 at `stop_reason=max_batches` while still `rebuilding` | step now says to repeat until `status` reports complete, and names both stop reasons |
| 7 | (NIT) Japanese step 6 said the rebuild requirement twice | dropped |

#2 deserves a note on direction. The tokenizer stops at flags, `<placeholders>` and `UPPER_CASE` words, so the argument count it validates is a lower bound; `validateDocumentedLeafArgs` pads with opaque probes so a minimum-arity command documented with placeholders still passes. Under-counting means a missed finding, never a false one — the right bias for a gate whose two earlier iterations died of false positives.

Coverage after the fixes: 136 files, 474 shell-fence invocations, up from 134 / 410.

```
$ go run ./cmd/repo-tooling docs verify-commands
documentation command check passed
Summary: class 1 unresolved=0, class 2 fenced groups=0, files scanned=136, shell-fence invocations checked=474
```

`go build ./...`, `go test ./...`, `go tool golangci-lint run` (0 issues), `docs verify-i18n` green. Stop-reason strings checked against `application/usecase/search_projection_usecase.go:158,186,207`; English and Japanese pairs re-read against each other.

sol also confirmed, as separate checks with no defect found: step 7's cobra paths and arity, and the claim that `compact plan` refuses while migration-032 objects remain — including on a store newly created by this version, since 032 creates the family and 052 leaves it resident (#1847).

Issue #1853, opened 2026-08-12
v0.36-15: reclaim the 2.24 GB of duplicated command_executed bodies as operator-run, resumable maintenance


duck8823:
v0.35. Follow-up to #1851, which deferred the reclaim out of v0.34's unconditional first-open upgrade.

#1675 stopped `command_executed` events from persisting a composed `events.body` (`command` + `INPUT` + `OUTPUT`) alongside the same payload in `command_audits`. New rows no longer double-store, and every reader was moved to the audit columns in PR #1731. **Historical rows still carry both.**

On the maintainer's live store the duplicate is:

- 305,025 rows
- **2,240,591,061 bytes** of `events.body`
- measured ratio of body to audit payload: 1.00 — pure duplication

That is 2.24 GB of a 32 GB store held by data no surface reads.

The clear itself is a single `UPDATE` — the SQL removed in #1851 is the reference:

```sql
UPDATE events
   SET body = '', body_codec = ..., body_format_version = ...,
       body_plaintext_bytes = 0, body_encoded_bytes = 0,
       body_sha256 = 'e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855'
 WHERE kind = 'command_executed'
   AND EXISTS (SELECT 1 FROM command_audits a WHERE a.event_id = events.id)
   AND (length(CAST(body AS BLOB)) > 0 OR COALESCE(body_plaintext_bytes,0) > 0 OR COALESCE(body_encoded_bytes,0) > 0)
```

Two properties of it are load-bearing and must survive whatever shape this takes:

- The `EXISTS` guard. `traceary log --kind command_executed` writes a body with **no** audit row; that body is the only copy of the payload and must survive.
- The empty-string SHA-256 constant, so later payload integrity checks stay consistent with `encodePayload(..., identity)`.

## Requirements

This must not become the thing #1851 removed. Specifically:

- Operator-initiated, never implicit on store open.
- Resumable with a durable cursor, so an interruption keeps its progress instead of restarting from zero. The `payload_backfill_runs` bookkeeping from migration `000054` and the `store payload-backfill` run/resume/status shape are the precedent — same store, same problem, already solved once.
- Clearing bodies fires the projection invalidation triggers, so `events.body` changes drive the projection to `drifted`/`stale`. Say so in the procedure and point at the rebuild, the way `docs/storage/payload-backfill.md` step 6 does.
[...]

PR #1854, opened 2026-08-12
fix: defer migration 000048 out of the v0.34 first-open upgrade

### LINKED ISSUE #1851: Migration 000048 cannot complete inside a hook timeout and retries identically forever
Migration `000048` cannot complete when Traceary is invoked through hooks. Measured on the maintainer's live store; this blocks the v0.34 tag.

## The stall

`initializeAt` (`infrastructure/sqlite/database.go:349-372`) always calls `migrate()`, including from hook invocations. `migrate()` (`infrastructure/sqlite/migrate.go:34-41`) applies every unapplied migration unconditionally, and `applyMigration` (`:135-165`) runs each one in a single transaction — killed mid-flight it rolls back completely and the next open restarts it from zero.

`integrations/claude-plugin/hooks/hooks.json` sets no `timeout`, so the host default applies (60 s in Claude Code).

Measured on `~/.config/traceary/traceary.db` (32,130,342,912 bytes; 651,621 events), read-only via the `sqlite3` CLI with `mode=ro` and `PRAGMA query_only=1`:

| Migration | Class | Cost on the live store |
|---|---|---|
| `000035` | `data_dependent_offline` | two indexes over `event_metadata_projection`, 652,214 rows — seconds |
| `000041` / `000042` | `data_dependent_offline` | projection-state rows only — negligible |
| `000045` | `data_dependent_offline` | `raw_body_retention_entries` has **0 rows** — free |
| `000048` | `data_dependent_offline` | **33.7 s for the read half alone**; 305,025 rows; 2,240,591,061 body bytes to clear |

048's read half is 33.7 s against a 60 s hook timeout, and the `UPDATE` — six columns on all 305,025 rows, freeing 2.24 GB of overflow pages — is strictly more expensive. So the first hook after upgrade is killed mid-048, rolls back, and the next hook starts 048 from zero. Identically, forever. An operator who only ever reaches Traceary through hooks never gets past migration 047 and is told nothing.

`v0.33.1` ships migrations through `000034`, so every store upgrading to v0.34 walks this path.

The `UPDATE` itself is not measured: the machine has 40 GB free against a 32 GB store, so there is no room for the copy a safe write test would need. The ">60 s" step is inference from "the write is strictly more expensive than the read", not measurement.

## Why 048 can be deferred without losing the feature

The reader half of #1675 already shipped. PR #1731 changed 40+ files so that list, doctor sensitive-path, replay failure hotspots, MCP and the display surfaces read the `command_audits` columns joined in the same query rather than `events.body`. New `command_executed` rows already stop carrying a composed body.

048 only clears **historical** bodies. It is a 2.24 GB space reclaim, not a correctness requirement, and no v0.34 feature depends on it.

## Change

- Remove `schema/sqlite/migrations/000048_clear_command_executed_event_bodies.sql`.
- Remove entry 48 from `preparedMigrationManifest` (`infrastructure/sqlite/prepared_migration_catalog.go:55`) and from the catalog test.
- Correct the #1675 changelog entry in both languages: the write-path change ships in v0.34; the historical clear does not.
- Correct `docs/storage/README.{md,ja.md}` and `docs/operations/event-projection-contract.{md,ja.md}`, which currently state that the body extent for `command_executed` is zero after migration 048. Until the reclaim ships, historical rows still carry bodies.

Version 48 becomes a gap in the catalog. `rejectForeignAppliedMigrations` (`migrate.go:45-56`) tolerates missing versions by design — "released history contains gap upgrades" — but it rejects a ledger that records a version the catalog does not contain at or below its maximum, so a development store that already applied 048 will refuse to open. No release ever shipped 048, so this affects unreleased development stores only. Call it out in the PR.

## Follow-ups, not this issue

- The historical reclaim itself, as an operator-run command rather than an implicit migration.
- The general hazard: nothing in the runtime path consults `MigrationExecutionClass`. It exists only in `prepared_migration_catalog.go` and its test. 048 is the instance that bites today; the class is not gated at all.

duck8823:
Closes #1851

Migration `000048` cannot complete inside a hook timeout, and every retry repeats identically. This takes it out of v0.34.

## The stall

`initializeAt` (`infrastructure/sqlite/database.go:349-372`) always calls `migrate()`, hooks included. `migrate()` applies every unapplied migration unconditionally, and `applyMigration` (`migrate.go:135-165`) runs each in a single transaction — killed mid-flight it rolls back completely and the next open restarts it from zero. `hooks.json` sets no `timeout`, so the host default applies (60 s in Claude Code).

Measured on the maintainer's live store (32,130,342,912 bytes; 651,621 events) read-only via the `sqlite3` CLI with `mode=ro` and `PRAGMA query_only=1`:

| Migration | Class | Cost |
|---|---|---|
| `000035` | offline | two indexes over `event_metadata_projection`, 652,214 rows — seconds |
| `000041` / `000042` | offline | projection-state rows only — negligible |
| `000045` | offline | `raw_body_retention_entries` has **0 rows** — free |
| `000048` | offline | **33.7 s for the read half alone**; 305,025 rows; 2,240,591,061 body bytes |

048's read half is 33.7 s against a 60 s timeout, and the `UPDATE` — six columns on all 305,025 rows, freeing 2.24 GB of overflow pages — is strictly more expensive. So the first hook after upgrade is killed mid-048, rolls back, and the next hook starts over. An operator who only reaches Traceary through hooks never gets past 047 and is told nothing. v0.33.1 ships through `000034`, so every upgrading store walks this path.

048 is the only one that crosses the timeout. That is why this is a removal rather than the gating change.

The `UPDATE` itself is unmeasured: 40 GB free against a 32 GB store leaves no room for the copy a safe write test needs. The ">60 s" step is inference from "the write is strictly more expensive than the read".

## Why deferring loses nothing

The reader half of #1675 already shipped in PR #1731 — 40+ files moved to the joined `command_audits` columns, and new `command_executed` rows already carry no composed body. 048 only clears **historical** bodies. It is a 2.24 GB space reclaim, not a correctness requirement, and no v0.34 feature depends on it. #1853 tracks it as operator-run, resumable maintenance, which is what it should have been.

## Changes

- `schema/sqlite/migrations/000048_clear_command_executed_event_bodies.sql` deleted, with its manifest entry and its dedicated test.
- The `#1675` changelog entry corrected in both languages: the write-path change ships, the historical clear does not, and #1853 is named. A reader who upgrades and still finds 2.24 GB of duplicate bodies needs the changelog to have said so.
- `docs/storage/README.{md,ja.md}` and `docs/operations/event-projection-contract.{md,ja.md}` corrected — they stated the post-048 world as fact ("this column is empty", "extent for this kind is zero after migration `000048`"). Plus four code and SQL comments asserting the same, found by sweeping for the claim rather than for the migration number.

## The catalog gap, and a check that got stronger

Version 48 becomes a hole: 47 then 49. `rejectForeignAppliedMigrations` tolerates missing versions by design, and I confirmed it at runtime rather than by reading — a scratch store initialized with `TRACEARY_DB_PATH` reports `45,46,47,49,50,51,52`.

But `BuildPreparedMigrationPlan` carried a **contiguity** check that rejected any gap. Deleting it to accommodate an intentional gap would have removed the guard against unintentional ones — a migration dropped from the `go:embed` set would leave 49, 51 with 050 silently absent and publish a prepared store missing a schema change while reporting success. The manifest cross-check does not cover this: it walks `catalog[applied:]` asserting each catalog entry has a manifest entry, never the reverse.

So contiguity is replaced by a stronger property: **the catalog must contain exactly the versions the manifest declares**, matching name and digest. The manifest is source-owned, so it accommodates 048's absence automatically while still catching a vanished file. Coverage divides cleanly — the exact-prefix walk covers versions at or below the applied prefix, and manifest-set equality covers everything above it, since the manifest starts at 35 and prepared plans require `current >= 34`.

Two tests pin both directions: deleting `000020` (below the manifest range) is caught by the exact-prefix walk, and deleting `000050` is rejected naming version 50.

## Note for development stores

A store that already applied 048 will refuse to open — `rejectForeignAppliedMigrations` rejects a ledger recording a version the catalog does not contain at or below its maximum. No release ever shipped 048, so this is unreleased development stores only.

## Verification

`go build ./...`, `go test ./...`, `go tool golangci-lint run` (0 issues), `docs verify-i18n`, `docs verify-commands` all green. Runtime gap confirmed as above.

Design consultation with gpt-5.6-sol, which independently raised this as a release blocker; implementation delegated to gpt-5.6-luna; measurement, scope decision and review by Claude Opus 5.


duck8823:
gpt-5.6-sol review: 1 MUST, verified and fixed in `fdb0bd0f`. Four categories came back clean.

## The MUST

My claim that manifest-set equality is *strictly stronger* than the contiguity check it replaced was wrong. It is stronger only **at or above version 35**, where the manifest declares something. Below that the manifest declares nothing, and the exact-prefix walk compares the ledger against the catalog — so it passes whenever the two agree.

sol's counterexample:

- catalog missing `000020_add_session_model.sql`
- ledger missing version 20 as well, because the store was built by that same defective catalog
- ledger `1..19, 21..34`, names matching the catalog throughout

Manifest check inspects 35 and up: passes. Exact-prefix walk finds ledger and catalog in agreement: passes. `current` is 34: passes. Pending suffix matches the manifest: a plan is returned and a prepared store is published from a catalog missing a schema change, reporting success. The old contiguity check caught this by comparing version 21 against catalog index 20.

My code comment claimed "a ledger recording version 20 cannot pass if catalog version 20 is absent" — true, and beside the point. The dangerous case is the ledger not recording it either.

The existing test could not catch it: it deleted 20 from the catalog while leaving 20 in the ledger.

**Fix.** Use the strongest invariant available for each range, split at the manifest's minimum version rather than a hardcoded 35:

- below the minimum — catalog versions must be contiguous from 1, the only invariant available there
- at or above it — the catalog must equal the manifest exactly, name and digest, which tolerates 048's deliberate absence and is genuinely stronger than contiguity

New regression test initializes the store with the defective catalog rather than editing `schema_migrations` afterwards, so it reproduces the counterexample as stated, and asserts `discontinuity at version 20`. The three existing cases stay: catalog-only gap at 20 caught by the prefix walk, missing `000050` rejected naming version 50, current tree builds cleanly.

## Clean

- **Code assuming 048 ran.** sol swept independently for readers, size accounting, capacity, doctor, projection extent and the benchmark harness, not just comments. Everything is body-first with audit fallback: `event_bounded_datasource.go:346-365`, `event_preview_datasource.go:61-94`, `sql/list_sessions.sql:67-78` (`NULLIF(body,'')` first), `show.go:119-131`, `mcpserver/server.go:2250-2257`. Extents are summed as separate real families (`search_projection_rebuild.go:603`), capacity aggregates residual `body_stored_bytes` as-is (`capacity_inspector.go:117-147`), and `store-benchmark` seeds non-empty bodies on command events. No path returns a historical body as silently empty.
- **Development-store claim.** Confirmed both halves: version 48 is absent from the catalog and below its maximum of 59, so `migrate.go:61-76` rejects it; 048 landed after `v0.33.1` and is in no release tag.
- **English/Japanese parity**, including the `event-projection-contract` paragraph I repaired by hand.
- **Changelog accuracy.** sol notes the heading alone reads broadly but the body immediately states that existing bodies survive the upgrade; not a required change.

PR #1855, opened 2026-08-12
fix: bound payload rehearsal failure cleanup

### LINKED ISSUE #1791: v0.34.0 follow-up: payload rehearsal bounds Pause/ReleaseScrub on some paths and not others

duck8823:
Closes #1791

`application/usecase/payload_rehearsal.go` calls `Pause` / `ReleaseScrub` six times on terminal paths. Three were bounded to one second; three passed `context.WithoutCancel(ctx)` directly.

`context.WithoutCancel` strips the caller's deadline along with its cancellation, so those three waited on nothing. Both calls reach the store, and acquiring the store lease waits only on the context — so a concurrent lease holder made those sites hang with no ceiling. Lines 99 and 168 sit on the failure path of a `liveResult` check, which runs exactly when something is already wrong.

Losing a checkpoint is recoverable. Hanging is not. `infrastructure/sqlite/payload_backfill.go` already records that reasoning for the equivalent write, and the three bounded sites in this same function already implement it.

## Change

All three now use the idiom the bounded sites already use — `context.WithTimeout(context.WithoutCancel(ctx), time.Second)` with `cancel` called. The file has one spelling instead of two. No timeout value changed, no error handling changed, nothing outside this file and its tests touched.

## Tests

`TestPayloadRehearsalBoundsFailureCleanup`, table-driven over the three paths: `run` via the prepare-then-`liveResult`-failure route, `run` via the advance route, and `Scrub`. Each stub blocks until its context is done, so the test distinguishes "returned because the bound fired" from "returned immediately".

Verified they pin the defect rather than passing either way — with `application/usecase/payload_rehearsal.go` restored from `origin/main`:

```
--- FAIL: TestPayloadRehearsalBoundsFailureCleanup (4.51s)
    --- FAIL: .../run_prepare_live_result_failure (1.50s)
        payload_rehearsal_test.go:256: cleanup call did not return within the one-second bound
    --- FAIL: .../run_advance_live_result_failure (1.50s)
    --- FAIL: .../scrub_prepare_live_result_failure (1.50s)
```

All three pass on this branch.

## Documentation

Checked whether any release note claimed these paths were already bounded. The CHANGELOG's "migration, scrub, rollback, WAL reservations are bounded" refers to resource limits in the processing body, not failure cleanup, and no document claims all six cleanup sites are bounded. The release note was not overclaiming.

### DIFF
--- a/application/usecase/payload_rehearsal.go
+++ b/application/usecase/payload_rehearsal.go
@@ -96,7 +96,9 @@ func (u *payloadRehearsalUsecase) run(ctx context.Context, c types.PayloadRehear
 		return result, xerrors.Errorf("prepare payload rehearsal: %w", err)
 	}
 	if result, err = liveResult(result); err != nil {
-		_, _ = u.workflow.Pause(context.WithoutCancel(ctx), handle)
+		pauseCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), time.Second)
+		_, _ = u.workflow.Pause(pauseCtx, handle)
+		cancel()
 		_ = u.workflow.Close(handle)
 		return result, err
 	}
@@ -119,7 +121,9 @@ func (u *payloadRehearsalUsecase) run(ctx context.Context, c types.PayloadRehear
 				return result, xerrors.Errorf("advance payload rehearsal: %w", err)
 			}
 			if result, err = liveResult(result); err != nil {
-				_, _ = u.workflow.Pause(context.WithoutCancel(ctx), handle)
+				pauseCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), time.Second)
+				_, _ = u.workflow.Pause(pauseCtx, handle)
+				cancel()
 				return result, err
 			}
 			if c.StopAfterBatches > 0 && result.BatchCount-initialBatches >= c.StopAfterBatches && (!done || fieldIndex != len(fields)-1) {
@@ -165,7 +169,9 @@ func (u *payloadRehearsalUsecase) Scrub(ctx context.Context, c types.PayloadRehe
 		return types.PayloadRehearsalMetrics{}, xerrors.Errorf("prepare payload rehearsal scrub: %w", err)
 	}
 	if result, err = liveResult(result); err != nil {
-		_ = u.scrubber.ReleaseScrub(context.WithoutCancel(ctx), handle)
+		releaseCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), time.Second)
+		_ = u.scrubber.ReleaseScrub(releaseCtx, handle)
+		cancel()
 		_ = u.scrubber.CloseScrub(handle)
 		return result, err
 	}
--- a/application/usecase/payload_rehearsal_test.go
+++ b/application/usecase/payload_rehearsal_test.go
@@ -11,6 +11,9 @@ import (
 	"github.com/duck8823/traceary/application"
 	apptypes "github.com/duck8823/traceary/application/types"
 	"github.com/duck8823/traceary/application/usecase"
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"golang.org/x/xerrors"
 )
 
 type rehearsalBackendFake struct {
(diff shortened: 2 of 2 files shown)


duck8823:
gpt-5.6-sol review: **no MUST on this PR.** Every category came back clean.

- **The one-second bound.** Same idiom as the three sibling sites. `Pause` / `ReleaseScrub` enter `walBudgetedMutationSession.run` → `db.Conn(ctx)` → `BEGIN IMMEDIATE` → lease row update. The DSN `busy_timeout` is `LockTimeLimit`; where the cleanup second is shorter, the context is the ceiling. A best-effort terminal cleanup existing only to let the original error surface is the right place for a cap independent of the normal lock budget.
- **Discarding the `Pause` result stays recoverable.** On success the pause or scrub release commits durably. On timeout the handle is closed and the persisted lease expires within 30 seconds; `running` satisfies `CanResume()`, so the next `Resume` reacquires. Worth knowing: the *first* retry inside that 30-second window can still report "leased by another worker". Not a stall, and no checkpoint is lost.
- **Handle coverage.** Every return path in `run` and `Scrub` either pauses-and-closes explicitly or reaches the deferred close. No path exits leaving a handle open, panics aside.
- **Success path untouched.** No added context, SQL, lease operation, checkpoint, commit or WAL work. On a 32 GB store, success-path latency, lease hold time and checkpoint durability are unchanged. The change is confined to the failure branch.
- **The other five unwrapped `WithoutCancel` sites are safe, and sol says explicitly not to wrap them reflexively.** The four `ROLLBACK` calls (`payload_rehearsal_store.go:83`, `payload_rehearsal_migration.go:159`, `run_lineage_datasource.go:63`, `usage_observation_datasource.go:154`) each roll back a transaction the same connection already owns — no new connection, no lease. `search_projection_rebuild.go:991` is post-commit maintenance bounded to 8 steps × 16 output pages with the budget checked between steps; the batch's durable progress is already committed, so it cannot repeat forever.

One noted limitation, below follow-up severity: the blocking fake waits on `ctx.Done()` only, so the tests prove the current code returns on timeout but do not assert an elapsed-time lower bound or `context.DeadlineExceeded`. A future change passing an already-cancelled cleanup context would still pass them. The three subtests took ~3.47 s in this run, which does rule out immediate return for the code as it stands.

sol also raised a **separate** concern outside this PR's scope, at `infrastructure/sqlite/payload_rehearsal.go:467`. Being pursued separately; deliberately not merged into this branch.

`go build ./...`, `go test ./...`, `go tool golangci-lint run` (0 issues), `docs verify-i18n`, `docs verify-commands` green; CI green on all five jobs.

PR #1856, opened 2026-08-12
fix: disclose unavailable session search tier

### LINKED ISSUE #1844: traceary search drops the SESSIONS group silently when the projection is not ready
## What

`traceary search` drops the SESSIONS group silently when the projection is not read-ready. The MCP `search` tool does not: it calls `SearchSessionProjectionReady` and appends the `session_projection_not_ready` reason.

`SearchSessionHits` (`infrastructure/sqlite/search_projection_session_query.go:63`) returns an empty slice when `searchProjectionReadReady` is false — the same value a ready projection with no match returns. That ambiguity is the reason `SearchSessionProjectionReady` exists as a separate call, and its own comment says so. `presentation/mcpserver/server.go` uses it; nothing in `presentation/cli` does.

## Why it matters

A session-tier hit is a session whose stored summary or keyword text matches. That text is not the events' text, so a match can exist there and nowhere else. When the projection is incomplete — which every upgraded store is until its first generation finishes, and every store is again during a rebuild — the CLI reports zero session hits and says nothing. The operator cannot tell "no session matched" from "the session tier was not consulted".

This is the silent-false-negative class in its plain form, and it is on the surface most people use.

## Scope

- [ ] `traceary search` reports that the session tier was not consulted, on stderr, in the same place the `--kind` suppression notice already goes (`warnSearchSessionsSuppressedByKind`, `presentation/cli/search_output.go:40`) — that notice is the existing precedent for "your session group is empty for a reason"
- [ ] The notice states the recovery: `traceary store search-projection start` / `resume`, or simply that store opens advance it on their own
- [ ] It fires only when the tier was actually refused, never when it was consulted and matched nothing
- [ ] stdout stays byte-identical, as it does for the existing notices
- [ ] A test asserts the notice appears for a not-ready projection and does **not** appear for a ready one with no matches

## Not in scope

Whether `--json` should carry the reason. v0.34's `search --json` is a frozen array (see `docs/cli-stability.md`); the v0.35 object shape is where a machine-readable reason belongs.

Documented as a known behaviour in #1841 pending this fix. Found by gpt-5.6-sol reviewing #1843.

duck8823:
Closes #1844

`traceary search` dropped the SESSIONS group silently when the search projection was not read-ready.

`SearchSessionHits` returns an empty slice when `searchProjectionReadReady` is false — the same value a ready projection with no match returns. That ambiguity is why `SearchSessionProjectionReady` exists as a separate call, and the port's own doc comment says a surface that loses the distinction "reports 'nothing found' for history that exists, and nothing about that failure is visible to the caller". The MCP server asked for readiness and appended `session_projection_not_ready`. Nothing in `presentation/cli` did — on the surface most people actually use.

A session-tier hit is a session whose stored summary or keyword text matches. That text is not the events' text, so a match can exist there and nowhere else. Every upgraded store is in that state until its first generation finishes, and every store is again during a rebuild.

## Change

Two stderr notices, written where the existing `--kind` suppression notice goes. Both call sites get them — the normal path and the `--json --fields` metadata path.

Readiness is asked for **only when the hit page is empty**, mirroring the MCP branch, so a projection that was consulted and matched nothing never produces a notice.

stdout is byte-identical in every case, and a nil `projectionSessionSearch` behaves exactly as before.

## What the notice says, and how it got there

The wording took four review rounds with gpt-5.6-sol, because every draft that named a recovery named one that does not work in some reachable state — which would have produced exactly the permanent stall the notice exists to prevent.

| Draft | Why it was wrong |
|---|---|
| "the rebuild advances on ordinary store opens" | a `failed` generation is parked forever (`search_projection_usecase.go:354-368`) |
| "a mismatch only happens if you started it with custom budget flags" | true today, but only because `ConfigHash` and `SearchProjectionCapacitySemanticsVersion` happen to be in step — filed as #1861 |
| "if `status` reports the same checkpoint across several opens, it is not advancing" | not a progress signal: inventory advances a separate cursor and cleanup deletes rows while holding the checkpoint. Reports "stalled" for a healthy rebuild |
| "…and `start` replaces the generation" | `StartGeneration` refuses whenever the state is `rebuilding` (`:59-61`), so this is the one command that cannot clear a parked rebuild |

Final wording names the state and the two states that do not clear themselves, and nothing else:

> traceary: the session tier was not consulted because the search projection is not in the `complete` state. Run `traceary store search-projection status`: what the reported state needs differs, and two of them do not advance on ordinary store opens — a `failed` generation needs `traceary store search-projection start`, and a rebuild started with a non-default budget needs `resume` or `abort` carrying that same budget.

"is not in the `complete` state" rather than "has no complete generation": readiness is exactly `state == 'complete'` with an active generation (`event_search_query.go:66-73`), and rows from an earlier generation can still be present. The function's doc comment records why the checkpoint heuristic was removed, so it does not come back.

A second notice covers the case the first cannot: if the readiness check itself errors, `warnSearchSessionsProjectionReadinessUnknown` says the empty group could not be attributed. Staying silent there would restore the exact ambiguity this PR removes — an empty group that means nothing in particular, with nothing saying so.

## One change beyond the mechanical fix

The CLI held `ProjectionSessionSearchQuery` — the query half of the port — so readiness could only be reached through a type assertion, which silently no-ops on an implementation that lacks it. That reintroduces the same lost distinction one layer up, and the port's doc comment explicitly says readiness is "required rather than probed for". `RootCLI` now takes the composed `ProjectionSessionSearch`, as `mcpserver.Server` already did, so the compiler rejects that case instead of the notice quietly disappearing.

## Tests

`presentation/cli/search_session_projection_ready_test.go`, table-driven over not-ready/ready-empty/ready-with-hits, through both the text and `--json --fields` paths, plus a readiness-error case. The stdout assertions compare against the ready-and-empty run directly with `cmp.Diff` rather than against a golden string, so "stdout unchanged" is checked rather than asserted. The not-ready case counts the notices rather than only checking for the wrong one, so a third notice slipping in also fails.

## Documentation

The v0.34 changelog entry for #1717 and `docs/search-projection-rebuild.md` both stated that `traceary search` does not report this. That becomes false with this change, so both were corrected, in English and Japanese, along with `docs/cli/README.{md,ja.md}`. All six carry the same final wording — the rebuild document had been contradicting its own paragraphs at 24-27 and 29-36, which already described the two non-advancing states correctly.

## Follow-ups filed, not fixed here

- #1859 — hits and readiness are read in separate transactions
- #1861 — a future default-budget change could park an automatically started generation with no operator involvement


duck8823:
gpt-5.6-sol review: **2 MUSTs, both verified against the code and fixed** in `9155ae3d`. Clean on stdout/JSON contract, port change and wiring, and English/Japanese parity.

## MUST 1 — the recovery advice could send the operator into a permanent stall

The notice told every reader to run `start` or `resume` and promised "ordinary store opens advance it automatically". False in two states, and one never recovers by following it:

- **failed generation.** `CatchUp` parks it and skips identically on every open (`application/usecase/search_projection_usecase.go:354-368`). That code's own comment says it: "neither resume nor abort clears this state. Resume rejects a failed generation, and abort leaves the row failed with class abandoned. Only an explicit start replaces it." Following the advice would mean opening the store forever, seeing the same warning forever, and never making durable progress — the permanent-stall class, introduced by a notice meant to prevent a silent one.
- **rebuilding under a non-default budget.** Ordinary opens skip with "budget does not match generation configuration" (`:346-353`); recovery needs a `resume` carrying a matching budget.

Readiness is false for every state except `complete` with an active generation (`infrastructure/sqlite/event_search_query.go:67-73`), so the notice fires in exactly those states.

The notice now sends the reader to `traceary store search-projection status` first and names both families honestly. The same overclaim was corrected in `docs/search-projection-rebuild.{md,ja.md}`, `docs/cli/README.{md,ja.md}` and the #1717 changelog entry, both languages. The CLI does not re-derive the state machine — `status` remains the surface that reports it.

## MUST 2 — a readiness-check failure reproduced the exact bug this PR fixes

Swallowing the readiness error meant: `SearchSessionHits` correctly detects not-ready and returns empty → the readiness call fails → the CLI exits 0, prints an empty SESSIONS group, and says nothing. The original silent false negative, and the test pinned it as correct.

The exit code and stdout still do not change on a readiness failure — the events were found and printed — but a distinct notice now says readiness could not be determined, so an empty group is never silently ambiguous. Test renamed and rewritten to assert that.

## Beyond the review

`searchProjectionSessions` had grown to a five-value return with three parallel bools, and each of the two call sites repeated the same three `if` blocks. Collapsed into a `searchSessionNotices` value with a `write` method, so the "none of these may swallow another" rule lives in one place. No behaviour change; the `--json` omitted-count notice stays at its call site because it is the only one conditioned on output format.

Also replaced an `if tc.name == "..."` branch in the new table test with a `wantOnlyNotice` field.

## Verified clean by sol

- **stdout / JSON contract.** The notice only ever reaches `warnWriter`. The stdout assertions compare against the ready-and-empty run, which does prove the readiness branch changes nothing; existing tests separately pin the top-level `[]\n` shape. No serializer, field, array-shape or ordering change in the diff.
- **Port change and wiring.** CLI and MCP both now require the composed `ProjectionSessionSearch`. `*sqlite.EventDatasource` carries a compile-time assertion for it (`search_projection_session_query.go:19`), and `main.go:254,276` pass that same datasource to both — so it compiles by construction, not coincidence. There is no third production consumer holding the query half alone.
- **`--kind` combinations.** All six stable-state combinations produce the right notice or none; no state produces both. A refused tier makes the kind-less probe empty too, so kind-suppression cannot fire alongside not-ready. That combination was untested and now is.
- **Documentation parity.** No remaining file asserts the old "the CLI does not report this" behaviour, and nothing now claims the CLI reports something it does not.

## Split out, not fixed here

sol also found a TOCTOU: hits and readiness are read in separate transactions, so a generation transition between them can drop a notice or produce a stale one. Closing it properly means returning hits and disposition from one read transaction — a port and adapter change. Filed as #1859 rather than widening this PR.

duck8823:
## Review record — gpt-5.6-sol, eight passes

The code change was small. The wording took eight rounds, because five consecutive drafts of a one-line stderr notice were each false in some reachable state — and a notice that names a command the store will refuse creates exactly the permanent stall it was written to prevent.

| Pass | What was wrong |
|---|---|
| 1 | readiness reached through a type assertion, silently no-op on an implementation lacking it — the same lost distinction, one layer up |
| 2 | omitting the notice on a readiness error restored the original ambiguity |
| 3 | "a mismatch only happens if you started it with custom budget flags" — true only because `ConfigHash` and the capacity semantics version happen to be in step (#1861) |
| 4 | "watch whether the `status` checkpoint moves" is not a progress signal: inventory advances a separate cursor and cleanup deletes rows while holding the checkpoint |
| 5 | `start` is refused whenever the state is `rebuilding`, so the one command the notice named was the one that could not clear a parked rebuild; `abort` takes no budget and parks rather than unblocks |
| 6 | the `status` table omitted the obsolete-capacity-semantics override; `docs/cli/README` still told the reader to run `start` then `resume` |
| 7 | the override does not apply to `failed`, which stays parked deliberately; the Japanese line 7 still contradicted the table |
| 8 | **Clean. Merge OK.** |

What changed at pass 5 was the approach, not the sentence. The notice now names **no command at all** — it states that the tier was not consulted, why, and what that costs the reader, then sends them to `status` for the state and to `docs/search-projection-rebuild.md` for what that state needs. The recovery matrix lives in a table there, which has room to be right: seven state rows plus the semantics override and its `failed` exception.

Every finding was checked against the code before being accepted, and the deferral below was argued rather than assumed.

## Known limitation, deferred

**#1859 — hits and readiness are read in separate transactions.** A generation that changes state between the two reads can produce a notice for a tier that was consulted, or suppress one for a tier that was not. Fixing it means returning consultation status atomically with the page, which changes the `ProjectionSessionSearch` contract and both implementations — at RC time, to close a window that needs a concurrent generation transition inside one command.

sol reviewed the deferral and agreed: 「2 transaction 間の短い窓に限定され、影響も stderr notice の誤報・欠落です。常態的な結果欠落ではありません。」 Without this PR the tier is silently dropped in *every* not-ready state, not just during a race.

Also filed and not fixed here: **#1861** — a future default-budget change could park an automatically started generation with no operator involvement. The notice's wording is deliberately independent of that.

Issue #1858, opened 2026-08-12
v0.40-15: payload-rehearsal rollback verification has no wall-time budget


duck8823:
Deferred from v0.34. The release-note half is fixed in #1857; this issue is the code defect behind it.

## What

`store payload-rehearsal` rollback runs several full-file and full-store read-only operations with no wall-time budget, and the operator's `--wall-time-limit` does not reach any of them.

On the recovery path (`infrastructure/sqlite/payload_rehearsal.go:467`) and the explicit `rollback` path:

- `rehearsalPreparationBinding` → `fileDigest(c.BackupPath)` (`payload_rehearsal_target.go:378-389`) — a context-unaware `io.Copy` over the whole backup file
- `VerifyRollbackTarget` → `CanonicalEventAuditDigest` (`prepared_migration_verifier.go:20-34`) — a full canonical scan
- on the explicit path, backup verification, the copy, and a post-restore `PRAGMA integrity_check`

`--wall-time-limit` is a manual comparison inside the rehearsal batch loop (`application/usecase/payload_rehearsal.go:110`) and a ctx timeout for the migration build (`prepared_migration_recipe.go:57`). Neither covers this.

The state-mutating rollback itself **is** bounded — `preparedStoreUpgradeUsecase.Rollback` derives `lockCtx` from `run.Budget.PublishLockLimit`, capped at one second by `compaction_files.go:304`. The gap is entirely in the verification preceding it.

## Why it matters, and why it is not a v0.34 blocker

On a 32 GB store the backup hash alone is a minute-scale operation and the canonical scan is worse. The operator set a wall-time limit and it is silently ignored, with no progress output — `presentation/cli/store_payload_rehearsal.go:80-86` prints JSON only after everything finishes.

Two things I checked before deferring it:

- **No caller deadline is being stripped.** `context.WithoutCancel` at `:467` removes SIGINT/SIGTERM only; non-hook commands carry no deadline at all (`main.go:351-368`). The operation was already unbounded before that call.
- **Nothing retries it automatically.** Line 467 returns `ErrPreparedMigrationPublish` and the command exits. This is not the migration-048 shape, where a host re-ran the same work forever against a 60-second timeout.

So it is a bad experience on a large store, not a permanent stall. gpt-5.6-sol raised it, and withdrew both the deadline-loss and automatic-retry framings on being shown the above.

## Shape (sol's recommendation, recorded rather than settled)

[...]

Issue #1227, opened 2026-06-20
v0.21.3-2: add reversible historical hook content dedupe


duck8823:
Parent: #1226

## Motivation

`traceary doctor --client codex --warnings-ok --json` currently reports historical hook-originated prompt/transcript duplicates in the local DB even though the new write path suppresses fresh duplicate writes.

Current dogfood signal:

- `audit-reliability`: pass; recent command audits have no duplicate group / workspace drift candidate.
- `content-event-reliability`: warn; recent prompt/transcript hook events include historical duplicate groups.

The cleanup path should reduce operator noise and future context size without silently destroying local-first audit history.

## Decision

Do **not** dedupe historical rows automatically during ordinary upgrade/migration. Add an explicit, reversible maintenance path instead.

A schema migration may add additive support tables, but it must not move, delete, or rewrite existing `events` rows by itself.

## Proposed behavior

- Add a store maintenance command, for example `traceary store dedupe content-events`.
- Default mode is dry-run and prints candidate groups only.
- `--apply` is required to change data.
- Keep one canonical row per duplicate group and quarantine/archive the rest rather than hard-deleting them.
- Store enough metadata to restore quarantined rows or explain exactly what was moved.
- Use the same conceptual scope as the write-side guard:
  - only `client=hook`;
  - only `kind=prompt` and `kind=transcript`;
  - identity includes `kind`, `client`, `agent`, `session_id`, `workspace`, `source_hook`, and body identity;
  - command audits are not touched.
- Parse `created_at` in Go rather than relying on lexical SQL ordering for variable-width RFC3339Nano timestamps.

## Acceptance criteria

- Dry-run reports duplicate groups with kept event ID, candidate duplicate event IDs, reason, and counts; JSON output is available.
- Apply mode requires an explicit flag and either creates a backup or records a restore-capable quarantine run.
- Quarantined duplicate rows do not appear in normal `list`, `sessions --snapshot`, `doctor`, context, or MCP read surfaces by default.
- Restore path or documented rollback path can put quarantined rows back without overwriting existing event IDs.
- Running apply twice is idempotent.
- Ambiguous timestamp or malformed timestamp groups are skipped by default and reported.
- `content-event-reliability` points to the new dry-run command and makes clear that current write paths already suppress known fresh duplicates.
- Tests cover dry-run, apply, restore/rollback, idempotency, malformed timestamps, command-audit exclusion, and normal read-surface exclusion.
- Docs are updated in English and Japanese.

[...]

PR #1860, opened 2026-08-12
docs: correct v0.34 boundedness claims

### LINKED ISSUE #1857: v0.34 release notes claim boundedness the binary does not provide
The v0.34 CHANGELOG asserts hard boundedness in four places the binary does not provide. A release note that claims more than the binary does is the same failure class as a surface reporting a wrong number as exact: the reader acts on it, and nothing signals that it was wrong.

Found by gpt-5.6-sol auditing the v0.34 entries after PR #1855; each item below was independently re-verified against the code before filing. English and Japanese assert the same thing, so every fix is a paired edit.

## 1. `store compact` — "bounded candidate verification"

`CHANGELOG.md:13` / `CHANGELOG.ja.md:13`

`newStoreCompactionRunCommand` passes `cmd.Context()` straight through (`presentation/cli/store_compaction.go:39-65`), and non-hook commands get a context with no deadline at all (`main.go:351-368`). Under it, verification runs source and candidate `integrity_check`, `foreign_key_check`, a schema digest, a logical digest over every table, and the payload codec scrub (`infrastructure/sqlite/compaction_sqlite.go:86-154`, logical scan from `:177`).

Candidate verification is a full scan with no configured wall or resource cap. The word "bounded" beside "atomic" reads as a resource guarantee and is not one.

## 2. `store gc` — "Discovery is capped by range and wall-time budgets"

`CHANGELOG.md:21` / `CHANGELOG.ja.md:21`

`DiscoverCandidates` runs at `application/usecase/orphan_consolidation_usecase_impl.go:91`, immediately after `started := u.clock.Now()` and **before** any budget comparison. The wall-time budget is checked at `:112`, inside the per-candidate loop.

So the range limit caps discovery cardinality (default 5000) and the wall budget caps the processing loop between candidates. Discovery itself is not wall-time capped, and a single `LoadMaterial` / `Refine` that overruns the budget is not interrupted mid-candidate.

Accurate wording: discovery cardinality and the between-candidate processing loop are capped.

## 3. `store capacity` — "never reads or emits stored values or identifiers"

`CHANGELOG.md:11` / `CHANGELOG.ja.md:11`

The intended guarantee — no payload content, no event or session identifiers — holds. The sentence as written is broader than the code:

- `inspectPayloadClasses` reads and aggregates `event_metadata_projection.body_stored_bytes` (`capacity_inspector.go:109-147`)
- `inspectDBStat` reads `dbstat.name`, a SQLite object identifier, and emits it as `CapacityObject.Name` (`capacity_inspector.go:87-106`)

A privacy claim is exactly the kind a reader relies on literally. "never reads payload contents or row identifiers" states the real guarantee without overreaching.

## 4. `store payload-rehearsal` — "writes only versioned zstd shadow rows"

`CHANGELOG.md:12` / `CHANGELOG.ja.md:12`

True for payload writes, which is the point being made. Literally false as written: the workflow also writes run and checkpoint rows, migration and schema state, and the external prepared-upgrade journal.

Accurate wording: it writes encoded payload bytes only to versioned zstd shadow rows on the copied target.

## Also in scope — "rollback ... bounded" in the same sentence as 4

`CHANGELOG.md:12` / `CHANGELOG.ja.md:12` claim "migration, scrub, rollback, and WAL reservations are bounded".

The state-mutating rollback is bounded: `preparedStoreUpgradeUsecase.Rollback` derives `lockCtx` from `run.Budget.PublishLockLimit` (`application/usecase/prepared_store_upgrade.go:312-316`), and `compaction_files.go:304` rejects a rehearsal-migration run whose `PublishLockLimit` exceeds one second.

What is not bounded is the verification preceding it — `fileDigest` over the whole backup (`payload_rehearsal_target.go:378-389`, a context-unaware `io.Copy`) and `CanonicalEventAuditDigest` over the whole canonical set (`prepared_migration_verifier.go:20-34`), plus `integrity_check` on the explicit `rollback` path. `--wall-time-limit` does not reach any of it: it is a manual comparison inside the batch loop (`application/usecase/payload_rehearsal.go:110`) and a ctx timeout for the migration build (`prepared_migration_recipe.go:57`).

The underlying defect is filed separately. This issue covers only making the sentence true for v0.34.

## Not changed

Re-verified as supported, listed so a later reader does not re-audit them: rehearsal never modifies live canonical event/audit rows; rehearsal resumes at committed batch boundaries; `payload-backfill` batches are bounded and resumable (hard cap 4096 rows); compaction is resumable across swap/publish faults; dedupe never touches command audits; session resume reads a bounded window.

## Acceptance

- [ ] Each of the five statements above is either corrected or removed, in **both** languages, saying only what the code does
- [ ] No code change — this is a release-note accuracy fix
- [ ] `docs verify-i18n` green

duck8823:
Closes #1857

Documentation only. The v0.34 entries asserted more than the binary provides in nine places. A release note that claims more than the binary does is the same failure class as a surface reporting a wrong number as exact: the reader acts on it, and nothing signals that it was wrong.

Found by gpt-5.6-sol auditing the v0.34 entries after PR #1855 — five in the first pass, four more in the review of the fix, including one claim sol had endorsed in the first pass and withdrew on a second look. Every item was re-verified against the code before the edit.

## First pass

| Claim | Reality | Now says |
|---|---|---|
| capacity "never reads or emits stored values or identifiers" | aggregates `event_metadata_projection.body_stored_bytes` and emits `dbstat` object names (`capacity_inspector.go:87-147`) | "never reads or emits payload contents or row identifiers" — the guarantee that was actually intended |
| compact "bounded candidate verification" | `cmd.Context()` carries no deadline for non-hook commands (`main.go:351-368`); verification is `integrity_check`, `foreign_key_check`, schema digest, a logical digest over every table, and the codec scrub (`compaction_sqlite.go:86-154`) | "full-scan candidate verification without a configured wall-time or resource cap" |
| gc "Discovery is capped by range and wall-time budgets" | `DiscoverCandidates` runs at `orphan_consolidation_usecase_impl.go:91`, before any budget comparison; the wall budget is checked at `:112` between candidates | discovery capped by range cardinality, the between-candidate loop capped by wall time, and neither interrupts a candidate in flight |
| rehearsal "writes only versioned zstd shadow rows" | true of payload bytes — the point being made — but it also writes run, checkpoint, migration and journal state | "writes encoded payload bytes only to versioned zstd shadow rows on the copied target" |

## Second pass, on the fix itself

| Claim | Reality | Now says |
|---|---|---|
| rehearsal "state-mutating rollback is bounded by the publish lock limit" | `PublishLockLimit` wraps only the prepared-target exchange (`prepared_store_upgrade.go:312-316`), reached through `RollbackPrepared`. The artifact digests, the whole-file restore taken when no prepared target exists (`payload_rehearsal.go:986-990`), and the post-restore `integrity_check` are outside it | names the exchange as the bounded part and the three unbounded steps explicitly, pointing at #1858 |
| gc "the gc path remains the bounded fallback" | contradicted the very next sentence, which had just been corrected to say discovery is *not* wall-time capped | "the gc path remains the fallback" — the following sentence already states exactly what is capped |
| migration 000059 "The migration itself is bounded — measured at 0.096s" | one measurement on one store. Migrations run on store open (`migrate.go:35-42,135-164`) with no configured wall-time limit; `DROP INDEX` at `000059:19` has no cap | keeps the figure, drops "bounded", and states that no wall-time limit is configured on the migration path |
| dedupe "Command audits are not touched" | true today only because candidates are `client = 'hook' AND kind IN ('prompt','transcript')` (`content_event_dedupe_datasource.go:447-453`). `apply` runs a bare `DELETE FROM events` (`:801`) and `command_audits.event_id` is `ON DELETE CASCADE` (`000003_create_command_audits.sql:1-2`); bundle import and archive restore insert audits without checking the referenced event's kind | narrowed to what the candidate query guarantees, and says so — "a property of the candidate query rather than an invariant enforced on the audit table (#1862)" |

The last one is not a wording slip. A store that reaches that state loses an audit with no archive row and no way for `--restore` to bring it back, and neither `--apply`'s count nor `--restore`'s reports it. Filed as #1862 with three candidate fixes; not fixed here, because a change to the dedupe candidate query at RC time is a different kind of risk than a release-note correction.

Both languages, paired. No code change — `git diff --stat` is `CHANGELOG.md` and `CHANGELOG.ja.md` only.

## Deliberately not weakened

Each sentence still says what *is* capped. "bounded" was not deleted or softened to "generally bounded" anywhere it holds: a reader who cannot tell what is capped is no better off than one who was told the wrong thing.

## Re-verified as supported, and left alone

Listed so a later reader does not re-audit them: rehearsal never modifies live canonical event/audit rows; rehearsal resumes at committed batch boundaries; `payload-backfill` batches are bounded and resumable (hard cap 4096 rows); compaction is resumable across swap/publish faults; session resume reads a bounded window.

Out of scope: the already-shipped #1227 entry carries the same dedupe/audit wording in a released section. Correcting shipped release notes is a separate decision from getting v0.34's right.

`docs verify-i18n` green.


duck8823:
## Review record — gpt-5.6-sol, five passes

Documentation-only PR, five review rounds. sol found a defect in every one of the first four, including two claims it had itself endorsed as correct in an earlier pass and withdrew on a second look.

| Pass | Findings |
|---|---|
| 1 | 4 MUST — rollback bound stronger than the code; "bounded fallback" contradicting the corrected gc sentence; migration-059 "bounded" from a single measurement; the dedupe/command-audit cascade (sol's own endorsement withdrawn) |
| 2 | 3 MUST — pre-restore `integrity_check` missing from the unbounded list; `!preparedRolledBack` described as "no prepared target existed"; the shipped #1227 entry carrying the same false claim |
| 3 | 2 MUST — pre-exchange verification (`VerifyRollbackTarget`: integrity, foreign keys, codec scrub, canonical digest) also unbounded; the reversibility claims around the new hazard sentence contradicting it |
| 4 | 4 MUST — rehearsal-state scans unbounded too; the exchange deadline not observed by an in-flight `fsync`; the #1227 title still claiming "Reversible"; "event rows" implying a byte-for-byte restore the archive cannot give |
| 5 | **Clean. Merge OK.** |

Three omissions from the same list across three consecutive passes is what changed the approach. Enumerating what is *not* capped is only ever as good as its last audit, so the rollback sentence now states the shape — `rollback` as a whole is unbounded, with the one deadline that does exist named and its limit stated — instead of listing steps.

Every finding was verified against the code before being accepted. Nothing was taken on sol's word, and one framing of its own was corrected back (the `!preparedRolledBack` condition was checked against `payload_rehearsal.go:986-990` rather than restated).

Filed, not fixed here: #1858 (rollback verification budget), #1862 (dedupe cascade-deletes a command audit).

Issue #1865, opened 2026-08-12
v0.35: sessions --snapshot still costs 0.25s per active session after #1863


duck8823:
## Summary

After #1863 removed the superlinear term, `traceary sessions --snapshot` completes on a real store — but it still takes **201 seconds** at the default `--limit 500`. That is a surface v0.34 promotes as the replacement for the deprecated cockpit, so three and a half minutes is not a comfortable resting place.

## Measured

Store: 32.45 GB, 657,694 events, 27,540 sessions, 485 of them active. Same binary, warm.

| `--limit` | before #1863 | after #1863 |
|---|---|---|
| 100 | 32.2 s | 26.3 s |
| 200 | 642.7 s (10 m 43 s) | 98.9 s |
| 500 (default) | > 30 min, killed | **201 s** |

The remaining curve is linear: roughly **0.25 s per active session** after a ~3 s fixed cost.

## Where the remaining time goes

Two costs per distinct family, neither addressed by #1863:

1. **The recursive family query.** `infrastructure/sqlite/sql/session_lineage.sql` climbs to the topmost ancestor and then expands every descendant of that root, joining `events` for per-session aggregates. One execution per family.

2. **Latest-event body hydration per returned row.** `scanSessionSummary` (`infrastructure/sqlite/session_datasource.go:803-808`) calls `loadEventPlaintext` for each row of each family. On a large store these are SQLite overflow-page reads — the frames the original #1863 stack was sitting in. This hydration is genuinely needed for rows that are rendered, but it is currently paid for every row of every family, including rows the tree filter later drops.

Note that `Database.open` returns a shared read-only handle when one is set (`infrastructure/sqlite/database.go:262-266`), so the per-call open that #1863 mentioned is not obviously the dominant term. That should be measured rather than assumed before anything is changed.

## Directions, none of them chosen yet

- Hydrate latest-event bodies **after** the tree filter, for the rows that survive into the rendered snapshot, rather than during expansion.
- Fetch the families for all roots in one query instead of one query per root.
- Measure first: attribute the 0.25 s between the recursive query, the event aggregate join, and the body reads. Any of the above is guesswork until that split is known.

## Not urgent for v0.34

#1863 is the release blocker and it is fixed: the command returns, and its output is byte-identical to the old path where the old path finished at all. This issue is the follow-up on cost, deferred to v0.35.

duck8823:
Closing as obsolete rather than fixed.

[...]

PR #1866, opened 2026-08-12
fix: return from sessions --snapshot on a large store

### LINKED ISSUE #1863: v0.34.0: sessions --snapshot does not complete on a real store
## Summary

`traceary sessions --snapshot` does not complete on the maintainer's live store. Two runs were killed after **30 minutes** and **7 minutes** without producing output. This is the surface v0.34 tells operators to move to: `traceary` bare cockpit, `top` and the live `sessions` dashboard are all deprecated in v0.34 with `sessions --snapshot` named as the replacement.

Found by dogfooding the v0.34.0 release candidate (`3073a6de`) against `~/.config/traceary/traceary.db` (32.45 GB, 657,694 events, 27,540 sessions, schema 59).

## Measured

Same store, same binary, warm:

| `--limit` | elapsed |
|---|---|
| 1 | 3.0 s |
| 5 | 4.7 s |
| 20 | 7.9 s |
| 50 | 20.1 s |
| 100 | 30.5 s |
| 200 | **> 7 min, killed** |
| 500 (default) | **> 30 min, killed** |

The cost is superlinear in the number of active sessions, not linear.

## Where the time goes

`SIGQUIT` stack from the hung default run, innermost traceary frames outward:

```
modernc.org/sqlite/lib._getOverflowPage
modernc.org/sqlite/lib._accessPayload
...
infrastructure/sqlite.(*storeLeaseConn).QueryContext
infrastructure/sqlite.(*SessionDatasource).LineageOf
application/usecase.(*sessionUsecase).Lineage
presentation/cli.(*topDataLoader).expandSessionLineages    <- 485 summaries
presentation/cli.(*topDataLoader).loadSessionsWithReliability
presentation/cli.(*topDataLoader).loadSnapshot
presentation/cli.(*RootCLI).loadTopSnapshot
presentation/cli.(*RootCLI).runTopNamed
presentation/cli.(*RootCLI).runSessions
```

Three costs compound:

1. **One `Lineage` call per active session.** `expandSessionLineages` (`presentation/cli/top_data.go:401-424`) loops over every summary and calls `l.session.Lineage(ctx, summary.SessionID())`.

2. **`LineageOf` opens the store per call** (`infrastructure/sqlite/session_datasource.go:720-728`), so 485 sessions means 485 opens.

3. **Every row of every family is body-hydrated, then most of the work is thrown away.** `LineageOf` returns the *whole family* rooted at the session's topmost ancestor, and `scanSessionSummary` calls `loadEventPlaintext(ctx, q, latestEventID)` for each returned row (`session_datasource.go:803-808`) — the overflow-page reads visible in the stack. `expandSessionLineages` only deduplicates **after** that, against its `seen` map. So for a family of k members that are all active, the family is queried and hydrated k times and k-1 of those results are discarded. That is the superlinear term.

## Regression status

The per-row `loadEventPlaintext` hydration inside `scanSessionSummary` was introduced by `47d3a238` ("fix: unify payload metadata hydration", 2026-08-01), which is **not** an ancestor of `v0.33.1` — it is new in v0.34. The N+1 lineage loop itself predates it (`017d2bf0`, "refactor(cli): extract top data loader").

Not yet verified: whether v0.33.1 completes the same snapshot on the same corpus. The store is now at schema 59 and cannot be read by 0.33.0, so a comparison needs a separate copy of the pre-upgrade backup.

## Why this blocks v0.34

v0.34 deprecates the bare-`traceary` cockpit, `top`, and the live `sessions` dashboard, and directs operators to `sessions --snapshot` in the help text, the deprecation notices and the docs. Shipping that redirection while the target surface does not return on a real store hands the operator a command that appears to hang.

## Options

1. Deduplicate **before** hydrating: resolve the set of lineage roots first, fetch each family once, and drop the per-summary loop.
2. Do not hydrate `latest_event` bodies during lineage expansion — the expansion only needs identity and parentage; the body is needed for the rows that are finally rendered.
3. Reuse one open connection across the expansion instead of opening per `LineageOf` call.

(1) and (2) are independent and either alone should remove the superlinear term; (3) removes a constant factor.

## Reproduction

```sh
traceary sessions --snapshot            # default --limit 500
traceary sessions --snapshot --limit 100   # ~30 s, for comparison
```

Requires a store with several hundred active sessions in overlapping lineages.

duck8823:
Closes #1863

## What was wrong

`traceary sessions --snapshot` did not return on the maintainer's live store. Two runs were killed at 30 and 10 minutes. This is the surface v0.34 tells operators to move to, since the bare cockpit, `top` and the live `sessions` dashboard are all deprecated in this release and name `sessions --snapshot` as the replacement.

Found by dogfooding the v0.34.0 candidate against `~/.config/traceary/traceary.db` (32.45 GB, 657,694 events, 27,540 sessions, 485 of them active), not by a test.

## Root cause

Located from a `SIGQUIT` goroutine dump:

```
modernc.org/sqlite/lib._getOverflowPage
modernc.org/sqlite/lib._accessPayload
infrastructure/sqlite.(*SessionDatasource).LineageOf
application/usecase.(*sessionUsecase).Lineage
presentation/cli.(*topDataLoader).expandSessionLineages   <- 485 summaries
presentation/cli.(*RootCLI).runSessions
```

`LineageOf` returns the family rooted at the session's topmost ancestor, and `scanSessionSummary` hydrates each returned row's latest-event body through `loadEventPlaintext` — the overflow-page reads in the stack. `expandSessionLineages` deduplicated only *after* that work. So a family with k active members was queried and fully body-hydrated k times, and k-1 of those results were thrown away.

## The change, and the correction it needed

The first version of this PR was three lines: skip a session an earlier family already collected. I argued it was output-preserving because a session in another session's family shares its root and therefore its family.

**gpt-5.6-sol showed that argument was false, with a counterexample I then reproduced.** `sql/session_lineage.sql` capped the ancestor climb at `depth < 100`, so for a chain longer than that, `lineage_root` is not the true root but whatever the climb stopped at — a value that depends on *which member you asked about*. For a 102-node chain, `Lineage(n101) = {n001…n101}` and `Lineage(n100) = {n000…n100}`: `n100` is in the first set, yet the two sets differ. The skip would have dropped `n000` — a silently missing session row.

So the fix now has two parts:

1. **`sql/session_lineage.sql`** — the `depth < 100` clause is removed from the `ancestors` CTE. The `instr(path, …) = 0` predicate already terminates cycles, so the cap was not what made the recursion finite; it was what made the result depend on which family member asked. Removing it makes `LineageOf` return the family of the *true* topmost ancestor for every member, which is what its doc comment at `session_datasource.go:717-718` already claimed. The descendant cap is untouched: it truncates identically for every member of a family, so it does not break the invariant.

2. **`sql/session_lineage.sql` again** — a second round of review found the same invariant broken a second way. `SessionIDFrom` permits a comma in a session ID, and the visited-path guard was a substring test over raw comma-joined IDs, so `,root,` matched inside an ID like `leaf,root` and cut the climb short. Every ID entering the path is now `hex()`-encoded, which is injective and emits no delimiter, making the segment test exact.

3. **`expandSessionLineages`** — the skip, unchanged from the first version, now resting on an invariant that actually holds.

What the fix does **not** claim: the returned set is still subject to the descendant depth cap, so a family deeper than 100 comes back truncated and a member below the cap is absent from its own lineage. That is pre-existing — main truncates identically — and it is filed as #1867. It does not affect the skip: the truncated set is the same for every member, and a member outside it is never marked seen.

Connection reuse was considered and left out: `Database.open` returns a shared read-only handle when one is set, so the per-call open is not obviously dominant, and forcing reuse would change the datasource contract. That and the remaining per-family cost are #1865.

## Measured on the live store

| `--limit` | before | after |
|---|---|---|
| 100 | 32.2 s | 26.3 s |
| 200 | 642.7 s (10 m 43 s) | 98.9 s |
| 500 (default) | did not return in 30 min | **201 s** |

Output equivalence was re-taken after the SQL change against a **frozen 32.49 GB copy** of the store with no writer attached, running baseline → branch → baseline so the baseline's own drift is measured: the branch's output is **byte-identical (53,834 bytes) to the adjacent baseline run**. The earlier run pair differs only in one wall-clock-derived `avg_age` field, and the two baseline runs differ in that same field the same way. Details in a comment.

## Tests

- **`infrastructure/sqlite`** — a 150-session chain over a real temporary store asserts that `LineageOf(deepest)` and `LineageOf(deepest's parent)` return the same set and that it contains the true root. I ran this against the pre-change SQL and confirmed it fails there, with exactly sol's counterexample in the diff (`deep-048` present on one side, `deep-149` on the other). The cycle case is kept, since the path guard is now the only thing terminating the recursion.
- **`infrastructure/sqlite`** — comma-containing IDs, both at the end of an ID and in the middle with a further descendant, asserting the same-set property. Confirmed to fail against the pre-hex SQL with `root` missing from the shorter climb's result.
- **`presentation/cli`** — a table-driven internal test pins the `Lineage` **call count**, not just the merged output; an output-only assertion would have passed against the old, slow code.

`go build ./...`, `go test ./...`, `go tool golangci-lint run` (0 issues) and `docs verify-i18n` all pass, run independently rather than taken from the implementer's report.

## Attribution

Implementation by gpt-5.6-luna. Review by gpt-5.6-sol, whose counterexample changed the fix. Live-store measurement, counterexample reproduction and final review by Claude Opus 5.


duck8823:
## Review record — gpt-5.6-sol (round 1): REQUEST CHANGES, and it was right

sol reviewed the first version of this branch and rejected it. I reproduced its counterexample against the SQL before acting on it, and the fix changed as a result. Recording the findings and what each one led to.

### MUST 1 — the skip silently dropped a session (accepted, fixed)

> `session_lineage.sql:10,26` caps both the ancestor climb and the descendant expansion at depth 100. For a 102-node chain `n000 → … → n101`: `Lineage(n101) = {n001 … n101}`, `Lineage(n100) = {n000 … n100}`. `n100 ∈ Lineage(n101)`, yet the two sets are not equal. With active input `[n101, n100]` the old code's second call contributed `n000`; the new code skips `n100` as already seen and drops `n000`.

Confirmed by reading the SQL: both recursions carried `WHERE … depth < 100`. My "output-preserving because it shares its root and therefore its family" argument assumed `LineageOf` returns the family of the true topmost ancestor, which the ancestor cap made false. This is a silent false negative — a session row that exists and stops being rendered — which is the failure class this project treats as most serious, so it is fixed at the cause rather than worked around: commit `4e7e0a57` removes the cap from the `ancestors` CTE, and `TestDatasource_LineageOfDeepChainUsesTrueRootForEveryMember` builds a 150-session chain over a real store and pins the invariant. I ran that test against the pre-change SQL and it fails there with sol's counterexample in the diff.

Worth being explicit: the maintainer's live store has a maximum chain depth of **1**. The case was unreachable on the data I measured. It is fixed anyway, because the argument the skip rests on has to hold in principle, not on one store's shape.

### MUST 2 — merged order could change (accepted; resolved by the same fix)

> With input `[n101, n100, unrelated, n000]` the old code added `n000` during the second call, before `unrelated`; the new code adds it after.

This followed from the same broken invariant. Once `LineageOf` returns the whole family for any member, the first call emits the entire family, so a later member contributes nothing and the first-occurrence order that drives `merged` is unchanged. Re-checked against the counterexample after the SQL change: the orders match.

### MUST 3 — the changelog asserted a guarantee that did not hold (accepted, fixed)

Rewritten in `1bad3a25` / `d4051c9c` / `5abb7f02`. The entry now names the depth cap as the reason the invariant did not hold, says removing it establishes the invariant for the first time rather than restoring it, and only then calls the skip safe.

### SHOULD 1 — the stub test assumed the premise (accepted, fixed)

The CLI test pins the `Lineage` call count and does fail against the old code, so it is a valid performance regression test, but its stub assumed same-family-same-result. The datasource test above supplies what it could not. The cycle case is kept, since the path guard is now the only thing terminating the ancestor recursion.

### SHOULD 2 — "linear in active sessions, about 0.25 s each" (accepted, removed)

Correct that the call count now tracks distinct families rather than active sessions, and that per-call cost varies with family size and body sizes. The rate and the linearity claim are gone; the three measured points stand as measurements.

### Concurrent writes — noted, not treated as a regression

sol also observed that a descendant added between call 1 and call k would only be picked up by a later call, which the skip now removes. That window is not created here: the summary list this expansion iterates is fetched once before the loop, so a session appearing mid-scan is already outside the snapshot. Fewer queries narrows the window rather than opening a new one. If per-snapshot consistency should be a guarantee it needs a single statement or a transaction, which is a separate question from this fix — #1865 is where the remaining query shape is tracked.

### Process note

sol reports that the read-only sandbox refused Go's temporary build directory, so it ran the targeted tests once outside the sandbox. No repository changes; the branch is unmodified by the review.

duck8823:
## Output equivalence, re-taken against a frozen store

The earlier byte-identical result was taken before the SQL change, and a repeat on the live store differed — inspection showed a concurrent codex session writing to it (newer event timestamps, `large_payloads` count 78 → 79), which is drift, not semantics. To remove the ambiguity I took a `.backup` of the live store to a standalone 32.49 GB file with no writer attached and ran the comparison there.

Run order was **baseline, branch, baseline**, so the baseline's own drift is measured rather than assumed:

| run | binary | elapsed | bytes |
|---|---|---|---|
| 1 | `traceary-rc` (main `3073a6de`) | 48 s | 53,834 |
| 2 | branch (`5abb7f02`) | 33 s | 53,834 |
| 3 | `traceary-rc` (main `3073a6de`) | 31 s | 53,834 |

Command: `sessions --snapshot --limit 100 --db-path <frozen copy>`, 264 lines each.

- **Run 3 vs run 2 — byte-identical**, all 53,834 bytes. These are the two adjacent runs, so nothing clock-derived separates them.
- Run 1 vs run 2 differ on exactly one field, `avg_age=27h13m` vs `27h14m`, in the `candidate_age` diagnostic — computed from wall-clock now. **Run 1 vs run 3, both baseline, differ on that same field in the same way**, so it is the clock advancing between run 1 and the later pair, not the change. Excluding that field, run 1 vs run 2 is identical too.

Run 1's 48 s is a cold page cache on a freshly written file; the 31 s baseline in run 3 is the fair comparison against the branch's 33 s. At `--limit 100` this store has few enough duplicate families for the saving to be inside the noise, which matches the live-store table — the gain shows up at `--limit 200` and at the default limit, where the duplicate count is high enough to matter.

duck8823:
## Review record — gpt-5.6-sol (round 2): REQUEST CHANGES

One finding accepted and being fixed, one accepted as a defect but rejected as a property of this change, one accepted in part.

### MUST 2 — comma in a session ID defeats the path guard (accepted, fix in progress)

> `SessionIDFrom` permits commas. For `root` and a child literally named `leaf,root`, the ancestor path for `leaf,root` is `,leaf,root,`, and searching it for `,root,` falsely reports `root` as already visited. `LineageOf("leaf,root") => leaf,root`, `LineageOf("root") => root | leaf,root`.

Reproduced in `sqlite3`. `domain/types/session_id.go:13-19` only trims and rejects empty, so this is a permitted ID. The same-family invariant this PR is built on therefore still does not hold, which means the skip can still drop rows — a member whose climb is cut short by the false positive resolves to a root *below* the true root, while a sibling branch without the poisoned ID in its path climbs higher, so the sibling's family is strictly larger and skipping it loses the rows above.

Being fixed by hex-encoding every ID that enters the path, in both CTEs, anchor and recursive terms. `hex()` is injective and emits only `0-9A-F`, so no encoded ID can contain the delimiter and the segment test becomes exact. Verified before delegating:

```
climb from "leaf,root" with hex paths  ->  leaf,root | root
```

A regression test with comma-containing IDs comes with it, confirmed to fail against the pre-hex SQL.

### MUST 1 — the descendant cap (accepted as a defect, filed as #1867; rejected as a property of this change)

The defect is real and I filed it: `LineageOf("deep-149")` on a 150-chain returns `deep-000 … deep-100`, so the queried session is absent from its own lineage, and `top_data.go:415-417` only substitutes the summary when the lineage is **empty** — so a truncated-but-non-empty result drops the active row from the snapshot. That is a silent false negative and it should be fixed. It is **#1867**, not this PR: the descendant cap is untouched here, and main drops the same row for the same reason.

Two claims attached to it I do not accept:

> The ordering argument is not sound: the first call does not emit the whole family.

The ordering argument does not depend on the family being complete — it depends on it being the **same** for every member. `Lineage(X) = descendants(root(X), depth ≤ 100)`. If `B ∈ Lineage(A)` then `B` is a descendant of `root(A)`, and with the climb uncapped `root(B) = root(A)`, so `Lineage(B) = Lineage(A)` exactly — the same truncated set, not a subset. The first call emits it in full, later members contribute nothing, and first-occurrence order is unchanged. This is what round 1 said: *"every member of a family reaches the same root, so the downward expansion is identical."* That still holds; the cap truncates identically for everyone.

> Those active summaries are never added to `seen`, so their lineage queries are still repeated.

Correct, and it is why the cap costs nothing in correctness here: a session outside the truncation is never marked seen, so it is never skipped and never dropped by this change. It is a missed saving on deep families, not a lost row.

> The test establishes the truncated result as expected behavior rather than detecting the false negative.

Fair as a description of what the test pins. It pins the property this PR needs — same set for two members of one family — and #1867 is where the set's completeness gets decided, with that test named as the place to extend. I would rather not encode a guess at the intended semantics here: option 1 in #1867 (always include the queried session) and option 3 (drop the cap) give different sets, and choosing between them is a behaviour decision with a cost question attached, not a fix to this regression.

### MUST 3 — the CHANGELOG claims more than the diff supports (accepted)

Correct on both counts. "Returns the whole family rooted at the topmost ancestor" is false for families deeper than the descendant cap, and naming only the ancestor cap as what broke the invariant was incomplete — the comma-collidable guard broke it too. Both entries are being rewritten to claim only that every member resolves to the same root and therefore the same lineage set, with #1867 named for the truncation.

### Checks I am taking as confirmed

Self-parent and ordinary cycles terminating on the exact-ID case, missing parents terminating naturally, upward diamonds being impossible under a single `parent_session_id`, and the concurrent-write argument holding for a pre-fetched candidate set. The quadratic path/`instr` growth on long chains is noted; it stays with #1865.

duck8823:
## MUST 2 fixed — `de4276fa`, verified

`session_lineage.sql` now hex-encodes every ID that enters the visited-path, in both CTEs, anchor and recursive terms, path construction and `instr` test alike. `hex()` is injective and emits only `0-9A-F`, so no encoded ID can contain the delimiter and the segment test is exact.

`TestDatasource_LineageOfCommaIDsUsesTrueRootForEveryMember` covers both shapes: a comma at the end of a child's ID (`leaf,root` under `root`), and a comma in the middle with a further descendant (`a,root,b` under `root`, with `deep-child` under it) — the shape that lets the skip lose rows, since the poisoned climb resolves to a root below the true one while a sibling branch climbs higher. Each case asserts every member returns the same set and that the set contains the true root.

Run against the pre-hex SQL, it fails with sol's counterexample exactly:

```
--- FAIL: .../comma_in_leaf_ID
    LineageOf("leaf,root") IDs mismatch (-want +got):
      []string{
        "leaf,root",
    -   "root",
      }
--- FAIL: .../comma_in_middle_ID_with_descendant
    LineageOf("a,root,b") IDs mismatch (-want +got):
      []string{
        "a,root,b",
        "deep-child",
    -   "root",
      }
```

`go build ./...`, `go test ./...` (exit 0), `go tool golangci-lint run` (0 issues) and `docs verify-i18n` all pass, run here rather than taken from the implementer's report.

**Output equivalence re-taken** on the frozen store after the hex change, branch then baseline back to back: 52,799 bytes and 261 lines each, identical except the wall-clock `avg_age` field the two runs straddled a minute boundary on. As expected — the maintainer's store has **0** session IDs containing a comma or a percent sign, so hex encoding cannot change what it returns; the point of the change is that nothing prevents such an ID from existing.

Both CHANGELOG entries now say the invariant was broken by **both** causes — the ancestor depth cap and the comma-collidable raw-ID guard — that removing the cap and hex-encoding establish it, and that the returned set is still subject to the descendant cap, with #1867 named for that limitation.

One deviation from the packet: the SQL change and the changelog correction landed in a single commit rather than two. The diff is small enough to read either way; noting it rather than rewriting history on a branch under review.

duck8823:
Closing without merging.

The owner's decision on 2026-08-12 is that the session-graph browsing surface goes as a category — `sessions` and its snapshot, `session tree`, and the MCP `session_status` lineage/tree actions — because always-on browsing serves neither of the two pillars (`.ai/spec/1606.md` D13). That is #1869, and it closes #1863 along with #1865, #1867 and #1868.

This PR optimises and repairs `session_lineage.sql` and `expandSessionLineages`, which are exactly the machinery being removed. Merging it would ship correct code for a surface scheduled to disappear, and its remaining open finding — the cycle counterexample below — would cost another round to close.

The review is worth keeping on the record, because the defects were pre-existing and are properties of the query, not of this change:

1. The ancestor climb's `depth < 100` cap made `LineageOf` return a different family depending on which member was asked.
2. The visited-path guard compared raw comma-joined IDs, and `SessionIDFrom` permits a comma, so `,root,` matched inside `leaf,root` and cut the climb short.
3. `hex()`, used to fix (2), is not injective over SQLite storage classes: `hex(TEXT 'A') = hex(BLOB x'41')`.
4. The descendant cap combined with a parent cycle makes a session absent from its own lineage, and `top_data.go:415-417` only substitutes the summary when the lineage is empty — so the snapshot silently drops an active row.

All four were found by dogfooding v0.34 against a 32 GB store and by gpt-5.6-sol across three review rounds, and all four disappear with the surface. None of them is on the path from "the agent recorded this" to "the agent gets the summary back".

The branch `fix/1863-sessions-snapshot-lineage` is left in place; the measurements and counterexamples are in the comments above if any of this needs to be revisited.

Issue #1871, opened 2026-08-12
v0.35.0: retire the MCP server — 16 of 659,304 events came through it


duck8823:
Part of #1870.

## Proposal

Remove the MCP server and its 9 tools. `traceary mcp-server` goes with them.

## What the store says

`events.client` identifies writes that came through the MCP server.

| | |
|---|---:|
| events written via MCP | **16** |
| total events | 659,304 |
| share | **0.0024%** |
| last MCP write | **2026-07-19** |

Memory provenance is the same story. Of 37,329 memories: `extracted` 29,141, `extracted-hidden` 7,446, `remember-intent` 539, `compact-summary` 186, `manual` 17. None came from MCP.

The write half of the MCP surface — `record_event`, `manage_session`, `manage_memory` — is unused because hooks do the writing: 99.99% of recording is hook-driven, and every extracted memory is. Nothing in either pillar routes through MCP.

**Reads are not measured.** Nothing records an MCP read, so `get_context`, `search`, `list_events`, `get_report`, `query_memory` and `session_status` could be in use and leave no trace. This issue does not claim otherwise.

## Why the unmeasured read half does not save it

Every host Traceary ships an integration for has a shell: `integrations/claude-plugin`, `gemini-extension`, `codex`, `kimi-plugin`, `grok-plugin`, `antigravity-plugin`. There is no shipped integration for a shell-less host such as Claude Desktop or a web client. Anything an agent can ask through `get_context` or `search`, it can ask through `traceary context` or `traceary search` in the same session.

What is genuinely lost is the typed schema: MCP tools declare their inputs and outputs, so the model does not have to construct a command line or parse text. Against that stands the cost of every session carrying 9 tool schemas in its context, and of every CLI change needing a mirrored MCP change and a second set of tests.

## What must be checked before removing

1. `integrations/gemini-extension/gemini-extension.json:5-9` and `integrations/antigravity-plugin/mcp_config.json:2-6` both declare `traceary mcp-server` as an MCP server. Confirm neither host's hook wiring depends on the server being present — if a hook path reaches Traceary only through MCP on those hosts, that path needs a shell equivalent first.
2. `integrations/claude-plugin/hooks/hooks.json:83,145` match on `mcp__.*` — that matcher observes *other* servers' tool calls and is unrelated to Traceary's own server. Confirm this reading before touching it.
[...]

PR #1887, opened 2026-08-13
Make search --json the events/sessions object

### LINKED ISSUE #1775: v0.35.0: make search --json the announced events/sessions object
v0.34.0 promised this change and shipped the notice for it, but no issue owns the change itself. Found while closing out Wave 4.

## The promise already shipped

`docs/cli-stability.md` states it twice — once in the "Currently deprecated" registry:

> `traceary search --json` top-level array → `{"events": [...], "sessions": [...]}` object (replaced in v0.35.0; announced in v0.34.0)

and once in the contract-change section, which says v0.34.x keeps the array and emits a stderr notice naming the count of omitted session hits and the new shape. That notice is live: `presentation/cli/search_output.go:33` `warnSearchSessionsOmittedFromJSON`, with the EN/JA text at `:19-28`, fired unconditionally whenever session hits were dropped.

So a v0.34 user is being told, in their terminal, that `search --json` becomes an object in v0.35. Nothing was scheduled to make that true. The reference in the docs is to #1717, which announced it and is closed.

## Why the shape has to change

#1717 made the bounded search projection the authoritative read path, which introduced a second class of result: session-tier hits for history older than the recent window. That class cannot live inside a top-level array of events without interleaving session rows among event rows, so JSON callers currently receive an event-only array plus a stderr line telling them results are missing. Text output already shows both groups.

## Scope

- `search --json` emits `{"events": [...], "sessions": [...]}`; both keys always present, empty arrays when a tier has no hits — so consumers never branch on key existence
- remove `warnSearchSessionsOmittedFromJSON` and its notice text; the shape now carries the information
- `--json-fields` behaviour against the new envelope needs an explicit decision, not a default: today it selects event fields
- update the contract snapshot tests, the CLI reference pair, the CHANGELOG pair, and move the registry entry to the historical removal log
- check the MCP `search` tool for the same split — #1727 is already asking whether it should get the session tier the CLI got. If both land in v0.35 they must agree on one envelope.

This is a **breaking JSON contract change**, which is why it waits for a minor boundary. `docs/cli-stability.md` freezes the contract for the notice window, and v0.35 is the end of that window.

Parent: #1693

duck8823:
## Summary

Completes the v0.34.0 announcement (#1717) for issue #1775:

- `traceary search --json` now always emits `{"events":[...],"sessions":[...]}`
- Both keys are always present (empty arrays when a tier has no hits)
- Explicit `--fields` still selects event fields inside `.events`; session objects keep `session_id`, `summary`, `event_count`, `started_at`
- Removes `warnSearchSessionsOmittedFromJSON` and its EN/JA notice
- Text search output is unchanged except that deprecation notice path is gone
- Contract goldens, CLI reference, CHANGELOG, and `docs/cli-stability` registry updated (moved from "Currently deprecated" to the historical removal log)

## MCP note

MCP `search` already returns an object envelope with `events` and a `sessions` field (#1727). It also carries coverage/continuation/reasons fields that are MCP-specific, so this PR does not reshape MCP. Full MCP retirement remains #1871.

## Test plan

- [x] `go test ./presentation/cli -count=1 -run 'Search|GoldenJSON|MetadataProjection'`
- [x] `go tool golangci-lint run ./presentation/cli/...`
- [x] Contract goldens updated via real CLI invoke path (no hardcoded envelope that skips the writer)
- [ ] CI on the draft PR

Closes #1775


duck8823:
## AI review

Codex gpt-5.6-sol: APPROVE. CRITICAL/HIGH empty. search --json is the events/sessions object.

**総合判定:** APPROVE

Issue #1693, opened 2026-08-07
v0.35.0: finish the cleanup


duck8823:
# v0.35.0 — finish the cleanup

v0.34.0 (#1606) narrows Traceary to two pillars: **記録** (capture, summarise, compress, evict) and **記憶** (consolidate and supply automatically). v0.35.0 finishes what v0.34 could only give notice of, and lands the pieces deliberately deferred to keep v0.34 small.

## Removals (notice shipped in v0.34, one full minor of overlap per `docs/cli-stability.md`)

- #1690 remove `traceary top` — the permanent-alias promise is withdrawn in #1688
- #1764 remove the cockpit (`tui` / `dashboard`) and make bare `traceary` print help in a TTY too — notice in #1687
- #1766 remove the `sessions` live dashboard, leaving `sessions` a plain text command — notice in #1765
- #1691 remove the commands whose backing table has never held a row (#1689): `memory admin graph`, `session label`

**Wave 4 spike, 2026-08-10 — two entries left this list.** `list --failures` is not dead: the predicate is `ca.failed = 1 OR non-zero exit code`, and `failed = 1` matches 1,943 rows still being written today (#1767). `store workspace-alias` was pulled because the alias mechanism is live — it reclassifies stored conflicts, suppresses reviewed-away pairs, and promotes future observations to `explicit_alias` — so no "nothing is lost" evidence can be produced today; its 14,684 conflicts turn out to be **147** distinct `(session_id, workspace)` pairs, so the remedy is reviewable, just unreviewed (#1768).

> The original wording said `workspace-alias` "cannot be deprecated at all" because the flow requires naming a canonical replacement. **Wave 4 invalidated that reason itself**: #1689 added the no-replacement notice form, and `docs/cli-stability.md` now permits a notice with no successor when the entry states the evidence that nothing is lost. Every retirement decision in this milestone is tested against **"can we show nothing is lost"**, not **"is there a successor."** Recorded as D15 in `.ai/spec/1606.md`.

After #1764 and #1766, `memory inbox review` is the only interactive surface left.

## Contract changes announced in v0.34

- #1775 make `search --json` the `{"events": [...], "sessions": [...]}` object it was announced as — v0.34 already emits the stderr notice promising it (`presentation/cli/search_output.go:33`), so this is a shipped promise with a deadline, not an option

[...]

PR #1888, opened 2026-08-13
fix: remove the commands deprecated in v0.34

### LINKED ISSUE #1691: v0.35.0: remove the commands deprecated in v0.34

duck8823:
## Summary
- Remove `traceary memory admin graph add` / `graph list` (no replacement; zero reference-store edges).
- Remove the whole session-label CLI surface: `session label`, `session list --label`, the `LABEL` text column, and the `label` JSON field on `session list`.
- Move both entries from Currently deprecated to the historical removal log in `docs/cli-stability.md` (+ Japanese pair).
- Update CHANGELOG and CLI reference pairs; removed paths now fail as unknown command/flag without a `DEPRECATED` notice.

## Non-goals
- Does not drop the `memory_edges` table (still used by gc and bundle).
- Does not drop the `sessions.label` column (CLI hide only).
- Does not touch `list --failures` (#1767) or `store workspace-alias` (#1768).
- Leaves parent #1693 open (implementation PRs must not complete the parent).

## Test plan
- [x] `go test ./presentation/cli`
- [x] `go tool golangci-lint run ./presentation/cli/ .`
- [x] `go build ./...`

Closes #1691


duck8823:
## 🤖 AI コードレビュー結果

### レビュアー
- Codex gpt-5.6-sol round 1: REQUEST_CHANGES — tests accepted any error on retired `graph` / `label`
- Codex gpt-5.6-sol round 2 (SHA ad36a5e34): ✅ APPROVE（CRITICAL/HIGH なし）
- Grok 4.6 orchestrator: ✅ CI Test/Lint/Docs/Integrations SUCCESS; MERGEABLE

### 🔴 Critical Issues
- なし

### 🟠 Major Issues
- なし（round 1 HIGH は `unknown subcommand "graph"|"label"` / `unknown flag: --label` の明示アサーションで解消）

**総合判定:** APPROVE

Issue #1872, opened 2026-08-12
v0.35.0: fold store size reduction into compact and delete its 30 operator commands


duck8823:
Part of #1870.

## The situation

The store grows **+9.9 GiB/month** (`.ai/spec/1606.md`, 2026-08-07; the reference store is now 32.45 GB). Five mechanisms exist to hold that down and **all five are off by default**:

| mechanism | what it removes | how it runs today |
|---|---|---|
| dedupe | exact-duplicate bodies — **2.91 GiB of the 5.05 GiB body total, 57.7%** | `store dedupe content-events --apply`, by hand |
| gc | retained bodies past their limit | `store gc`, by hand |
| retention | raw bodies, against a reviewed plan | `plan` then `apply`, by hand — **`apply` has 0 recorded invocations** |
| payload | body encoding / compression | `payload-backfill`, `payload-rehearsal` — **all 4 backfill leaves have 0 invocations** |
| search-projection | the index family — **15.83 GiB, 64.1% of the store** | `start` / `resume`, by hand |

30 of the 97 CLI commands exist to operate these. In 2.5 months of continuous use, the sequence has never been completed.

## Why it is never completed

**None of the first four shrink the file.** SQLite frees pages; it does not return them to the filesystem. `dedupe --apply` moves rows to an archive table, `gc` discards bodies, `retention apply` deletes raw bodies, `payload` re-encodes — and after all of it the file is exactly as large as before.

The disk comes back only when the file is rewritten, which is `store compact`, implemented as `VACUUM INTO` (`infrastructure/sqlite/compaction_sqlite.go:62`).

So the design is **four preparation steps with no visible effect, then one payoff step**, each its own command with its own plan/apply/status, run by hand in order. Nobody gets to the end, and the measured result is a store that has never been reduced.

## The change

Owner's decisions, 2026-08-12:

1. Fold the reduction into `compact` as a filter on the copy it already performs.
2. Keep `compact` an explicit command. **When it runs is the operator's call.**
3. Warn when there is something worth reclaiming, and say so.

```
traceary store compact
  → read the old file, write a new one, and while copying:
      do not copy duplicate bodies (keep the canonical row of each group)
      do not copy bodies past their retention limit
      do not copy raw bodies excluded by the retention policy
      encode bodies as they are written
      do not rebuild retired index families
  → atomic exchange; the old file stays until the new one is accepted
```

[...]

PR #1896, opened 2026-08-13
fix: keep folded sessions wake-eligible after orphan reduction

### LINKED ISSUE #1877: v0.35.0: orphan consolidation marks correctly-folded sessions degraded
Found while dogfooding the 記録 compression path end to end (#1606). **Not biting in production today** — `store gc` has never been run to completion, so nothing has triggered it. **#1872 makes it bite**, because folding the reduction into `compact` makes this path routine. Filed as a v0.34 follow-up and a prerequisite for #1872.

## What happens

A session that is folded *correctly* — an agent writes a real summary with motivation and change — becomes **ineligible for wake injection** as soon as the reduction runs.

Measured on a controlled store (never the live one):

| step | result |
|---|---|
| agent writes a refinement covering every event that exists | generation 1, `degraded=0` |
| `session end` writes a `session_ended` event | now one event sits after `covers_to` |
| `store gc` consolidates that one-event orphan range | generation 2, **`degraded=1`** |
| rows eligible for wake injection | **0** |
| rows if `degraded` were ignored | 1 |

The stored summary is not damaged. The agent's prose is preserved verbatim and the synthesised part is clearly fenced:

```
Motivation: verify that the record pillar's compression half works end to end.
Change: recorded twelve transcript events, fired the Stop hook, and folded the
session so its bodies become discardable.

---
[degraded section: orphan events after 0c17e4be… through c1a4fb40… (agent text
above is unchanged), synthesised by gc:orphan-consolidation]
Mechanical summary (degraded=1).
…
Events: 1 total
  session_ended: 1
Commands: (none)
```

So the entire reason this summary is dropped from wake injection is a fenced footnote saying "one session_ended event occurred here".

## Why it is systematic, not a harness artifact

The agent can only fold events that exist when it folds. `session end` writes an event **after** that, by construction:

1. the Stop hook fires and asks for a fold (`requestConsolidationIfDue`)
2. the agent runs `traceary session refine --covers-to <newest event at that moment>`
3. the session later ends, writing `session_ended`
4. the reduction sees a one-event orphan tail and supersedes the refinement with `Degraded: true`

Every well-behaved session lands in the same shape. The flag that is supposed to mark "「なぜ」 was never folded" ends up set on precisely the sessions where 「なぜ」 *was* folded.

## The two halves of the mismatch

**The producer's meaning is defensible.** `orphan_consolidation_usecase_impl.go:175-178`:

> degraded=true even when the row still contains agent-authored text. The flag means the stored summary is not purely agent-written: a mixed row includes gc-synthesised prose, so consumers must not treat the whole row as agent reasoning. Seam markers in the text separate authorships.

**The consumer's meaning is different.** `list_session_wake_summaries.sql:16` filters `r.degraded = 0`, and `SessionWakeSummaryQueryService` documents itself as loading "finished (non-degraded) top-level session summaries for a workspace so the wake path can inject them".

One flag is carrying two questions:

| question | correct answer for the row above |
|---|---|
| does this row contain synthesised text? | yes |
| is there agent reasoning here worth injecting? | **yes** |

Wake injection needs the second and reads the first.

## Consequence under #1872

#1872 folds the reduction into `compact`, which the operator is expected to run regularly. At that point:

- every refined session acquires `degraded=1` on its next compaction
- wake injection silently returns nothing, for a store full of good summaries
- `degraded` also stops distinguishing `--force` (mechanical) summaries from real ones, which is exactly the signal #1872's `--force` guard depends on

That is a silent false negative on the 記憶 reuse path — the surface reports nothing for state that exists — and it is caused by two components each behaving as designed.

## Options

**(a) Do not consolidate an orphan range that contains only lifecycle events.** A `session_ended` event carries no agent reasoning, so synthesising "the why is gone for this range" about it is both untrue and destructive. Smallest change; removes the universal case.

**(b) Separate the two questions.** Keep `degraded` meaning "contains synthesised text", and give wake injection a predicate for "has agent-authored reasoning" — a second column, or eligibility derived from whether any agent-authored generation exists for the session.

(a) alone does not close it: *any* event landing after the fold (a late audit, a subagent turn) reproduces the same outcome. (b) is the correct model; (a) is worth doing regardless because consolidating a lifecycle-only range produces a summary with no information in it.

Recommend implementing (b), with (a) as a bundled simplification. Not decided here — flagging that this is a modelling decision, not a one-line fix.

## Done when

- A session whose reasoning was folded by an agent stays eligible for wake injection after the reduction runs
- A session whose reasoning was never folded stays ineligible
- Lifecycle-only orphan ranges do not produce a summary whose entire content is "one session_ended event occurred"
- `--force` in #1872 remains distinguishable from an agent-authored fold
- Regression coverage pins the shape: fold → end session → reduce → still eligible

duck8823:
## Summary

- Split wake eligibility from `degraded`. `degraded` still means the stored summary contains synthesised text; wake injection now reads `session_refinements.has_agent_reasoning`
- Migration `000060` adds the column (`NOT NULL DEFAULT 0`, 0/1 check) and backfills `CASE WHEN degraded = 0 THEN 1 ELSE 0`. Classified ConstantInPlace (one row per session)
- Agent first-writes keep `has_agent_reasoning=1` from `!degraded`. Orphan compose onto an existing fold preserves a stored 1 so mixed rows stay injectable
- Lifecycle-only orphan tails (`session_started` / `session_ended` only) advance `covers_to` without attaching a mechanical "why is gone" footnote. No existing fold means skip, not a synthesised empty summary
- Mechanical-only rows stay ineligible. `--force` in #1872 can still tell synthesised text (`degraded`) from an agent fold (`has_agent_reasoning`)

Closes #1877

Leaves parent #1693 open

## Why coverage-only instead of a bare skip

A correctly folded session always grows a `session_ended` tail after the fold. Skipping that tail without advancing coverage would leave every well-behaved session as a perpetual orphan candidate. #1872 folds reduction into `compact`, so that would starve real orphans behind a 5000-candidate cap. Coverage-only keeps text and flags, widens `covers_to`, and drops the session from discovery.

## Test plan

- [x] `go test ./...`
- [x] fold → end session → reduce → `ListEligible` still includes the session (coverage-only, no mechanical footnote)
- [x] fold → extra notes → compose → `degraded=1` and `has_agent_reasoning=1` → still eligible
- [x] mechanical-only first write stays ineligible
- [x] lifecycle-only with no existing fold does not insert a row

## Notes

- Existing rows backfill as `has_agent_reasoning = NOT degraded`. Mixed rows created by older binaries (degraded=1 after compose) will stay ineligible until an agent fold or a later compose onto a stored 1. That is the honest backfill: one row per session cannot recover a prior generation.
- Live 30 GiB store is out of this bar. Tests drive shipped functions on scratch DBs.


duck8823:
Codex gpt-5.6-sol r1: REQUEST_CHANGES

HIGH: `applyPostCompactDigest` omitted `HasAgentReasoning`. A later agent fold over a mechanical-only row stayed `has_agent_reasoning=0`, so wake injection still missed it.

Addressed in 49990979:
- supersede treats a non-degraded write as an upgrade
- post-compact digest sets `HasAgentReasoning: true`
- regression: mechanical first write → agent fold → `ListEligible`

duck8823:
Codex gpt-5.6-sol r2: APPROVE

SHA 49990979. r1 HIGH is gone. findings=[]

Issue #1870, opened 2026-08-12
v0.36-17: reduce the command surface to the two pillars (97 to 29 CLI, 9 to 0 MCP)


duck8823:
## Why

Traceary exposes **97 invocable CLI commands** and **9 MCP tools / 21 actions**. Measured against the two pillars of `.ai/spec/1606.md` — 記録 (record, then summarize / compress / retire from the index) and 記憶 (memory and its automatic reuse) — only 39 of the 97 belong to a pillar. The largest group, 37 commands under `store`, is not a pillar at all: it is the manual operation of five size-reduction mechanisms that are all off by default while the store grows +9.9 GiB/month.

Full inventory, with per-command measured invocation counts and what the measurement cannot see: `.ai/spec/command-inventory-20260812.md`.

Owner's decision, 2026-08-12: reduce the surface as a whole rather than command by command.

## Target

| | now | target |
|---|---:|---:|
| CLI leaf commands | 97 | ~~29~~ **41** (corrected 2026-08-18 — see the correction at the end and #2121) |
| MCP tools | 9 | **0** |
| Skills | 4 names x 6 host packages | **4 names**, one per job |

### CLI — the 29 that stay

(28 as originally filed, plus `memory store propose` — see the 2026-08-13 correction below.)

**記録 (7)** — `audit`, `log`, `session start`, `session end`, `session run`, `session refine`, `hooks install` (absorbs `hooks print` as `--dry-run`)

**閲覧 (5)** — `list` (absorbs `tail` as `--follow`, `timeline` as `--blocks`), `show`, `search`, `context` (absorbs `session handoff`), `report`

**記憶 (9)** — `memory search` (absorbs `memory list` as `--all`), `memory show`, `memory store propose`, `memory inbox list`, `memory inbox show`, `memory inbox accept`, `memory inbox reject`, `memory inbox review`, `memory inbox restore`

**保守 (4)** — `store backup create`, `store backup restore`, `store compact` (absorbs `plan` / `apply` / `resume` / `status`, and the dedupe / gc / retention / payload reduction — #1872), `store compact rollback`

**移送・基盤・参照 (4)** — `bundle export`, `bundle import`, `doctor` (absorbs `store capacity` and `report workspace-identity`), `session active`

`mcp-server` is not in the keep list: #1871 retires the MCP surface with it.

### What each removed group becomes

| group | count | destination |
|---|---:|---|
| `store` maintenance | 30 | folded into `compact` as a filter on its copy — #1872 |
[...]

Issue #1692, opened 2026-08-07
v0.35.0: final surface sweep against the two pillars


duck8823:
After the v0.34 removals, walk the remaining surface once and record, for each command, which pillar it serves:

1. **記録** — capture, summarise, compress, evict
2. **記憶** — consolidate and supply automatically

Baseline: 136 registered cobra commands, 22 hidden, 65 documented entries — of which `memory` alone is 28.

Anything serving neither gets an explicit decision: keep with a stated reason, or deprecate. Usage counts are **not** grounds on their own; the criteria are empty backing data, duplication, or no pillar.

Output is a table in this issue plus the resulting deprecation notices.

Parent: #1606

duck8823:
#1692 inventory (PR #1900). Shipped source: `presentation/cli/pillar_inventory.go`. Hidden hook plumbing is not listed. Criteria: empty backing, duplication, or no pillar. Usage counts are not grounds.

Visible leaves: **77**. Notice published: **1** (`memory store remember` → `memory store propose`, removal v0.36.0).

Remaining #1870 groups stay. They were never in the v0.34 registry. `sessions --snapshot` stays (v0.34 announcement). `replay` stays (only HTML export). Absorb flags do not exist, so `tail` / `timeline` / `handoff` / `hooks print` / `memory list` are not noticed.

#1870 and parent #1693 stay open.

| Command | Pillar | Reason | v0.36 notice |
|---|---|---|---|
| `traceary audit` | 記録 | persist command-audit records |  |
| `traceary log` | 記録 | persist a manual event |  |
| `traceary list` | 記録 | read recorded events; list --follow does not exist so tail is not a duplicate |  |
| `traceary search` | 記録 | search recorded events and session summaries |  |
| `traceary tail` | 記録 | stream recorded events; no absorb flag on list |  |
| `traceary timeline` | 記録 | workspace timeline of recorded work; list --blocks does not exist |  |
| `traceary show` | 記録 | show one recorded event |  |
| `traceary context` | 記録 | assemble recorded context for a session |  |
| `traceary sessions` | 記録 | sessions snapshot; v0.34 already published --snapshot as the surviving view, so the #1870 keep-list omission does not win |  |
| `traceary session start` | 記録 | open a recorded session |  |
| `traceary session end` | 記録 | close a recorded session |  |
| `traceary session run` | 記録 | run a bounded recorded session |  |
| `traceary session refine` | 記録 | store the L2 summary that makes transcript bodies discardable |  |
[...]

PR #1897, opened 2026-08-14
feat: fold store size reduction into compact

### LINKED ISSUE #1872: v0.35.0: fold store size reduction into compact and delete its 30 operator commands

duck8823:
## Summary

- `traceary store compact` now copy-filters while it rewrites: drop the retired search family, drop non-canonical duplicate bodies, discard covered discardable-age transcript bodies, encode remaining payloads, then `VACUUM INTO` and atomic exchange
- The dedicated operator commands are gone (`store gc`, `dedupe`, raw-body `retention plan|apply|restore`, `payload-rehearsal`, `payload-backfill`, `search-retire`, `compact plan|apply|resume|status`). Invocations fail as unknown commands with no `DEPRECATED` notice
- Body gate: refuse only when covered is empty and unrefined discardable-age sessions remain. A partial fold reclaims what those sessions authorize. `--force` writes mechanical summaries through the existing orphan consolidation path and names the loss of agent reasoning
- Warning: doctor plus one stderr line on non-hook commands when the store file is at least `compact.reclaim_warn_bytes` (default 1 GiB). A malformed config still warns. Hidden hook entrypoints, `compact`, and `doctor` skip the one-liner
- `store compact rollback RUN_ID` and `store search-projection` remain. Peak disk is about 3× (source + work copy + candidate)

Closes #1872

Leaves parent #1693 open

## Test plan

- [x] `go test ./...`
- [x] unrefined-only store refuses Compact and names `traceary-session-refine` / `--force`
- [x] partial fold discards the folded body and leaves the unrefined body
- [x] `store gc` / `dedupe` / `search-retire` / `payload-rehearsal` / `payload-backfill` / `compact plan` are unknown
- [x] `store compact rollback` and `store search-projection` remain; `store retention files` remains
- [ ] Codex gpt-5.6-sol review (CRITICAL/HIGH empty or addressed)

## Notes

- Live ~30 GiB store is out of this bar. Tests drive shipped Compact on scratch DBs.
- Peak space is source + `.work` + candidate. Plan reserves temporary = source size for the work copy.
- #1870 stays open after this PR. Remaining 97→29 groups (memory admin, absorbs) were never in the v0.34 deprecation registry; they belong to #1692 notices, not this deletion.


duck8823:
Codex gpt-5.6-sol r3: ✅ APPROVE

Receipt: `codex-1897-r3-review.md`

```json
{
  "reviewer": "Codex gpt-5.6-sol",
  "verdict": "APPROVE",
  "summary": "問題なし",
  "findings": []
}
```

r1 HIGH rematched in `f4123aae` (verify, --force fail-closed, doctor preview, deleted-command pointers).
r2 HIGH rematched in `ecd45324` (dedupe-plan drops, source_hook/created_at_norm, decode fail-closed).

Issue #1904, opened 2026-08-14
v0.36.0: finish v0.35 leftovers and the interface they drag


duck8823:
# v0.36.0 — leftovers from v0.35, plus the interface they drag

Cleanup continuation. No new daily product surface.

## Why this minor exists

v0.35.0 shipped the in-scope children of #1693. Nine issues titled `v0.35.0:` were left open on purpose (notice window, parked projection, measurement). Those, the one measurement they depend on (#1743), and the operator-visible interface they pull with them are this minor.

Search-projection leftovers that already had a `v0.35.0:` title (#1819, #1833) stay here. Completing the projection family is #v0.37.

## Children (titles use `v0.36-N:` so the tagged-release parent matcher stays unique)

See the planning comment after the children are retitled.

## Exit

- Every child of this parent is closed by its own implementation PR (`Closes #<child>` only).
- `memory store remember` is removed (notice already shipped; target v0.36.0).
- Parent stays OPEN until the tagged `v0.36.0` release workflow closes it.
- Do not close #v0.37 / #v0.38 / #v0.39 from this work.

## Non-goals

- The rest of the search-projection family (v0.37)
- Hook / Log capture loss, except where a leftover already owns it
- Destructive dedupe (#1777, #1862) — those moved to v0.38
- Tag / Homebrew (after this parent is ready)

duck8823:
## Planning (2026-08-14)

Recommended cut, applied. Sibling parents: #1905 (v0.37 projection), #1906 (v0.38 records), #1907 (v0.39 shelf).

Child titles are `v0.36-N:` so `release.yml` will not treat them as the `v0.36.0:` parent.

| # | Title |
|---|---|
| #1686 | v0.36-1: retire the event metadata projection, once events is actually narrow |
| #1704 | v0.36-2: fold session active into session latest --active |
| #1706 | v0.36-3: retire sessions.summary in favour of session refinements |
| #1743 | v0.36-4: measure whether a body side table is worth it, then decide |
| #1747 | v0.36-5: wire Ctrl-C into non-hook command contexts |
| #1767 | v0.36-6: establish what failed=1 means before deciding the future of list --failures |
| #1768 | v0.36-7: decide what a workspace conflict is for before touching store workspace-alias |
| #1779 | v0.36-8: decide the remaining payloadCodecIdentity call sites |
| #1809 | v0.36-9: report the operator's own storage cost instead of a global monthly claim |
[...]

PR #1908, opened 2026-08-14
feat: fold session active into session latest --active

### LINKED ISSUE #1704: v0.36-2: fold session active into session latest --active
## Problem

`traceary session active` and `traceary session latest` are two subcommands over one implementation. Both call `runSessionLatest` with the same `sessionLatestCommandInput`; `active` differs only by setting `activeOnly: true` and carrying two extra flags:

| | `latest` | `active` |
|---|---|---|
| implementation | `runSessionLatest` | `runSessionLatest` |
| `--db-path` `--client` `--agent` `--workspace` `--json` | yes | yes (duplicated wiring) |
| `activeOnly` | false | true |
| `--stale-after` `--allow-stale` | — | yes |

`presentation/cli/session.go` carries the flag wiring twice — five identical `Flags().StringVar` / `BoolVar` calls per command — plus two long-form help texts that have to be kept consistent by hand. `active`'s own help already explains itself in terms of the other one: "Unlike session latest, this only returns non-ended sessions."

That sentence is the tell. A flag that a user reads about in another command's help is a flag, not a command.

## Proposal

Keep one subcommand and express the difference as what it is.

```
traceary session latest [--active] [--stale-after <d>] [--allow-stale]
```

`--stale-after` and `--allow-stale` are meaningful only with `--active`; reject them otherwise rather than silently ignoring them.

This removes one subcommand and one copy of the shared flag block, and leaves the two stale-handling flags attached to the mode that actually uses them.

## Alternative considered

Keeping `active` and dropping `latest` does not work: `latest` is the more general query and `active` is a filter over it.

## Breaking change

`traceary session active` is released surface. v0.34 is already a surface-reduction release (#1687 removes the cockpit, #1688 withdraws the `top` permanence promise, #1689 deprecates empty-backed commands), so this belongs with that group rather than accumulating as a separate future break.

Document the old → new mapping in the release notes, and check the plugin packages under `integrations/` and `plugins/` for callers before removing.

## Acceptance

- `traceary session active` is gone; `traceary session latest --active` produces its behaviour, stale handling included
- `--stale-after` / `--allow-stale` without `--active` is rejected with a clear message
- The shared filter flags are wired once
- Any caller in `integrations/` or `plugins/` is updated in step
- Release notes carry the mapping

## Not in scope

The semantics of "latest", the stale threshold default, and the filter resolution order. This changes only how the command is spelled.

Parent: #1606

duck8823:
## Summary
- Remove `traceary session active`. Use `traceary session latest --active` (same stale defaults).
- `--stale-after` / `--allow-stale` without `--active` are rejected.
- Update skills, CLI reference, pillar inventory, and repo-tooling skill contract.

## Non-goals
- Does not change Latest()/Active() query semantics.
- Leaves parent #1904 open (implementation PRs must not complete the parent).

## Test plan
- [x] `go test ./presentation/cli ./cmd/repo-tooling`
- [x] `go tool golangci-lint run ./presentation/cli/ ./cmd/repo-tooling/`

Closes #1704


duck8823:
gpt-5.6-terra medium: ✅

```json
```

Independent review of #1908 / #1704. CI Test+Lint green. No CRITICAL/HIGH.

Leaves parent #1904 open.

PR #1909, opened 2026-08-14
fix: cancel non-hook command contexts on Ctrl-C

### LINKED ISSUE #1747: v0.36-5: wire Ctrl-C into non-hook command contexts
Follow-up from #1685 / PR #1745, found while documenting the payload backfill cancellation contract.

## What is true today

`commandContext` (`main.go:353-369`) wires `signal.NotifyContext(os.Interrupt, syscall.SIGTERM)` only for **hook** commands. Every other command — including long-running ones like `store payload-backfill run`, `store payload-rehearsal`, `search projection rebuild` and `store compact` — gets a plain `context.WithCancel(context.Background())` that nothing ever cancels.

So `Ctrl-C` on those commands is Go's default SIGINT behaviour: the process dies immediately. The context is never cancelled, and no cleanup path runs.

## Why it matters

Long-running store commands were built to treat cancellation as a first-class stop:

- `payload-backfill` persists a `paused` checkpoint on cancellation so `resume` picks it up and `status` does not report the run active. On a hard kill it writes nothing and the run stays at `running` — recoverable via `resume`, but `run` refuses until then, and the operator has to know that.
- Other commands have their own cancellation handling with the same gap.

None of this is reachable from the CLI today, so the cleanest stop we implement is the one an operator cannot ask for.

## Scope

Extend the signal-derived context to non-hook commands, or at minimum to the long-running store commands. Two things need care:

1. The hook path deliberately layers a soft deadline (`TRACEARY_HOOK_SOFT_DEADLINE`) on top of the signal context; non-hook commands must not inherit a deadline.
2. The TUI installs its own SIGINT handler (`presentation/cli/tui/runner.go:51`), so it must keep winning for `tui` / `top`.
3. A second `Ctrl-C` should still hard-kill, so a wedged cleanup cannot trap the operator.

Once wired, drop the "does not currently wire signals" caveat from `docs/storage/payload-backfill.md` and its Japanese pair.

## Acceptance

- `Ctrl-C` on a long-running non-hook store command cancels its context and lets the command's own cleanup run.
- The hook soft deadline and the TUI's signal handling are unchanged.
- A second `Ctrl-C` terminates without waiting for cleanup.
- A test covers the signal-to-context wiring for at least one non-hook command.

duck8823:
## Summary
- Non-hook CLI commands now receive a signal-derived context (SIGINT/SIGTERM) with **no** hook soft deadline.
- After the first signal, the default handler is restored so a second Ctrl-C hard-kills a wedged cleanup.
- `memory inbox review` keeps `context.WithCancel` so Bubble Tea owns SIGINT and can restore the terminal.
- Docs drop the "does not currently wire signals" caveat on payload-backfill.

## Non-goals
- Does not change hook soft-deadline semantics or detached worker behaviour.
- Leaves parent #1904 open (implementation PRs must not complete the parent).

## Test plan
- [x] `go test .` (package main, includes SIGTERM-to-context and TUI no-NotifyContext)
- [x] `go tool golangci-lint run .`

Closes #1747

### DIFF
--- a/.ai/spec/1747.md
+++ b/.ai/spec/1747.md
@@ -1,0 +1,18 @@
+# #1747: wire Ctrl-C into non-hook command contexts
+
+## Policy
+
+- Hook commands keep `signal.NotifyContext` plus the existing soft deadline (`TRACEARY_HOOK_SOFT_DEADLINE`). Detached hook workers stay signal-only.
+- Non-hook commands get `signal.NotifyContext` **without** a deadline. After the first signal, `stopSignals()` restores the default handler so a second Ctrl-C hard-kills.
+- The only shipped TUI is `memory inbox review`. That path keeps `context.WithCancel` so Bubble Tea owns SIGINT/SIGTERM and can restore the terminal. `top` is already unknown.
+
+## I/O
+
+- `commandContext` in `main.go` is the single installer. Tests call it from its real start state (hook / detached worker / non-hook / TUI).
+- Non-hook signal coverage uses SIGTERM (same NotifyContext set as SIGINT) so the test process is not killed by the default SIGINT handler.
+- Docs: drop the "does not currently wire signals" caveat from `docs/storage/payload-backfill.md` and the Japanese pair.
+
+## Non-goals
+
+- Changing hook soft-deadline semantics.
+- Adding a new command surface.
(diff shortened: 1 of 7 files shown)


duck8823:
gpt-5.6-terra medium: ✅ (rematch after `--db-path` TUI detection)

```json
```

First pass HIGH: `isTUICommandArgs` missed `traceary --db-path /path memory inbox review`. Fixed in 432e30e1. Rematch clean. CI Test+Lint green on 432e30e1.

Leaves parent #1904 open.

PR #1910, opened 2026-08-14
fix: diagnose Antigravity hook permission denials on exit 0

### LINKED ISSUE #1864: v0.36-16: Antigravity readiness is not diagnosable: the probe reports a wrong reason and doctor has no check
## Summary

The Antigravity host's readiness cannot be diagnosed. The release-verification probe reports a wrong reason for the failure, and `doctor` — which the documentation says reports this exact condition — has no check for it at all. Found while dogfooding the v0.34.0 candidate (`3073a6de`).

## What happened

`scripts/verify-antigravity-headless-markers.sh` fails, twice in a row, with:

```
antigravity headless marker probe: expected public response marker was not returned
```

That is not the reason. Running the probe's own `agy` invocation directly:

```
$ agy --print --mode plan --sandbox --print-timeout 120s \
    "Reply with exactly TRACEARY_ANTIGRAVITY_HEADLESS_OK. Do not call tools."
exit 0
stdout: (empty)
stderr: jetski: no output produced — a tool required the "command" permission that headless
        mode cannot prompt for, so it was auto-denied.
```

The real cause is that this environment has no `permissions.allow` entries for the four packaged hook entrypoints — `~/.gemini/antigravity-cli/settings.json` has no `permissions` key at all — so the hooks are auto-denied in headless mode. This is exactly the condition `docs/integrations/antigravity.md:105-125` documents.

## Defect 1 — the probe reports the wrong reason

`scripts/verify-antigravity-headless-markers.sh:38-49` only consults stderr for the permission wording **inside the non-zero-exit branch**:

```sh
if ! agy --print --mode plan --sandbox --print-timeout 120s "$prompt" >"$stdout_path" 2>"$stderr_path"; then
  if grep -qi "permission" "$stderr_path"; then
    echo "... scoped hook permission is absent or shadowed" >&2
  else
    echo "... agy exited before the marker response" >&2
  fi
  exit 1
fi
if ! grep -Fxq "$marker" "$stdout_path"; then
  echo "... expected public response marker was not returned" >&2
  exit 1
fi
```

`agy` exits **0** here, with empty stdout and the permission message on stderr. The permission branch is therefore never reached, and the operator is told the model failed to answer when in fact the hooks were denied. The diagnosis the script already knows how to produce is unreachable in the case it was written for.

The fix is to check stderr for the auto-denied wording regardless of exit status, or to treat "exit 0 with empty stdout" as a distinct outcome rather than folding it into the marker check.

## Defect 2 — `doctor` cannot tell the operator either

`docs/integrations/antigravity.md:119-125` states that placing the entries in the wrong settings file means "`doctor` therefore reports that configuration as not ready". It does not. On this environment `doctor` emits 45 checks, and the only Antigravity one is `antigravity-plugin-version`:

```
claude-config, claude-event-coverage, claude-mcp, claude-global-config, claude-host-capabilities, ...
codex-config, codex-mcp, codex-host-capabilities, ...
gemini-config, gemini-event-coverage, gemini-mcp, gemini-global-config, gemini-host-capabilities, ...
antigravity-plugin-version          <- the only antigravity check
```

Claude, Codex and Gemini each have a config-readiness check and a host-capabilities check. Antigravity has neither. Nothing in the non-test, non-docs source reads `~/.gemini/antigravity-cli/settings.json`, so no check can be reporting on its `permissions.allow` content.

Antigravity is a declared supported host — `application/hostcoverage/matrix.json:216-222` lists it with `"doctor_client": "antigravity"` — so a supported host currently has version-only diagnostic coverage.

Either `doctor` gains the check the documentation promises, or the documentation stops promising it. The first is the useful direction: without it, an operator whose hooks are silently auto-denied has no surface that says so.

## Impact on the v0.34.0 release

#1621's host-integration criteria require that wake injection is observed on every host the RC declares as supported, that an injection failure is diagnosable by host and reason and never silent, and that no host is listed as supported without that observation.

As of this candidate the Antigravity observation cannot be completed in this environment, and when it fails it fails with a wrong reason. The narrow release action is to record Antigravity as **not observed** for v0.34.0 rather than to claim it. Fixing the two defects above is what makes the observation possible next time.

## Not part of this issue

Adding the allow-rules to the operator's own `~/.gemini/antigravity-cli/settings.json` is an environment change, not a code change, and `docs/integrations/antigravity.md:124-125` is explicit that the plugin does not overwrite operator-owned settings. Weakening the probe with `--dangerously-skip-permissions` or by dropping `--sandbox` is ruled out by the same document.

duck8823:
## Summary
- `scripts/verify-antigravity-headless-markers.sh` now diagnoses scoped hook permission denials even when `agy` exits 0 with empty stdout.
- `traceary doctor --client antigravity` already reports `antigravity-headless-hooks`; default doctor clients stay `claude,codex,gemini`.
- Tests drive the real script with a fake `agy` (no live store).

## Non-goals
- Does not add Antigravity to default doctor clients.
- Leaves parent #1904 open.

## Test plan
- [x] `go test ./cmd/repo-tooling -run TestVerifyAntigravityHeadlessMarkers`
- [x] `go tool golangci-lint run ./cmd/repo-tooling/`

Closes #1864

### DIFF
--- a/.ai/spec/1864.md
+++ b/.ai/spec/1864.md
@@ -1,0 +1,13 @@
+# #1864: Antigravity readiness diagnosis
+
+## Policy
+
+- The probe reports a permission denial whenever `agy` stderr mentions permission, **including exit 0 with empty stdout**. That is the live-host failure mode (hooks auto-denied in headless mode).
+- A non-zero `agy` exit without permission wording is still "exited before the marker response".
+- A zero exit without the marker and without permission wording stays "marker was not returned".
+- `traceary doctor --client antigravity` already owns `antigravity-headless-hooks` (exact allow of the four packaged hook commands). Default doctor clients stay `claude,codex,gemini` (existing pin). This ticket does not add a command.
+
+## I/O
+
+- `scripts/verify-antigravity-headless-markers.sh` classifies after capturing stdout/stderr.
+- Tests drive the real script with a fake `agy` on PATH and `TRACEARY_DOGFOOD_BINARY` so they do not build or talk to a live store.
--- a/CHANGELOG.ja.md
+++ b/CHANGELOG.ja.md
@@ -10,6 +10,9 @@ release note と同じ粒度で、版ごとの要点だけをまとめていま�
 ### Changed
 - **hook 以外の Ctrl-C が command context を cancel する (#1747)** — `store compact` など長い非 hook コマンドは signal 由来の context を使う（hook の soft deadline は付けない）。2 回目の Ctrl-C は即終了。`memory inbox review` の SIGINT はこれまでどおり Bubble Tea が持つ。
 
+### Fixed
+- **Antigravity headless probe が exit 0 の permission 拒否を報告する (#1864)** — `scripts/verify-antigravity-headless-markers.sh` は `agy` が空 stdout で exit 0 しても stderr の permission 文言を見る。`traceary doctor --client antigravity` の `antigravity-headless-hooks` は既存。
+
 ### Removed
 - **`traceary session active` を `session latest --active` に畳んだ (#1704)** — `session active` は unknown subcommand（非ゼロ、`DEPRECATED` なし）。`session latest --active` が以前の stale 既定（24h、`--stale-after`、`--allow-stale`）を引き継ぐ。`--active` なしの 2 flag は拒否する。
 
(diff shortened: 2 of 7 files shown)


duck8823:
gpt-5.6-terra medium: ✅ (rematch after raw agy exit capture)

```json
```

First pass HIGH: `if ! agy; then agy_exit=$?` recorded 0. Fixed in b6eb863b (`agy ... || agy_exit=$?`) with a nonzero-exit classification test.

Leaves parent #1904 open.

PR #1911, opened 2026-08-14
docs: keep compact rollback copies until the operator deletes them

### LINKED ISSUE #1827: v0.36-12: Decide when the compaction rollback copy is released

duck8823:
## Summary
- Decision: the compact rollback copy is retained until the operator deletes it. Apply-time `VerifyPair` is not in-use proof.
- No new command. Success JSON adds `rollback_retained: true` and keeps `rollback_path`.
- `doctor` reports leftover `<db>.rollback-*` siblings as `compact-rollback-copy`.

## Non-goals
- Does not auto-delete on commit or the next successful open.
- Does not add `store compact release`.
- Leaves parent #1904 open.

## Test plan
- [x] `go test ./presentation/cli -run 'TestStoreCompact|TestInspectCompactRollback|TestDoctor_JSON_Golden'`
- [x] `go tool golangci-lint run ./presentation/cli/`

Closes #1827

### DIFF
--- a/.ai/spec/1827.md
+++ b/.ai/spec/1827.md
@@ -1,0 +1,23 @@
+# #1827: when the compaction rollback copy is released
+
+## Policy
+
+Apply-time `VerifyPair` is real (compatibility, filtered/logical digest, attestation) and is **not** proof the compacted store is correct in use. Auto-deleting the rollback inode on commit, or on the next successful open, trades a disk problem for an unrecoverable data-loss risk.
+
+Therefore:
+
+- The rollback copy is **retained until the operator deletes it**.
+- No new command. `store compact rollback RUN_ID` restores; `rm` on the named `rollback_path` is the release.
+- Success JSON always names `rollback_path` and sets `rollback_retained: true`.
+- `doctor` reports a sibling `<db>.rollback-*` so the operator does not have to remember the filename.
+
+## I/O
+
+- CLI encode in `store compact` (existing JSON object).
+- Doctor walks only `filepath.Glob(dbPath + ".rollback-*")` — no extra store open, no live-store requirement.
+- Tests drive the real `store compact` command and the doctor check from a scratch path.
+
+## Non-goals
+
+- A `store compact release` / `forget` subcommand.
+- Deleting the copy from the apply or open path.
--- a/CHANGELOG.ja.md
+++ b/CHANGELOG.ja.md
@@ -8,6 +8,7 @@ release note と同じ粒度で、版ごとの要点だけをまとめていま�
 ## [Unreleased]
 
 ### Changed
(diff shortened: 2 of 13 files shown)


duck8823:
gpt-5.6-terra medium: ✅ (rematch after growth-budget test count)

```json
```

First pass HIGH: 2 GiB doctor skipped compact-rollback-copy. Second pass HIGH: inspectStoreGrowthBudget tests still expected 2 checks. Both rematched on acd6a121. CI Test+Lint green.

Leaves parent #1904 open.

PR #1912, opened 2026-08-14
feat: report this store's measured cost from doctor

### LINKED ISSUE #1809: v0.36-9: report the operator's own storage cost instead of a global monthly claim
Raised while restating the #1620 release gate in rate-independent units.

## Why

The v0.34 gate used to be published as `<= 4 GiB store` and `<= +0.5 GiB/month`. Both are `bytes-per-unit-of-work x work-per-month`, and only the first factor is a property of Traceary. A heavy user reads the monthly figure as a promise that is broken on day one; a light user reads it as a bound that a 10x per-record regression would still satisfy. #1620 now gates on per-record units instead.

That fixes the release note but leaves the operator's actual question unanswered: **"how much is this going to cost *me*?"** That question is legitimate and it has a correct answer — it just cannot be answered by the project, only by the store in front of the operator.

## What to add

`traceary doctor` (and the `--json` form) should report the store's own measured figures, which are exactly the inputs the gate is stated in:

- resident bytes, and resident bytes per recorded event and per session
- recorded source-text bytes over a trailing window, per event and per session, split undiscardable vs foldable
- storage amplification: resident bytes / retained source-text bytes
- the observed event and session rate over that window
- the projection derived from the operator's own rate — e.g. "at your rate of N events/day, the undiscardable share grows ~X MiB/month" — which is legitimate here precisely because the rate is measured rather than assumed

For reference, the figures on the maintainer's rehearsal copy on 2026-08-11 were 2.67 KiB/event undiscardable, 41.7 KiB/session, 7.36 KiB/event resident, 2.55x amplification, ~300k events/month.

## Constraints

- Must not be a new top-level command. The `/goal` standing instruction for this sprint is to reduce surface, not add to it — this belongs inside `doctor`, or as a section of `status`, whichever already owns store-health reporting. Pick one and say why in the PR.
- Must be cheap enough to run on a multi-GiB store. `dbstat` over a whole large store is not free; if the full measurement is too slow for the default path, gate the expensive part behind a flag and report the cheap subset by default.
- Must never run against the live store from a repo build during development (`ref_live_db_migration_gap`) — migrations apply unconditionally on open.
- JSON contract: this adds fields to a frozen contract surface, so it needs a snapshot test alongside.

Related: #1620 (the gate restatement this comes from), #1797.

duck8823:
## Summary
- `traceary doctor` (already aliased as `status`) reports this store's measured cost. `store capacity` stays dbstat-only.
- JSON adds `operator_cost` (`traceary.operator_cost/v1`) and check `store-operator-cost`: resident bytes per event/session, undiscardable vs foldable source-text (projection + refinements, no body reads), amplification, and a monthly projection from the last 30 days of *this* store's rate.
- Stores ≥ 2 GiB stay metadata-only (resident file size only). No new command.

## Non-goals
- Does not open a ≥ 2 GiB store.
- Does not publish a global monthly bound.
- Leaves parent #1904 open.

## Test plan
- [x] `go test ./infrastructure/sqlite -run TestOperatorCost`
- [x] `go test ./presentation/cli -run 'TestBuildOperatorCost|TestResidentOnly|TestRootCLI_DoctorJSONIncludesOperatorCost|TestRootCLI_DoctorLargeStoreReturnsBounded'`
- [x] `go tool golangci-lint run ./application/... ./infrastructure/sqlite/ ./presentation/cli/ .`

Closes #1809

### DIFF
--- a/.ai/spec/1809.md
+++ b/.ai/spec/1809.md
@@ -1,0 +1,22 @@
+# #1809: report the operator's own storage cost
+
+## Policy
+
+- Lives in `traceary doctor` (`status` is already an alias). `store capacity` stays dbstat-only; this answers rate × bytes-per-unit for **this** store.
+- No new command.
+- Default doctor on stores it already opens (< 2 GiB) runs indexed aggregates on `event_metadata_projection`. It does not scan `events.body`.
+- Stores ≥ 2 GiB stay metadata-only: resident file size only, no SQLite open.
+- Foldable = transcript rows older than the compact cutoff (90 days) whose session has a `session_refinements` row. Undiscardable = retained source bytes minus foldable.
+- Rate window is 30 days. Monthly undiscardable projection = events/day × 30 × undiscardable-bytes-per-event. Omitted when the window is empty.
+
+## I/O
+
+- `application.OperatorCostInspector` / `types.OperatorCostReport` (`traceary.operator_cost/v1`).
+- Doctor JSON adds optional `operator_cost` and check `store-operator-cost`.
+- Tests drive the real inspector against a scratch SQLite file and the real `doctor --json` command.
+
+## Non-goals
+
+- New top-level command.
+- Opening a ≥ 2 GiB store from doctor.
+- Reading event bodies.
--- a/CHANGELOG.ja.md
+++ b/CHANGELOG.ja.md
@@ -7,6 +7,9 @@ release note と同じ粒度で、版ごとの要点だけをまとめていま�
 
 ## [Unreleased]
 
+### Added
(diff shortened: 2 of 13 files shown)


duck8823:
gpt-5.6-terra medium: ✅ 問題なし on `61025d54` (operator-cost feature).

Follow-up `30447952` is test-only: `TestRootCLI_DoctorJSONIncludesOperatorCost` now uses `setTracearyPathToCurrentExecutable` + isolated HOME + `executeDoctorAllowWarnings` so unrelated doctor fail-checks in CI do not mask the `operator_cost` assertion. No production behavior change after Terra approval.

CI Test on `30447952` is green (first rerun flake was `TestVerifyAttestationChain_ConcurrentAppendDoesNotFalseFail` database is locked on main, not this PR).

PR #1913, opened 2026-08-14
docs: establish what failed=1 means and keep list --failures

### LINKED ISSUE #1767: v0.36-6: establish what failed=1 means before deciding the future of list --failures

duck8823:
## Summary
- Decision from source (no live store): `failed=1` is the compatibility flag derived from structured `failure_reason`. Pre-classifier `unknown`+`failed=1` is the `000025` DEFAULT leaking onto `hookPayloadFailed` rows before `11691479` (2026-07-22). Current hook writes persist `host_error`.
- Keep `list --failures`. It is a 記録 surface (Claude `PostToolUseFailure` / Gemini spawn error / `hook_denied` / `signal` / `timeout`). Do not fold it into `doctor`.
- No new CHECK forbidding `failed=1 AND failure_reason='unknown'` — restore must keep historical rows.

## Non-goals
- Does not remove or deprecate `list --failures`.
- Does not query or rewrite the live store.
- Does not backfill historical `unknown` rows.
- Leaves parent #1904 open.

## Test plan
- [x] `go test ./domain/model -run 'TestCommandAuditClassifyOutcome|TestCommandAuditFromSnapshotPreservesLegacy'`
- [x] `go test ./presentation/cli -run TestRootCLI_ListFailuresKeepsStructuredAndLegacyFailedRows`

Closes #1767

### DIFF
--- a/.ai/spec/1767.md
+++ b/.ai/spec/1767.md
@@ -1,0 +1,26 @@
+# #1767: what `failed=1` means
+
+## Policy
+
+- Keep `list --failures`. It is a 記録 surface, not a `doctor` diagnostic.
+- Do not add a CHECK that forbids `failed=1 AND failure_reason='unknown'`. Restore must keep pre-classifier rows.
+- New writes never persist that pair. `ClassifyOutcome` upgrades `unknown` + structural failure to `host_error`. Hook `Failed` is `failureReason.IsFailure()`.
+
+## Answers (source only; no live store)
+
+1. `unknown` + `failed=1` is the `000025` DEFAULT leaking onto rows flagged by `hookPayloadFailed` before `11691479` (2026-07-22).
+2. That commit + `b1daa0e7` (same day) are the 2026-07-21/24 cut. Current hook writes cannot produce the pair. Restore/snapshot still can.
+3. Current `host_error` is hook `PostToolUseFailure` / structured host error (Claude top-level `error`, Gemini `tool_response.error`). Codex has no structured failure field. Grok `PermissionDenied` is `hook_denied`.
+4. Those rows are host-reported tool failures the operator asked to record. Keep `--failures`.
+5. No new CHECK. Domain write path already forbids unclassified new failures.
+
+## I/O
+
+- Docs: `docs/research/failed-flag-meaning.md` (+ `.ja.md`), storage + CLI pairs.
+- Tests drive real `hook audit` + `list --failures` on a scratch SQLite file.
+
+## Non-goals
+
+- Removing or deprecating `list --failures`.
+- Querying the live store.
+- Rewriting historical `unknown` rows.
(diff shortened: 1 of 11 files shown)


duck8823:
gpt-5.6-terra medium: ✅

```json
```

Read-only review of `7be82eaa` vs `origin/main` (`codex exec -m gpt-5.6-terra`, reasoning medium).

PR #1914, opened 2026-08-14
feat: count workspace conflict pairs and keep store workspace-alias

### LINKED ISSUE #1768: v0.36-7: decide what a workspace conflict is for before touching store workspace-alias

duck8823:
## Summary
- Decision from source (no live store): a workspace `conflict` is an operator-reviewable `(session_id, workspace)` pair. The classifier is doing what the contract says (no auto-alias of remote vs path).
- `report workspace-identity` keeps observation-row counts as volume and adds `conflict_pair_count`. Samples are one latest row per pair and include `workspace`.
- Keep `store workspace-alias`. It is the only public add/remove/list path. Auto-normalisation would violate the contract. Deprecation stays blocked.

## Non-goals
- Does not remove or deprecate `store workspace-alias`.
- Does not query or rewrite the live store.
- Does not change row-based `relationships.conflict` or `conflict_rate`.
- Leaves parent #1904 open.

## Test plan
- [x] `go test ./infrastructure/sqlite -run TestWorkspaceIdentityDatasource_`
- [x] `go test ./presentation/cli -run TestRootCLI_WorkspaceIdentityReportCountsConflictPairs`
- [x] `go tool golangci-lint run ./application/types/ ./infrastructure/sqlite/ ./presentation/cli/`

Closes #1768


duck8823:
gpt-5.6-terra medium: first pass HIGH on `session_id || char(31) || workspace` colliding distinct pairs. Fixed in `4d3904e1` with `SELECT DISTINCT session_id, workspace`.

Rematch on `4d3904e1af656939939c5cd359661a26c2d7b27e`: ✅

```json
```

PR #1915, opened 2026-08-14
feat: apply canonical payload codec on import and restore

### LINKED ISSUE #1779: v0.36-8: decide the remaining payloadCodecIdentity call sites

duck8823:
Closes #1779

Leaves parent #1904 open.

## Summary

Decide the remaining `payloadCodecIdentity` call sites individually (not a sweep).

| Site | Decision |
|---|---|
| Bundle import | Adopt `encodeCanonicalPayload` |
| Archive restore | Adopt `encodeCanonicalPayload` |
| Dedupe restore | Adopt `encodeCanonicalPayload` |
| Raw-body recovery restore | Adopt `encodeCanonicalPayload` |
| Retention markers | Stay identity |

Interchange formats (bundle / archive) stay plaintext. The writer on the way into the live store matches native `insertEventAndAudit`. Markers stay identity because apply/verify compare stored TEXT to `EventBodyUnavailableRetentionMarker`.

## Tests

- `go test ./infrastructure/sqlite -count=1 -run TestBundleImport_UsesCanonicalPayloadCodec\|TestArchiveRestore_UsesCanonicalPayloadCodec\|TestDedupeRestore_UsesCanonicalPayloadCodec`
- `go test ./...`
- `go tool golangci-lint run`

See `docs/research/payload-codec-call-sites.md`.


duck8823:
## gpt-5.6-terra r1

**CHANGES_REQUESTED** — HIGH: adopt sites lacked codec + byte-identical tests for `ImportCommandAudit`, archive `command_audits`, and raw-body restore. Tests only asserted `events.body_codec`.

Addressed on `a0b1dc3e` (`test: cover audit and raw-body restore codec adoption`). Rematching.

duck8823:
## gpt-5.6-terra r2

**APPROVE** — 問題なし. r1 HIGH (audit / archive / raw-body restore codec tests) is addressed on `a0b1dc3e`.

PR #1916, opened 2026-08-14
feat: calibrate storage gates against synthetic corpora

### LINKED ISSUE #1811: v0.36-10: calibrate the storage gates against synthetic worst-case corpora, not one maintainer's store
Raised by gpt-5.6-sol's adversarial review of the #1620 gate restatement, and deliberately **not** added as a gate row before the v0.34 RC because it cannot be measured honestly in the time available. Filing it rather than pretending the gap is not there.

## The gap

Every figure in the v0.34 release gate comes from one store: the maintainer's, on one machine, with one usage pattern. That includes both amplification rows, which are the rows most sensitive to corpus shape:

- **whole-store amplification** (2.55x measured) depends on average record size versus SQLite page size — a corpus of many tiny records pays far more page slack.
- **recent search index amplification** (3.37x measured, against a 2.16x constant the code still uses as a fallback — #1751) depends on trigram uniqueness. CJK text, base64 blobs, minified JSON and long unique identifiers each move it in different directions, and the maintainer's corpus is heavily one language.

A gate calibrated on one corpus does not tell an operator with a different corpus anything, and it does not tell the project whether a change is a regression or a corpus difference.

## What would close it

A small set of **synthetic corpora**, generated deterministically and committed as a generator (not as data), each isolating one dimension:

- many tiny records (page-slack worst case)
- few enormous records (the per-row byte-budget and exclusion path, #1794)
- CJK-heavy text (multi-byte, trigram-dense)
- high-entropy text (base64, hashes — trigram uniqueness worst case)
- highly repetitive text (dedupe and compression best case, to bound the other end)

Then run the same measurement script that produced the #1620 figures against each, and publish the **range** rather than a point. The gate rows become "within [min, max] across the corpus set" or gain a margin justified by the observed spread instead of a guess.

## Constraints

- Deterministic generation, committed as code with a seed. No sample data in the repository.
- The measurement has to be the same code path that measures a real store, or the numbers are not comparable.
- This is a benchmark, not a unit test — it must not run in the default `go test ./...`.
- Related existing surface: `cmd/store-benchmark` and `docs/operations/capacity-benchmark.md` already define a sanitized capacity-baseline format; extend that rather than inventing a second one.

Related: #1620 (the gate this calibrates), #1751 (the amplification constant), #1796.

duck8823:
Closes #1811

Leaves parent #1904 open.

## Summary

The #1620 whole-store amplification figure was one maintainer store. This PR adds five seeded corpora and a benchmark harness that reports a **range**.

| Corpus | Dimension |
|---|---|
| tiny | short unique records (page slack) |
| enormous | large records |
| cjk | multi-byte text |
| entropy | hashes |
| repetitive | compression best case |

```sh
go run ./cmd/store-benchmark --calibrate-gates DIR
```

Writers are production `EventDatasource.Save`. Measurement uses `CapacityInspector` and `OperatorCostInspector` (same as `store capacity` / `doctor`). Search-index amplification stays `unmeasured` without a completed search-projection generation (rebuild needs ≥ 8 MiB recent-tier sample). Not part of `go test ./...`.

See `docs/research/storage-gate-calibration.md`.


duck8823:
## gpt-5.6-terra r1

**APPROVE** — 問題なし.

PR #1917, opened 2026-08-14
feat: measure fold and wake release-gate rows on an explicit store copy

### LINKED ISSUE #1879: v0.36-18: measure the two release-gate rows that v0.34 could not evaluate
Carried over from #1620. Two rows of the v0.34 release gate could not be measured before the v0.34.0 tag, for a structural reason rather than a defect.

## The two rows

| row | gate | v0.34.0 baseline |
|---|---|---:|
| sessions with a refinement | >= 95% of sessions worth folding | **0%** |
| wake injection | works on every host, within budget | **0 eligible rows** |

Measured on the reference store at the tag: **27,552 sessions, 0 `session_refinements` rows, 0 wake-eligible rows.**

## Why they were unmeasurable at v0.34.0

Both numbers have one cause. The consolidation ask — the mechanism that causes an agent to fold a session — **ships for the first time in v0.34.0** (`aff94d67`, 2026-08-08; absent from `v0.33.0` and every earlier tag). Nothing had ever asked an agent to fold anything, so no refinement existed, so wake injection had nothing to inject.

Holding the tag for these rows would have meant holding it for the absence of the very mechanism the tag ships. They are post-release measurements by construction.

## What to measure in v0.35

With v0.34.0 in real use, both become real numbers with a known starting point (0 / 27,552).

1. **Refinement ratio.** Of sessions large enough to cross the consolidation threshold, what fraction acquired a refinement? "Sessions worth folding" needs a definition that is computable rather than judged — the threshold crossing is the obvious candidate, since that is exactly when the ask fires.
2. **Wake injection.** Does it inject on every host, within the byte budget? Measure per host, not in aggregate; capture coverage is known to differ (gemini and antigravity wire fewer events).
3. **Summary content.** #1874 landed the content specification in the ask (v0.34.0). Sample the summaries that actually arrived and check whether they carry motivation and the change, rather than only counting rows. A 95% ratio of contentless summaries would satisfy the gate and mean nothing.

## Two known obstacles, both filed

- **#1877** — `store gc` supersedes a correctly-folded refinement with `degraded=1`, and `list_session_wake_summaries.sql:16` filters `degraded = 0`. Measured: a session with an intact agent-authored summary yields **0** eligible rows. Under #1872 the reduction runs routinely, so the wake-injection row would fail even once refinements exist. **This row cannot be measured honestly until #1877 is fixed.**
- **#1873** — nothing in CI, the Makefile, or any test evaluates any gate threshold. Measuring these two by hand again in v0.35 repeats the v0.34 situation rather than fixing it.

## Done when

- Both rows have a measured value on a real post-v0.34 store, with the corpus described
- "Sessions worth folding" has a computable definition, stated here
- Wake injection is measured per host, not in aggregate
- A sample of arriving summaries is checked against #1874's content specification, not just counted
- Either the rows pass, or the miss is explained with the same rigour #1620 applied to the index-family rows

duck8823:
Closes #1879

Leaves parent #1904 open.

## Summary

The two v0.34 rows that could not be measured at tag time now have a computable definition and an explicit-copy harness.

**Sessions worth folding:** `session_refinements` row exists, or `SUM(event_metadata_projection.body_stored_bytes) >= consolidation.threshold_bytes` (default 64 KiB).

```sh
go run ./cmd/store-benchmark --fold-gates --db /path/to/copy.db
```

`--db` is required. The default live store (`~/.config/traceary/traceary.db`) is refused. Wake is counted per `sessions.client` with the #1877 `has_agent_reasoning` rule. Content is a structural sample, not a semantic parse of motivation/change.

The last honest live-store figure remains the v0.34.0 tag baseline (0 / 27,552 / 0). This PR does not query the live store.

See `docs/research/fold-gate-measurement.md`.


duck8823:
## gpt-5.6-terra r1

**CHANGES_REQUESTED**

- HIGH: live store could be reached via symlink (`filepath.Abs` only)
- HIGH: wake budget fit ignored the injection header/newlines

Addressed on `aeddce1a`. Rematching.

duck8823:
## gpt-5.6-terra r2

**APPROVE** — 問題なし. r1 HIGHs (symlink live path, wake envelope budget) are addressed on `aeddce1a`.

Issue #1686, opened 2026-08-07
v0.36-1: retire the event metadata projection, once events is actually narrow


duck8823:
> **Deferred out of v0.34 on 2026-08-10, after measurement.** The premise below — "once #1685 narrows
> `events`, the projection has no reason to exist" — is false. #1685 was rescoped before it shipped and
> delivered body **compression**, not the side-table move; `events` is still wide. Measured, retiring the
> projection today makes metadata reads 2.7x slower warm and up to 15x slower cold, because the cost is
> fetching a row from a 6.3 GiB body-bearing table, not scanning the index. Covering indexes cannot close
> the gap within the space this issue would reclaim. Full measurement, reproduction SQL and independent
> review in the [2026-08-10 comment](#issuecomment-5234712983).
>
> **Revised dependencies:** #1743's measurement *and* the body extraction actually landing, plus a decision
> on where `legacy_source_hook` lives and on `command_audits` payload locality (the projection denormalises
> audit facts too, so narrowing `events` alone is not sufficient). Re-run the 21 query shapes after that.

## Problem

`event_metadata_projection` exists so scans avoid the wide `events` table. Once #1685 narrows `events`, it has no reason to exist.

Measured on the live store 2026-08-08 (`mode=ro`):

```
event_metadata_projection                                     0.13 GiB
idx_event_metadata_workspace_session_created_at_norm_id_desc  0.10 GiB
idx_event_metadata_session_created_at_norm_id_desc            0.08 GiB
idx_event_metadata_workspace_created_at_norm_id_desc          0.06 GiB
+ remaining projection indexes                               ~0.10 GiB
                                                             ------
                                                             ~0.47 GiB
```

Each of those duplicates an `events` index one-for-one (`idx_events_workspace_session_created_at_norm_id_desc` is the same 0.10 GiB).

## Read surface

**20 SQL files reference the projection**, several more than once — `find_latest_session.sql` alone has 7 references, `find_active_session.sql` 5, `find_latest_session_boundary.sql` 3, `list_sessions.sql` 4. This is the largest single-table read surface in the store, so stage 1 is not a small edit.

## Scope

Two stages inside this minor:

1. Switch read paths back to `events`; keep the projection written but unread
2. Drop the table and its indexes

Staging matters: a single-step drop leaves no way back if one of those 20 files was missed.

## The comparison this needs (previously undefined)

"Query plans are not worse" has to be a measurement, not a judgement. Fix the method before stage 1:

[...]

PR #1918, opened 2026-08-14
feat: measure body side-table locality and keep events.body inline

### LINKED ISSUE #1743: v0.36-4: measure whether a body side table is worth it, then decide
## Why this is a separate issue

#1685 was originally scoped as "move event bodies out of the events table **and** compress them". The
move half was dropped from #1685 because its stated justification measured false. This issue keeps
the idea alive on the one argument that survives, and requires evidence before it is implemented.

## The justification that failed

#1685 argued:

> Discarding a body becomes `UPDATE body=''`, which does not release pages [...] A side table makes
> discard a `DELETE`, returning pages to the free list for reuse. **No VACUUM.**

Measured (50,000 rows × 4 KiB body, page_size 4096, `modernc.org/sqlite`, WAL checkpointed between
readings):

| | page_count | freelist | reusable |
|---|---|---|---|
| before | 56,525 | 0 | 0.0 MiB |
| after `UPDATE events SET body=''` | 56,525 | **50,000** | **195.3 MiB** of 220.8 |
| after `DELETE FROM events` | 56,525 | 56,522 | 220.8 MiB of 220.8 |

Blanking a large TEXT already returns its overflow pages to the free list. The gap is **88% vs 100%**,
not 0 vs 100 — the remainder is the leaf space holding the row's other columns. A body small enough
to fit inside its leaf page frees nothing either way (200 B bodies × 200,000 rows: freelist stayed 0).

And neither shrinks the file: `page_count` is identical in every row above. Only `store compact`
returns physical bytes. "No VACUUM" conflated free-list reuse with file size.

## The justification that survives, unmeasured

**Locality.** `events` is scanned for metadata far more often than for bodies. Inline multi-KiB bodies
push those scans through pages they do not need, so a metadata-only query pays for body pages it never
reads. `event_metadata_projection` exists partly to work around exactly this — and #1686 retires it.

That is a plausible claim. It is not measured, and it is on neither #1620 gate.

## What this issue must do before writing any schema

Measure it. On a corpus resembling the live store (~550k events, bodies compressed by #1685):

1. `EXPLAIN QUERY PLAN` and page-read counts for the metadata-only queries that matter — `traceary
   list`, `top`, session lookup, the search projection rebuild walk
2. the same with bodies in a side table
3. whether the win survives once #1685's compression has already shrunk the inline bodies by ~3x
4. whether it survives #1686 retiring `event_metadata_projection`

If the difference is not material, close this issue with the measurement recorded. That is a success,
not a failure.

## What it would cost, if the measurement justifies it

Recorded here so the trade is visible up front (all verified against the schema-52 tree):

- `ON DELETE CASCADE` plus the `ON UPDATE` question — the current schema handles event-ID changes
  explicitly (`schema/sqlite/migrations/000034_create_event_metadata_projection.sql:184-206`), and the
  FK default is `NO ACTION`
- a cross-table invariant SQLite's `CHECK` cannot express (no subqueries), so "does this event have a
  body" becomes application-enforced in every mutation path, fail-closed
- relocating three trigger families that key on `UPDATE OF body`:
  `search_projection_events_update` (038), `literal_search_event_update` (039),
  `search_projection_complete_event_update` (041)
- re-wiring four consumers, each a **silent content-loss** path if missed:
  `infrastructure/sqlite/canonical_event_audit.go:22-26,140-161`,
  `domain/archive_segment.go:59-66`, `infrastructure/sqlite/store_archive.go:492-517`,
  `infrastructure/sqlite/bundle_datasource.go:532-570`
- `infrastructure/sqlite/payload_codec_compatibility.go:31-40` `legacy_index` mode reads a partial
  index on `events.body_codec`; a side table drops out of the compatibility answer entirely
- the `events.body` column can then be dropped, but note the measured `ALTER TABLE` behaviour under
  `legacy_alter_table=0`: SQLite refuses the drop when a trigger *body* or a view names the column
  (safe), but **succeeds and leaves a dangling reference** when the only mention is an
  `AFTER UPDATE OF body` column list — which is the shape of all three families above

## Done when

Either a measurement showing the locality win is material and a design note that accounts for the
costs above, or a measurement showing it is not and this issue closed with the numbers recorded.

Design note and all measurements: [`.ai/spec/1685.md`](../blob/main/.ai/spec/1685.md), decision D1.

Parent: #1606

duck8823:
Closes #1743

Leaves parent #1904 open

## Decision

Do not move `events.body` to a side table.

`go run ./cmd/store-benchmark --measure-body-locality DIR` compares projection, inline `events`, and a scratch `event_bodies` layout. Live store path is refused.

Scratch 512×8 KiB (2026-08-15):

- entropy: inline 1.87× projection; side 1.01× projection; DB bytes grew 3.70 → 3.83 MiB
- repetitive (codec shrinks): inline 1.07×; side does not help; DB bytes also grew

The 2026-08-10 live 2.7× / 15× gap (cited from #1686, not re-queried here) is why the projection stays, not why we extract bodies. Current metadata reads already avoid `events`.

See `docs/research/body-side-table-locality.md`.


duck8823:

確認済み。live store の直接・親ディレクトリ拒否、scratch-only の実測、timing からの gate 算出、実ハーネスを通すテスト、UPDATE body='' + VACUUM による側表シミュレーション、記録値に整合する not_material 判断に CRITICAL/HIGH の問題はありません。

PR #1919, opened 2026-08-14
feat: keep event_metadata_projection while events stays wide

### LINKED ISSUE #1686: v0.36-1: retire the event metadata projection, once events is actually narrow

duck8823:
Closes #1686

Leaves parent #1904 open

## Decision

Keep `event_metadata_projection`. Do not switch metadata reads back to `events`.

#1743 decided not to extract `events.body`, so `events` is still wide. The 2026-08-10 live comparison (2.7× warm / 15× cold for `LIMIT 5000`) still applies. Reclaiming ~0.47 GiB is not worth that on a sheet that already closed at ~2.57 GiB without this lever.

See `docs/research/event-metadata-projection-retention.md`.


duck8823:

#1743 の決定と 2026-08-10 の 2.7× warm / 15× cold は projection を残す判断を支える。schema / shipped SQL の drop や書き換えはなく、retention test は shipped SQL を読み、live store は開いていない。

PR #1920, opened 2026-08-14
feat: report recent_source_bytes cache against a generation SUM

### LINKED ISSUE #1819: v0.36-11: recent_source_bytes is a cache with no verifier
`recent_source_bytes` (migration `000058`, added by #1796 / PR #1817) is a **cache** of a quantity the store can compute exactly: the source bytes of `search_projection_recent_documents` rows belonging to the rebuilding generation. Interleaved eviction reads the cache, never the ground truth, because the ground truth is a scan of the table the counter exists to keep small.

The invariant is maintained in exactly one place — `ApplyBatch`'s single `UPDATE … recent_source_bytes=recent_source_bytes+?` — and nothing else in the store enforces it. Today that is sound, and the review of #1817 established why:

- Only `Class == "eviction"` decrements, so the terminal `cleanup` phase deleting *other* generations' rows correctly leaves the counter alone (`search_projection_rebuild.go:923`).
- `cleanup_scope='all'` does delete the current generation's rows without decrementing, but that generation ends `drifted` and the next `Start` resets the counter to 0 (`:227`).
- Replay is fenced in the source phase by `(?=0 OR recent_source_bytes=?)` (`:977`), and in the eviction phase by the delete-must-affect-exactly-one-row rule (`:920`).
- A writer that does not know the column cannot exist in the wild: the entire projection subsystem is new in v0.34 (`git ls-tree -r v0.33.1 | grep search_projection` returns nothing, and v0.33.1's catalog stops at migration 034), so no released binary both rebuilds the projection and lacks 058.

What is missing is a **verifier**. Every one of those arguments is a proof about today's call sites. Add a call site that deletes a recent document, or a future writer that admits one, and the counter silently drifts:

- drift **low** → eviction under-fires; the peak this fix bounds becomes unbounded again, and the completed recent tier can exceed its ceiling.
- drift **high** → eviction over-fires; recent documents are deleted that should have been retained.

Neither is detectable from the outside. Both look exactly like correct operation.

## What to consider

1. **Trigger-maintained counter.** `AFTER INSERT` / `AFTER DELETE` on `search_projection_recent_documents` doing `UPDATE search_projection_state SET recent_source_bytes = recent_source_bytes ± length(…) WHERE singleton=1 AND generation_id = NEW.generation_id`. The generation predicate makes old-generation cleanup a no-op for free. This moves the invariant into the schema, so *any* writer maintains it, and `ApplyBatch`'s arithmetic goes away. Cost: a singleton-row update per projected document on the hottest write path in the rebuild — measure before adopting.
2. **Reconcile at a phase boundary.** The source→eviction transition already re-derives capacity and walks `dbstat`; one `SUM` over the generation's rows there would catch drift once per generation at a point that is already expensive. It does not protect the source phase itself, which is where the peak is.
3. **Report it rather than enforce it.** `traceary doctor` (or `store search-projection status`) computes the true sum and reports the delta. Cheapest, catches drift after the fact, does not prevent it.

Option 1 is the only one that makes the invariant unbreakable; option 3 is the only one with no hot-path cost. They compose.

## Not urgent because

No released binary can produce the drift, and #1817's own migration `000058` already resets any generation left mid-rebuild by an earlier development build — the one real instance of this class.

## Origin

Raised by gpt-5.6-sol reviewing PR #1817, as a CRITICAL claiming a v0.33.1 rollback could desynchronise the counter. That specific mechanism was checked and rejected — v0.33.1 has no projection code — but the underlying observation, that a cache with no verifier is one call site away from silent drift, is correct and worth carrying into v0.35.

Related: #1796, #1620.

duck8823:
Closes #1819

Leaves parent #1904 open

## Decision

Option 3: `traceary store search-projection status` reports the `recent_source_bytes` cache against `SUM(decoded_bytes)` for `search_projection_state.generation_id`.

No triggers. Status does not rewrite the cache. The SUM is the rebuilding generation, not `recent_bytes` (active generation).

See `docs/research/recent-source-bytes-verifier.md`.


duck8823:

generation_id を使う読み取り専用の status 経路で cache と SUM(decoded_bytes) を比較している。実スクラッチDBのインフラ・CLIテストがあり、start/resume 制御経路への新規SUMやライブDB参照はない。

Issue #1852, opened 2026-08-12
v0.36-14: nothing in the runtime path consults MigrationExecutionClass, so offline migrations apply implicitly


duck8823:
v0.35. Follow-up to #1851, which removed the one migration that bites today without addressing the mechanism that let it through.

`MigrationExecutionClass` is a reviewed, source-owned property. `MigrationDataDependentOffline` is documented as "requires preparation outside publication". **Nothing in the runtime path reads it.** It appears only in `infrastructure/sqlite/prepared_migration_catalog.go` and its test; `migrate()` (`infrastructure/sqlite/migrate.go:34-41`) loops every unapplied migration and calls `executeExactMigration` regardless of class.

So the classification is a comment with a digest attached. Five migrations carry it — 35, 41, 42, 45, 48 — and all five were applied unconditionally on first open. #1851 established that this produces a permanent stall when the work exceeds the host's hook timeout: `applyMigration` is one transaction per migration, so a killed migration rolls back completely and the next open restarts it from zero, identically, forever.

048 was the instance large enough to cross 60 s on a real store. The next one will be found the same way — in production, by an operator whose store stopped advancing and who was told nothing.

## Design

gpt-5.6-sol's recommendation, recorded here so the next cycle does not re-derive it:

`migrate()` consults the class and refuses to apply `data_dependent_offline` migrations implicitly. A hook or ordinary CLI invocation fails immediately — milliseconds, not a 60 s block — with an error naming the pending versions and the command that applies them. The store is not modified.

Route the explicit application through the **existing** `traceary store init`, which is already documented as the explicit initialization path. Do not add a command; the standing direction for this surface is to fold duplicates away, not to grow it. Split `Initialize(ctx)` (implicit, every current caller) from an operator-authorized variant used only by `runInit` (`presentation/cli/init.go:55-67`).

`store init` should say what it is about to do before it starts, and not go silent for minutes.

A newly created empty store must still reach the current schema on implicit open alone — a fresh install must not require a maintenance step. Migration 042 already uses the shape this needs: `NOT EXISTS(SELECT 1 FROM events LIMIT 1)`.

Reclassify 041 and 042 while here. They create tables and run a bounded `INSERT ... SELECT` over projection state, not over `events`; on the maintainer's 32 GB store their cost is negligible. They are `constant_in_place` in substance. That leaves 035 and 045 as the genuinely data-dependent ones — both `CREATE INDEX`, expensive but bounded and restartable at the same cost.

[...]

Issue #1672, opened 2026-08-07
v0.34.0-32: take the compact digest the host actually delivers


duck8823:
## Problem

`presentation/cli/hook_runtime.go:674` calls `SetSummaryIfEmpty` on the **pre-compact** branch. The real digest arrives at **post-compact**. Result: 4 digests across 26,728 sessions.

## Host reality (measured)

`compact_summary` events total 2,035. Bodies over 200 B: **186, all Claude**. Codex 964, Kimi 309, Grok 38 are empty markers — those hosts expose no compacted summary body.

## Scope

- Read the digest from the `post-compact` payload, not `pre-compact`
- Route it through the #1673 write port so it merges instead of first-write-wins
- Treat the Claude digest as free material, not as the mechanism (it covers 6.8% of events)
- Record the pre-compact marker as an orphan-range signal only (#1683)

## Done when

- A Claude post-compact fixture with a digest lands as a `session_refinements` row
- A marker-only fixture (Codex/Kimi/Grok) creates no refinement and does not error
Spec: [`.ai/spec/1606.md`](../blob/main/.ai/spec/1606.md) (landed by #1671)

Parent: #1606

PR #1926, opened 2026-08-15
feat: read session summaries from refinements

### LINKED ISSUE #1706: v0.36-3: retire sessions.summary in favour of session refinements
## Problem

After #1673 the store has two places that hold "the summary of this session":

| carrier | written by | read by |
|---|---|---|
| `sessions.summary` | `traceary session end --summary`, and `SetSummaryIfEmpty` from the compact hook (`presentation/cli/hook_runtime.go`) | handoff, timeline, session list, MCP payloads |
| `session_refinements.summary` | `traceary session refine` (#1673) | nothing yet |

#1672 removes the second writer of `sessions.summary` by routing the compact digest through the #1673 write port instead. At that point `sessions.summary` is fed only by an explicit `session end --summary`, which almost never carries text — the measured store has **4 summaries across 26,728 sessions**, and those 4 came from the compact-hook path that #1672 is moving.

So after wave 1 the readers are pointed at a column that is, in practice, empty, while the column that actually accumulates summaries has no readers. That is not a small inconsistency — it is the whole point of the wave failing to reach the surfaces that would show it.

## Proposal

Make `session_refinements` the single carrier and retire `sessions.summary` as a summary source.

1. Point every reader at the refinement: handoff, timeline, `session list`, and the MCP payloads that currently surface `sessions.summary`.
2. Decide what `session end --summary` means. Either it writes a refinement (a `generation`-1 write covering the whole session) or it stops accepting text. Writing a second carrier is not an option.
3. Remove `SetSummaryIfEmpty` from the session use case once #1672 has removed its last caller — it encodes first-write-wins, which #1673 exists to replace.
4. Drop the column, or keep it only if something needs a session-end note that is explicitly *not* a summary. If it stays, rename it so it stops being read as one.

## Why it is a separate issue

#1673 is the storage and the write port. Repointing every reader touches handoff, timeline, list and MCP — a different blast radius and a different set of tests. Bundling them would repeat the scope inflation diagnosed on #1682 (a bug fix that reached +1,817/−214 across 22 files because four extra concerns rode along).

## Depends on

#1673 (introduces the refinement) and #1672 (removes the compact-hook writer). Land both, then repoint once.

## Acceptance

- No reader reads `sessions.summary` as a session summary
- `session end --summary` either writes a refinement or no longer accepts text, with the choice recorded
- `SetSummaryIfEmpty` is gone
- The column is dropped, or renamed with a stated reason for keeping it
- Handoff, timeline, `session list` and MCP all show a refinement when one exists

Parent: #1606

duck8823:
## Summary

Retire `sessions.summary` as a session-summary source after #1672 / #1673.

- list / tree / lineage / handoff read `session_refinements.summary`.
- `session end --summary` writes a refinement (`covers_to` = the end event).
- `SetSummaryIfEmpty` is removed. Post-compact already refined; it no longer dual-writes the column.
- The `sessions.summary` column stays. Dropping it is data-dependent offline; implicit migrate on a populated store is refused (#1852).

See `docs/research/sessions-summary-retirement.md`.

## Test plan

- [x] `go test ./...`
- [x] leftover `sessions.summary` is not listed; `session end --summary` is
- [x] empty end does not clobber leftover column text

Closes #1706.
Leaves parent #1904 open.


duck8823:
gpt-5.6-terra medium: CHANGES_REQUESTED (HIGH declined)

```json
{
  "reviewer": "gpt-5.6-terra",
  "verdict": "CHANGES_REQUESTED",
  "findings": [
    {
      "severity": "HIGH",
      "file": "application/usecase/session_usecase_impl.go:349",
      "title": "End-summary refinement is not idempotent with hook redelivery"
    }
  ]
}
```

Declined: hook session-end / subagent-end call `End(..., "")`. `writeEndRefinement` no-ops on empty summary, so redelivery never uses the unpersisted event id as `covers_to`. Non-empty `--summary` is the CLI path only (not hook-spooled). FinalizeOneShot `already_applied` is the same empty-summary hook path in production.

second reviewer: not run (terra available; opus 5 slug unverified)

Issue #1905, opened 2026-08-14
v0.37.0: finish the search-projection family


duck8823:
# v0.37.0 — finish the search-projection family

This minor is **the search-projection family**, not a generic performance bucket. Hydration (#1730), orphan-consolidation connection count (#1722), and other slowness stay out unless a child of this parent proves they are the same family.

## Already done in v0.36 (do not redo)

- #1819 `recent_source_bytes` verifier
- #1833 lock-wait vs row work

`--recent-age` (#1755) and `fts_logical_bytes` (#1787) belong here.

## Exit

- Every child closed by its own PR. Parent stays OPEN until the tagged `v0.37.0` release.
- Do not treat “the store feels faster” as the gate. Close the listed children.

## Non-goals

- v0.36 leftovers and CLI folds
- Capture loss / destructive dedupe (v0.38)
- Remainder shelf (v0.39)

duck8823:
## Planning (2026-08-14)

Search-projection family only. Sibling parents: #1904, #1906, #1907.

`--recent-age` (#1755) and `fts_logical_bytes` (#1787) are in this list. #1819 / #1833 already sit under #1904.

| # | Title |
|---|---|
| #1736 | v0.37-1: structural searches should not be gated on the literal projection |
| #1737 | v0.37-2: projection invalidation triggers do not watch every column the decoder depends on |
| #1739 | v0.37-3: use the previous complete generation's fingerprints while a rebuild is in progress |
| #1746 | v0.37-4: make the timeline summary candidate walk exact and cheap |
| #1751 | v0.37-5: make the search projection's capacity derivation trustworthy enough to bound the store |
| #1752 | v0.37-6: make obsolete-generation replacement atomic and generation-fenced |
| #1753 | v0.37-7: decide and hold a disk promise for the two-generation rebuild peak |
| #1754 | v0.37-8: separate capacity semantics from execution tuning in the projection config hash |
| #1755 | v0.37-9: retire --recent-age if the byte budget always binds first |
| #1756 | v0.37-10: measure whether the session tier needs a real index before the deferral becomes permanent |
| #1787 | v0.37-11: fts_logical_bytes counts integer columns as decimal digit counts |
| #1807 | v0.37-12: recent-cutoff derivation times out precisely on the stores that need it |
| #1822 | v0.37-13: read session hits and projection readiness from one snapshot |
| #1825 | v0.37-14: Rework projection cleanup addressing before any WITHOUT ROWID conversion |
[...]

PR #1930, opened 2026-08-15
fix: do not gate structural search on the literal projection

### LINKED ISSUE #1736: v0.37-1: structural searches should not be gated on the literal projection

duck8823:
## Summary

Closes #1736

Leaves parent #1905 open.

Empty-query (structural) search already ran against canonical `events` / `command_audits` and never read `literal_search_fingerprints`. This PR pins that contract and applies the same criteria sanity checks (offset, limit, from/to, page window) before the structural branch so invalid criteria still fail closed.

## Design note

- **Requirement:** a structural search under `authority='tiered'` must succeed while the literal generation is rebuilding / stale / missing / empty. It must still refuse invalid offset, limit, and from/to. Non-empty queries stay on the fail-open decode walk from #1718.
- **Boundary:** `searchTieredMetadataTx` validates criteria, then answers empty queries via `queryStructuralEventIDs` *before* reading projection state. The structural SQL does not join `search_projection_source_sequence`.
- **Refuse vs availability:** projection unreadiness is not a structural refusal. Criteria sanity is.

## Tests

- `TestTieredAuthorityStructuralSearchIgnoresLiteralProjectionState` — empty query + workspace filter returns the two matching events while literal is rebuilding, stale, missing, or has an empty generation. Both `SearchMetadata` and `SearchPage`.
- `TestTieredAuthorityStructuralSearchRefusesInvalidCriteria` — negative offset, limit 0, and from>to still error; the message is not `incomplete or stale`.
- Existing `TestTieredAuthoritySearchAnswersWhenProjectionUnusable` still covers non-empty fail-open.

```
go test ./infrastructure/sqlite/ -count=1 -run 'TestTieredAuthorityStructuralSearch|TestTieredAuthoritySearchAnswersWhenProjectionUnusable'
```

## Reviewers

Please review as gpt-5.6-terra medium (and a second of claude opus 5 / antigravity gemini 3.7-flash if available).

### DIFF
--- a/CHANGELOG.ja.md
+++ b/CHANGELOG.ja.md
@@ -5,6 +5,11 @@
 このファイルは、Traceary の各リリースで何が入ったかを時系列で追いやすくするための changelog です。  
 release note と同じ粒度で、版ごとの要点だけをまとめています。
 
+## [Unreleased]
+
+### Fixed
+- **空クエリの structural search を literal projection の状態で拒否しない (#1736)** — workspace / session / kind / time のフィルタは、literal generation が rebuilding / failed / drifted でも正本テーブルに対して動きます。無効な offset / limit / from-to はこれまでどおり fail-closed です。非空クエリは generation が使えないとき decode walk に fail-open します。
+
 ## [v0.36.0] - 2026-08-15
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@
 This file summarizes what changed in each Traceary release in chronological order.
 It mirrors the same level of detail as the GitHub release notes, but keeps the history in the repository.
 
+## [Unreleased]
+
+### Fixed
+- **Structural (empty-query) search is not gated on the literal projection (#1736)** — workspace, session, kind, and time filters run against canonical tables even while the literal generation is rebuilding, failed, or drifted. Invalid offset, limit, and from/to still fail closed. Non-empty queries still fail open onto a decode walk when the generation is unusable.
+
 ## [v0.36.0] - 2026-08-15
 
 ### Added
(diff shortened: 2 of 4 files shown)


duck8823:

Acceptance tests drive real EventDatasource.SearchMetadata/SearchPage on a scratch SQLite fixture. Structural empty-query path does not consult literal projection state. Invalid offset/limit/from-to still fail closed.

antigravity sandbox review did not produce a usable verdict (returned sandbox help). CI Test/Lint/Docs/Integrations are green.

PR #1931, opened 2026-08-15
fix: watch decoder columns on search projection invalidators

### LINKED ISSUE #1737: v0.37-2: projection invalidation triggers do not watch every column the decoder depends on

duck8823:
## Summary

Closes #1737

Leaves parent #1905 open.

Search-projection invalidators watched a narrower `UPDATE OF` set than `hydrateEventPayload` reads. A codec-metadata-only mutation left fingerprints describing a superseded decode (silent false negative).

## Design note

- **Requirement:** a codec-only update of a projected event must drift a complete generation and bump `search_projection_source_revision` during rebuild. Decoder columns and invalidators must not be able to drift apart silently. Existing `body` / retention / redaction paths stay the same.
- **Choice:** extend `UPDATE OF` on the three event invalidator families (`search_projection_complete_event_update`, `search_projection_events_update`, `literal_search_event_update`) with `body_codec`, `body_format_version`, `body_plaintext_bytes`, `body_encoded_bytes`, `body_sha256`.
- **`events.id`:** immutable via `BEFORE UPDATE OF id` abort, not an invalidation. Changing id would orphan `search_projection_source_sequence.event_id`.
- **Migration 063:** constant_in_place DROP/CREATE only. No store-sized scan.

## Tests

- Per-column codec-only UPDATE drifts bounded+literal state.
- Complete identity codec tuple drifts and `SearchMetadata` still finds the event (no stale-fingerprint miss).
- Codec-only UPDATE during rebuild bumps `search_projection_source_revision`.
- Pin: the three triggers' `UPDATE OF` lists contain every `eventPayloadDecoderColumns` name.
- `events.id` update aborts.
- Existing body-update drift test is unchanged.

```
go test ./infrastructure/sqlite/ -count=1 -run 'TestTieredAuthorityCodecMetadata|TestTieredAuthorityCompleteIdentityCodecUpdate|TestSearchInvalidatorsWatchEveryDecoderColumn|TestEventsIDIsImmutable|TestBuildPreparedMigrationPlanClassifiesExactSuffix'
```

### DIFF
--- a/CHANGELOG.ja.md
+++ b/CHANGELOG.ja.md
@@ -8,6 +8,7 @@ release note と同じ粒度で、版ごとの要点だけをまとめていま�
 ## [Unreleased]
 
 ### Fixed
+- **search-projection の invalidator が decoder の全列を見る (#1737)** — 投影済み event の codec metadata だけの UPDATE は complete generation を drift させ、rebuild 中は `search_projection_source_revision` を進めます。`events.id` は watch せず DB で不変です。`body` を既に含む body / retention / redaction 経路は変わりません。
 - **空クエリの structural search を literal projection の状態で拒否しない (#1736)** — workspace / session / kind / time のフィルタは、literal generation が rebuilding / failed / drifted でも正本テーブルに対して動きます。無効な offset / limit / from-to はこれまでどおり fail-closed です。非空クエリは generation が使えないとき decode walk に fail-open します。
 
 ## [v0.36.0] - 2026-08-15
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ It mirrors the same level of detail as the GitHub release notes, but keeps the h
 ## [Unreleased]
 
 ### Fixed
+- **Search-projection invalidators watch every decoder column (#1737)** — a codec-metadata-only update of a projected event now drifts a complete generation and bumps `search_projection_source_revision` during rebuild. `events.id` is immutable at the database instead of being watched. Body / retention / redaction paths that already mention `body` are unchanged.
 - **Structural (empty-query) search is not gated on the literal projection (#1736)** — workspace, session, kind, and time filters run against canonical tables even while the literal generation is rebuilding, failed, or drifted. Invalid offset, limit, and from/to still fail closed. Non-empty queries still fail open onto a decode walk when the generation is unusable.
 
 ## [v0.36.0] - 2026-08-15
--- a/infrastructure/sqlite/event_search_authority_internal_test.go
+++ b/infrastructure/sqlite/event_search_authority_internal_test.go
@@ -1105,6 +1105,176 @@ func TestTieredAuthorityBodyUpdateOfProjectedEventDriftsAndAnswersMutated(t *tes
 	}
 }
 
+// TestTieredAuthorityCodecMetadataUpdateDriftsProjectedEvent is #1737:
+// a codec-column-only UPDATE must drift a complete generation so stale
+// fingerprints cannot silently exclude the row.
+func TestTieredAuthorityCodecMetadataUpdateDriftsProjectedEvent(t *testing.T) {
+	ctx := context.Background()
+	base := time.Date(2026, 3, 1, 12, 0, 0, 0, time.UTC)
+
+	updates := []struct {
+		name string
+		sql  string
+	}{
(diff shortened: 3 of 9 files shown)


duck8823:
gpt-5.6-terra medium: ✅ APPROVE — 問題なし (after addressing two HIGH: decoder SELECT now built from eventPayloadDecoderColumns; schema version expectations updated to 63). CI Test/Lint/Docs/Integrations green on 42cb1aee.

PR #1932, opened 2026-08-15
fix: return session hits and readiness from one snapshot

### LINKED ISSUE #1822: v0.37-13: read session hits and projection readiness from one snapshot
Raised by gpt-5.6-sol while reviewing #1727 (PR #1820) and deliberately left out of it.

## The defect

`search` (MCP) resolves the session group in two separate reads:

```go
hits, err := s.projectionSessionSearch.SearchSessionHits(ctx, criteria, exclude)   // read-only tx #1
if len(hits) == 0 {
    ready, err := s.projectionSessionSearch.SearchSessionProjectionReady(ctx)      // read-only tx #2
    if !ready { reasons = append(reasons, "session_projection_not_ready") }
}
```

`presentation/mcpserver/server.go:933-953`.

The two transactions can observe different states, so the reason can contradict the hits it is explaining:

- **A generation completes between them.** The search ran while no generation was published, so it returned empty; readiness then answers `true`. The client receives an empty `sessions` array with **no reason at all** — indistinguishable from "the projection was consulted and nothing matched", for a query that would have matched. Transient and self-healing on the next call, but the response is a false negative with no signal attached.
- **A rebuild starts between them.** The projection genuinely had no match; readiness now answers `false`; the response claims `session_projection_not_ready` for a search that *was* served. Misleading rather than lossy.

Both windows are narrow — publication happens once per generation — which is why this was not treated as a blocker for #1727.

## Why the datasource already knows

`SearchSessionHits` opens one read-only transaction and calls `searchProjectionReadReady(ctx, tx)` inside it before deciding to return an empty page (`infrastructure/sqlite/search_projection_session_query.go:58-67`). The readiness answer that the caller needs is **already computed in the same snapshot as the hits** and then thrown away, which is why the caller has to go and ask for it again in a second transaction.

## What to decide

Return both from one call, so the pair is snapshot-consistent by construction rather than by luck:

```go
type SearchSessionPage struct {
    Hits  []SearchSessionHit
    State SearchSessionTierState  // ready | not_ready | not_applicable
}
```

The third state matters: `SearchSessionHits` returns early — before opening the database at all — when the query is empty, when `--kind` is set (the tier fails closed, see #1728), and when the caller has paged past the first page. In those cases the tier was never consulted, so neither "ready" nor "not ready" is an honest answer, and today's code reports whichever readiness happens to say.

Doing this retires `ProjectionSessionSearchReadiness` as a separate port, so the surface goes down rather than up. It touches the CLI as well, which holds the same query port (`presentation/cli/root.go:21`) — that is why it was not folded into #1727.

## What #1727 did fix

The related and more dangerous half: readiness was an **optional** interface, obtained by type assertion, defaulting to `ready = true` when the assertion failed. Removing the method from the concrete datasource would have made MCP silently stop reporting `session_projection_not_ready`, with nothing failing at build or test time. #1727 made readiness a required part of the port MCP holds (`queryservice.ProjectionSessionSearch`) and asserted it at the implementation (`var _ queryservice.ProjectionSessionSearch = (*EventDatasource)(nil)`), so that regression is now a compile error.

## Related

- #1727 / PR #1820 — added the session group and the two reasons to MCP `search`.
- #1728 — the `--kind` fail-closed path that motivates the `not_applicable` state.
- #1717 — made the bounded projection authoritative for search.

duck8823:
## Summary

Closes #1822

Leaves parent #1905 open.

Session-tier search asked for hits in one read transaction and readiness in another. A generation completing or a rebuild starting between them made `session_projection_not_ready` (and the CLI stderr notice) contradict the page.

## Design note

- **Requirement:** hits and the reason for an empty page must come from one snapshot. Empty query, `--kind`, and later pages must not pretend the tier was consulted.
- **Port:** `ProjectionSessionSearch` is now `SearchSessionPage` only. `ProjectionSessionSearchQuery` and `ProjectionSessionSearchReadiness` are gone.
- **States:** `ready` (consulted), `not_ready` (refused in that tx), `not_applicable` (never opened the store).
- **CLI:** `traceary search` uses `page.State()` for the not-ready notice. `--kind` takes readiness from the kind-less probe's page, not a third transaction.
- MCP `search` is already retired; the remaining consumer is the CLI.

## Tests

- `TestSearchSessionPageReportsStateFromTheSameSnapshot` — real EventDatasource: ready+hits, ready+empty, empty query `not_applicable`, idle generation `not_ready`.
- Existing kind/offset paths assert `not_applicable`.
- CLI stub implements `SearchSessionPage`; not-ready notice still appears; page errors fail the search.

```
go test ./infrastructure/sqlite/ ./presentation/cli/ -count=1 -run 'TestSearchSessionPageReportsStateFromTheSameSnapshot|TestSearchProjectionReadPath|TestRootCLI_Search'
```

### DIFF
--- a/CHANGELOG.ja.md
+++ b/CHANGELOG.ja.md
@@ -8,6 +8,7 @@ release note と同じ粒度で、版ごとの要点だけをまとめていま�
 ## [Unreleased]
 
 ### Fixed
+- **session-tier の hits と readiness を同じ snapshot から返す (#1822)** — `SearchSessionPage` が 1 つの read transaction で hits と `ready` / `not_ready` / `not_applicable` を返します。空クエリ、`--kind`、2 ページ目以降は store を開かず `not_applicable` です。`SearchSessionProjectionReady` は廃止し、空ページと後から読んだ generation の readiness を組み合わせられないようにします。`traceary search` の session notice もこの page を使います。
 - **search-projection の invalidator が decoder の全列を見る (#1737)** — 投影済み event の codec metadata だけの UPDATE は complete generation を drift させ、rebuild 中は `search_projection_source_revision` を進めます。`events.id` は watch せず DB で不変です。`body` を既に含む body / retention / redaction 経路は変わりません。
 - **空クエリの structural search を literal projection の状態で拒否しない (#1736)** — workspace / session / kind / time のフィルタは、literal generation が rebuilding / failed / drifted でも正本テーブルに対して動きます。無効な offset / limit / from-to はこれまでどおり fail-closed です。非空クエリは generation が使えないとき decode walk に fail-open します。
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ It mirrors the same level of detail as the GitHub release notes, but keeps the h
 ## [Unreleased]
 
 ### Fixed
+- **Session-tier hits and readiness share one snapshot (#1822)** — `SearchSessionPage` returns hits plus `ready` / `not_ready` / `not_applicable` from a single read transaction. Empty query, `--kind`, and later pages report `not_applicable` without opening the store. `SearchSessionProjectionReady` is gone so callers cannot pair an empty page with a later generation's readiness. `traceary search` uses that page for the session notice.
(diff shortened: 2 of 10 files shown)


duck8823:

SearchSessionPage returns hits + ready/not_ready/not_applicable from one snapshot. SearchSessionProjectionReady is gone. CI Test/Lint/Docs/Integrations green on 1680e12d.

PR #1933, opened 2026-08-15
fix: read --kind session-tier notices from one snapshot

### LINKED ISSUE #1859: v0.37-20: session-tier hits and readiness are read in separate transactions
Split out of PR #1856 (#1844) rather than widening it. Raised by gpt-5.6-sol reviewing that PR.

## What

A surface that needs both "did the session tier match anything" and "was the session tier consulted at all" makes two calls in two separate read transactions:

- `SearchSessionHits` (`infrastructure/sqlite/search_projection_session_query.go:55-81`) opens its own `BeginTx`, checks `searchProjectionReadReady` inside it, and returns an empty slice if not ready
- `SearchSessionProjectionReady` (`:87-105`) opens a second `BeginTx` and re-checks

A generation transition committing between the two produces an answer describing two different states:

- refused → complete between the calls: the tier really was refused, readiness now says ready, and the notice is dropped
- consulted-and-empty → rebuilding: the tier really was consulted, readiness now says not ready, and a false notice appears
- with `--kind`, the kind-less probe is a third transaction, so a transition before it can produce the kind-suppression notice and a stale not-ready notice together

Both consumers are affected — `traceary search` on stderr (#1844) and the MCP `search` tool's `session_projection_not_ready` reason.

## Why it is not urgent

The window is one commit wide. Both failure directions are advisory only: stdout, the JSON contract, and the exit code are unaffected, and the events themselves are always correct. Automatic catch-up on store open does mean transitions happen without an operator doing anything, so the window is not zero.

## Shape

Return hits and disposition from a single read transaction, so an empty page arrives already labelled as "consulted and matched nothing" or "refused" rather than being re-derived afterwards. That is a port change: `ProjectionSessionSearchQuery` would carry the disposition on its result and `ProjectionSessionSearchReadiness` would stop being a separate round trip for this purpose.

Worth doing at the same time as any other change to that port, and worth checking whether the readiness call has any remaining caller that genuinely needs it standalone.

duck8823:
## Summary

After #1822, `traceary search --kind` still called `SearchSessionPage` twice. The kind-set call is always `not_applicable` and does not open the store, so it cannot contribute hits or readiness. This PR skips that call and keeps only the kind-less probe.

A parked-failed scratch store is searched twice (with and without `--kind`): both emit one not-ready notice, `--kind` does not add kind-suppression, and two runs agree.

## Test plan

- [x] `go test ./presentation/cli/ -run 'TestRootCLI_SearchKind|TestRootCLI_SearchParkedProjection|TestRootCLI_SearchWithoutKind'`
- [x] `go tool golangci-lint run ./presentation/cli/`
- [ ] CI

Closes #1859
Leaves parent #1905 open


duck8823:

Antigravity gemini: skipped (sandbox headless denied `command` permission).

PR #1934, opened 2026-08-15
fix: snapshot generation-scoped search projection status fields

### LINKED ISSUE #1839: v0.37-17: search projection status mixes generations across its statements

duck8823:
## Summary

`SearchProjectionStatus` re-resolved `active_generation_id` in each generation-scoped statement. A cutover between those statements mixed counters from two generations.

This PR reads those fields inside one read-only transaction after resolving `active_generation_id` / `generation_id` once. A test commits a cutover between the id read and the counters and asserts `recent_documents`, `summary_sessions`, `keyword_rows`, `fingerprint_rows`, `recent_source_bytes_measured`, and exclusions stay on the pre-cutover generation.

The dbstat `physical_bytes` walk stays outside that transaction. `selectSearchProjectionRecentRangeSQL` is not in tree; `recent_cutoff_norm` comes from the same state-row snapshot.

## Test plan

- [x] `go test ./infrastructure/sqlite/ -run 'TestSearchProjectionStatusKeepsGenerationScopedFieldsOnOneSnapshot|TestSearchProjectionStatusReportsRebuildingGenerationExclusions'`
- [x] `go tool golangci-lint run ./infrastructure/sqlite/`
- [ ] CI

Closes #1839
Leaves parent #1905 open


duck8823:

PR #1935, opened 2026-08-15
fix: walk every timeline summary candidate without a depth cap

### LINKED ISSUE #1746: v0.37-4: make the timeline summary candidate walk exact and cheap
Follow-up from #1685 / PR #1745, raised by the adversarial review of the timeline summary rewrite.

## What #1745 changed

`list_timeline_blocks.sql` used to rank summary candidates with `TRIM(body) != ''` and hand Go exactly one id per kind. That predicate reads the stored bytes, so it cannot see through a compressed body — and SQLite's `TRIM` does not strip tabs or newlines either. Either way a blank row took `rn = 1` and the block lost its summary, because Go never saw a second candidate.

The query now returns up to **three** ranked candidate ids per kind, and `firstNonBlankCandidate` (`infrastructure/sqlite/event_datasource.go`) decodes them in order and keeps the first non-blank one.

## What is still not exact

The depth is a cap, not a guarantee.

1. **A kind whose first three candidates are all blank falls through to the next kind** rather than to its own fourth candidate. Before the backfill, `TRIM` on plaintext would have skipped all of them and found the real one. Reaching this needs three whitespace-only bodies of the same kind in one block, each long enough that zstd shrinks it, all ahead of the real summary — unlikely, but the cap is real and silent.

2. **Worst-case hydration cost rose.** Only the first non-blank candidate per kind is decoded, so the usual cost is unchanged at one hydration per kind. But when candidates *are* blank it is up to 9 per (block, workspace) row instead of 3, and `loadEventPlaintext` issues a schema check plus a body read per call — so up to 18 queries per row against the previous 6. The block `limit` bounds blocks, not workspace rows, so a store with many workspaces in one block pays it per row. The per-call schema check is a pre-existing N+1.

## What would make it exact

Either of:

- an unbounded candidate list per kind (paginate the ranked ids and let Go keep asking for the next one), or
- a blankness signal that survives encoding, so SQL can rank without reading the body at all.

The second is the better shape and overlaps with #1686 (retire `event_metadata_projection`) — whatever replaces the projection is where a derived "body is blank" bit would live. Worth deciding together rather than bolting a column onto a table that is being retired.

Hoisting the schema check out of `loadEventPlaintext` is independent and cheap, and would cut the query count in half on every hydration path, not just this one.

## Acceptance

- The timeline summary is the first non-blank candidate of its kind, with no depth cap.
- No regression in query count per (block, workspace) row against `main`.
- A test with more blank candidates than the current cap.

duck8823:
## Summary

`list_timeline_blocks.sql` ranked at most three summary candidates per kind. Four compressed whitespace bodies of the same kind hid a later real summary.

This PR returns the unbounded ranked id list (`json_group_array`) and keeps the first non-blank decode. `TRIM(body) != ''` still drops identity whitespace in SQL. The `body_codec` schema check is once per `ListTimelineBlocks`, so a one-candidate row stays at one schema read plus one body read (not worse than main).

A durable blankness column is out of scope (write path + backfill).

## Test plan

- [x] `go test ./infrastructure/sqlite/ -run 'TestTimelineSummary|TestDatasource_ListTimelineBlocks'`
- [x] `go tool golangci-lint run ./infrastructure/sqlite/`
- [ ] CI

Closes #1746
Leaves parent #1905 open

### DIFF
--- a/CHANGELOG.ja.md
+++ b/CHANGELOG.ja.md
@@ -8,6 +8,7 @@ release note と同じ粒度で、版ごとの要点だけをまとめていま�
 ## [Unreleased]
 
 ### Fixed
+- **timeline の summary candidate walk を exact かつ cheap にする (#1746)** — kind ごとの ranked id は上限なし（`json_group_array`）。Go は decode 後の最初の非 blank を残すので、圧縮された blank が 4 件あっても同じ kind の後続 prompt を隠しません。`body_codec` の schema check は candidate ごとではなく `ListTimelineBlocks` あたり 1 回です。
 - **search projection status の generation 付き field を 1 snapshot にする (#1839)** — `recent_documents` / `summary_sessions` / `keyword_rows` / `fingerprint_rows`、rebuild 側の `recent_source_bytes` と exclusions は、1 つの read-only transaction の中で `active_generation_id` / `generation_id` を 1 度だけ解決します。dbstat の `physical_bytes` はその transaction の外です。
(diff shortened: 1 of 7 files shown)


duck8823:

Issue #1906, opened 2026-08-14
v0.38.0: stop losing and destroying records


duck8823:
# v0.38.0 — stop losing and destroying records

Two kinds of “the store is not what happened”:

1. Capture never wrote it (hooks, Log, transcript snapshots).
2. A later writer deleted or froze it (destructive dedupe).

#1777 and #1862 sit here, not in the v0.39 remainder.

## Exit

- Every child closed by its own PR. Parent stays OPEN until the tagged `v0.38.0` release.

## Non-goals

- Search-projection family (v0.37)
- Remainder: tests, gc lexical compare, dead retention-plan library, release-gate evaluator (v0.39)

duck8823:
## Planning (2026-08-14)

Capture loss plus destructive dedupe. Sibling parents: #1904, #1905, #1907.

#1777 and #1862 are here on purpose (they destroy records; they are not remainder).

| # | Title |
|---|---|
| #1696 | v0.38-1: guard the spooled transcript replay path against redelivery |
| #1697 | v0.38-2: Supersede the earlier snapshot when a Kimi transcript turn grows |
| #1710 | v0.38-3: Log returns an event it may not have persisted |
| #1713 | v0.38-4: Grok Stop path can drop a turn: readiness check and recording read the wire log twice |
| #1716 | v0.38-5: hook state marker keys collide after sanitisation |
| #1777 | v0.38-6: store dedupe can permanently freeze a session refinement by deleting a coverage endpoint |
| #1862 | v0.38-7: store dedupe content-events can cascade-delete a command audit |

Implementation PRs: `Closes #<child>` only. Leave parent #1906 open.

PR #1954, opened 2026-08-15
fix: supersede earlier Kimi transcript when a turn grows

### LINKED ISSUE #1697: v0.38-2: Supersede the earlier snapshot when a Kimi transcript turn grows
## Problem

`#1681` collapses Kimi `Stop` redeliveries by keying idempotency on `(session_id, wire turn ID, content fingerprint)`. That is correct — Kimi's `Stop` re-fires while a turn is still streaming, so keying on the turn ID alone would freeze the first, truncated snapshot and discard everything the turn produced afterwards.

The consequence is that a turn which grows still writes one row per distinct fingerprint. Measured on the live store (agent=kimi, kind=transcript, 2026-08-01..07, 52,475 consecutive same-session pairs):

- 49,718 pairs identical — collapsed to one row by #1681
- **2,053 pairs where the same turn gained blocks** — each of these still produces an extra row
- 153 of those gained the turn's only `"type":"text"` block

So ~24 firings per turn become roughly 2–3 rows instead of 1. Only the last of those rows is the complete turn; the earlier ones are strict prefixes that carry no information the last row lacks.

These partial snapshots are **not** collapsible by `traceary store dedupe content-events` (#1682): that command groups by exact body identity, and a prefix is not identical to its extension.

## What is needed

When a firing's turn ID matches the marker but its fingerprint differs, the earlier row for that same turn should be superseded rather than left alongside the new one.

`runHookKimiTranscript` deliberately does not do this today because the hook has no delete port — `EventUsecase` only exposes `Log`. Superseding needs one of:

1. a narrow delete/replace port on the event write side, keyed on the event id the previous firing wrote (the marker would have to carry that id as a third field), or
2. an update path that rewrites the existing row's body in place.

Option 1 keeps the write side append-only in spirit and makes the supersede auditable; option 2 is cheaper but mutates a recorded event, which conflicts with how every other write path behaves. Prefer 1 unless reading the code says otherwise.

## Scope

Kimi only, same as #1681. Other hosts do not re-fire mid-turn (measured repeat ratios: codex 1.8x, claude 1.1x, antigravity 1.3x, grok 1.0x).

## Acceptance

- A turn that grows across N firings leaves exactly one row in `events`, holding the final content.
- A turn that never grows still leaves exactly one row (no regression on the #1681 behaviour).
- Every failure mode still fails open: a supersede that cannot be performed must leave the newly recorded row in place rather than drop the turn.
- Regression test: fire `stop`, append a block under the same turn ID, fire again, assert one row whose body contains the appended content.

duck8823:
## Summary
- Growing Kimi turns now leave one final `kind=transcript` row: the new body is written first, then the previous event id from the turn marker is deleted.
- Two-field `#1681` markers and a failed delete fail open (the new row stays; the old row is left).
- Marker format is now `turnID\tfingerprint\teventID`.

Closes #1697
Leaves parent #1906 open

## Test plan
- [x] `go test ./...`
- [x] `TestRootCLI_HookKimiStop_SupersedesGrowingTurnOnScratchStore` fires `hook kimi stop`, appends same-turn text, fires again, and asserts one decoded transcript row containing the appended text
- [x] Stub cases: growth supersedes, two-field marker cannot supersede, delete failure keeps both rows, unadvanced turn still collapses to one row

### DIFF
--- a/CHANGELOG.ja.md
+++ b/CHANGELOG.ja.md
@@ -8,6 +8,7 @@ release note と同じ粒度で、版ごとの要点だけをまとめていま�
 ## [Unreleased]
 
 ### Fixed
+- **成長する Kimi turn は最終 transcript 1 行を残す (#1697)** — 同じ wire turn ID で fingerprint が変わったあとの Stop は長い本文を先に書き、marker の前の event id を消す。delete 失敗と 2 フィールドの `#1681` marker は fail-open（新しい行は残し、古い行は残す）。scratch: `hook kimi stop` のあと、同じ turn の長い payload では `kind=transcript` が 1 行になり、追記テキストを含む。
 - **spool と generic な Kimi transcript 記録が turn guard を継承する (#1696)** — `hook transcript kimi` と `command=transcript` の spool replay は、`hook kimi stop` と同じ (session, wire turn, fingerprint) ガードを通る。live store（`mode=ro`、2026-08-15）: `source_hook=stop` / `kind=transcript` / `agent=kimi` は 249,042 件（`client` は `hook`）。`hook_deliveries` に Kimi の `stop` 行はないので spool と live はそこでは分けられない。fail-open と成長 turn の記録は変えない。
 
 ## [v0.37.0] - 2026-08-15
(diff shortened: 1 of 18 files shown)


duck8823:

PR #1955, opened 2026-08-15
fix: Log returns the canonical id on exact redelivery

### LINKED ISSUE #1710: v0.38-3: Log returns an event it may not have persisted
## Problem

`EventUsecase.Log` returns the `*model.Event` it constructed, unconditionally. The persistence layer underneath does not always store that event: `persistEventDelivery` (`infrastructure/sqlite/event_delivery_store.go`) detects an exact redelivery by delivery fingerprint, records an attempt against the **existing** row, and returns `inserted = false` without inserting.

So on redelivery `Log` hands the caller an event ID that names no row. Nothing in the signature says this can happen.

Today that is harmless for every caller, because callers use the return value for logging shape rather than as a foreign reference. #1672 is the first place the returned ID becomes a foreign key: the post-compact digest passes it to `session_refine` as `covers_to`.

## Why it is not urgent

Verified for #1672 (PR #1709): no host's compact payload carries a field on the proven delivery-ID allowlist (`provenHookDeliveryIDFields`, `presentation/cli/hook_resolve.go:196` — claude/kimi `tool_use_id`, grok `tool_use_id`/`prompt_id`, codex `event_id`, gemini `timestamp`). With no native ID, `attachHookDelivery` attaches no evidence and the exact-redelivery branch is unreachable, so a compact event always inserts.

If a host later adds one of those fields to its compact payload, the refinement use case rejects the unknown `covers_to` before writing anything, the hook warns, and the range falls to the #1683 gc fallback. Degraded, not corrupt.

## Scope

Make the contract say what happens:

- `Log` reports whether the event was newly inserted, and on an exact redelivery returns the **canonical existing** event ID rather than the discarded new one.
- Same treatment for `LogWithAudit` / the session-boundary paths that share `saveEventTransaction`.
- Audit the existing callers: any that would now want the canonical ID.

## Done when

- A caller can tell an insert from a redelivery without reaching into the infrastructure layer
- An integration test against real SQLite proves a redelivered hook returns the first event's ID
- `#1672`'s comment in `applyPostCompactDigest` describing this hazard can be deleted

Follow-up from #1672 / PR #1709. Parent sprint: #1606

duck8823:
## Summary
- `EventUsecase.Log` and `Audit` now return `EventWriteResult` so a caller can tell an insert from an exact hook redelivery without reading infrastructure.
- Shared `saveEventTransaction` stamps the persist outcome on the in-memory event and rebinds the discarded constructed id to the existing row.
- Session Start/End reuse that same stamp, so a redelivered boundary returns the first event's id.
- Removed the #1672 hazard comment on `applyPostCompactDigest`.

Closes #1710
Leaves parent #1906 open

## Test plan
- [x] `go test ./...`
- [x] `TestEventUsecase_Log_ExactRedeliveryReturnsCanonicalEventID` drives real `EventUsecase.Log` on scratch SQLite with the same `event_id:` delivery twice
- [x] Audit exact redelivery returns the first id; session Start redelivery returns the same id

### DIFF
--- a/CHANGELOG.ja.md
+++ b/CHANGELOG.ja.md
@@ -8,6 +8,7 @@ release note と同じ粒度で、版ごとの要点だけをまとめていま�
 ## [Unreleased]
 
 ### Fixed
+- **Log は insert と exact redelivery を報告し、既存の正規 id を返す (#1710)** — `EventUsecase.Log` / `Audit` は `EventWriteResult` を返す。共有の `saveEventTransaction` が persist 結果を in-memory の event に載せるので、redelivery は捨てた構築 id ではなく既存行の id を返す。scratch: 同じ `event_id:` delivery の `Log` 2 回で `events` は 1 行、2 回目の id は 1 回目と一致する。
 - **成長する Kimi turn は最終 transcript 1 行を残す (#1697)** — 同じ wire turn ID で fingerprint が変わったあとの Stop は長い本文を先に書き、marker の前の event id を消す。delete 失敗と 2 フィールドの `#1681` marker は fail-open（新しい行は残し、古い行は残す）。scratch: `hook kimi stop` のあと、同じ turn の長い payload では `kind=transcript` が 1 行になり、追記テキストを含む。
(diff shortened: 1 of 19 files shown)


duck8823:

Issue #1969, opened 2026-08-16
v0.40-16: main Test job fails on attestation concurrent-append flake


duck8823:
v0.40-16: main Test job fails on attestation concurrent-append flake

Parent: duck8823/traceary#1968

## Observed

`main` CI Test is red after the v0.39.0 tag. Docs / Lint / Integrations pass.

| Run | SHA | When |
|---|---|---|
| https://github.com/duck8823/traceary/actions/runs/31910198208 | `eddcbe09` (v0.39.0 release-prep merge) | 2026-08-15 |
| https://github.com/duck8823/traceary/actions/runs/31910511600 | `3dfc21f2` (Homebrew formula merge #1967) | 2026-08-15 |

Both fail only this test:

```
--- FAIL: TestVerifyAttestationChain_ConcurrentAppendDoesNotFalseFail
    attestation_store_test.go:344: concurrent verify/append error = inspect table presence: database is locked (261)
FAIL	github.com/duck8823/traceary/infrastructure/sqlite
```

CI on `d6e4fcea` (#1965) was green. This is contention, not a v0.39.0 product regression.

The same failure showed up locally during v0.39 work and passed on retry.

## Cause (from the test, not guessed beyond the log)

`TestVerifyAttestationChain_ConcurrentAppendDoesNotFalseFail` (`infrastructure/sqlite/attestation_store_test.go`) runs 4 `EventDatasource.Save` goroutines and 8 `VerifyAttestationChain` goroutines against one file. Verify opens a second `sql.Open` connection (`openAttestationDB`) that does not set a busy timeout. `inspect table presence` then returns SQLITE_BUSY (261).

The test intends to prove concurrent append does not make verify false-fail. SQLITE_BUSY is not that invariant; it is lock contention on a second connection.

## Done when

- `go test ./infrastructure/sqlite/ -run TestVerifyAttestationChain_ConcurrentAppendDoesNotFalseFail -count=50` is green
- Concurrent verify still fails only on a real chain break, not on SQLITE_BUSY
- `main` CI Test is green on a subsequent push (or an equivalent retry of the current SHA after the test change lands)

Leaves parent #1968 open.

Issue #1968, opened 2026-08-16
v0.40.0: remaining shelf and main CI flake


duck8823:
# v0.40.0 — remaining v0.39 children and main CI

The v0.39.0 tag shipped four children. The leftover fifteen `v0.39-N:` issues and the red Test job on `main` belong here. No subset. Implementation PRs close children only.

## Exit

- Every child closed by its own PR.
- This parent stays OPEN until the tagged `v0.40.0` release.

## Children

Filled in a follow-up comment after retitles.

## Non-goals

- Restoring the retired `store gc` CLI. #1721 / #1724 remap onto `CollectGarbage` / compact.
- The 97→29 CLI cut (never filed as a v0.39-N child).

duck8823:
## Children (complete set)

No subset. Implementation PRs close these children only. Parent #1968 stays open through those PRs and the release-prep PR.

| # | Title |
|---|---|
| #1699 | v0.40-1: Dedupe identification aborts entirely on a single undecodable body |
| #1701 | v0.40-2: content-event-reliability diagnostic reports duplicates the repair will never touch |
| #1702 | v0.40-3: A failed dedupe apply drops the run id that would recover its committed batches |
| #1721 | v0.40-4: let gc delete during a long consolidation drain by folding oldest ranges first |
| #1722 | v0.40-5: orphan consolidation opens one read-only connection per candidate |
| #1724 | v0.40-6: long-lived active sessions are never consolidated but their old events are deleted |
| #1730 | v0.40-7: batch command audit payload hydration |
| #1744 | v0.40-8: dedupe quarantine archive stores decoded plaintext |
| #1760 | v0.40-9: body_codec accepts unknown values that only fail at read time |
| #1762 | v0.40-10: retention plan blames active_session for exclusions it does not cause |
| #1763 | v0.40-11: retention apply rechecks against the wall clock instead of the plan cutoff |
| #1780 | v0.40-12: payload rehearsal aborts intermittently on WAL autocheckpoint |
| #1797 | v0.40-13: body discard is transcript-only |
| #1815 | v0.40-14: normalize remaining lexical timestamp ordering on payload_backfill_runs |
| #1858 | v0.40-15: payload-rehearsal rollback verification has no wall-time budget |
| #1969 | v0.40-16: main Test job fails on attestation concurrent-append flake |

#1721 / #1724 still say `store gc` in the body. That CLI is retired. Work them on `CollectGarbage` / compact, do not restore the command.

[...]

PR #1978, opened 2026-08-16
fix: keep dedupe identification running after one unreadable body

### LINKED ISSUE #1699: v0.40-1: Dedupe identification aborts entirely on a single undecodable body
## Problem

`identifyDedupeGroups` returns immediately when `decodeDedupeCandidateBody` fails (`infrastructure/sqlite/content_event_dedupe_datasource.go`, at the `decodeDedupeCandidateBody` call site). One oversized or corrupt body anywhere in the eligible set therefore makes the whole repair permanently unrunnable — the scan never reaches the other 246,656 rows.

## Why it is asymmetric

A malformed `created_at` is handled the opposite way: `planContentEventDedupe` collects the group into `plan.skipped` with `contentEventDedupeReasonMalformed` and carries on, and the CLI reports it. Both cases are "this one row cannot be reasoned about safely", but only the timestamp case degrades gracefully.

## Why it did not matter before

The behaviour predates #1682. It was invisible because an apply could not run at full store size at all, so nobody reached the failure. Making the repair runnable makes this the next thing that stops it.

## Direction

Treat a payload-integrity failure the same way as a malformed timestamp: skip the row, record it in `plan.skipped` with its own reason, and let the operator see which rows were passed over.

Narrow the tolerance deliberately — `errors.As` on `*PayloadIntegrityError` only. A decode failure that is not an integrity error (a genuinely broken codec, a schema mismatch) should still abort, because it indicates the store is not in the shape the repair assumes.

## Acceptance

- One unreadable body does not prevent the remaining rows from being planned or applied
- The skipped row appears in the result with a distinguishable reason, not silently
- A non-integrity decode error still aborts

duck8823:
## 概要
Closes #1699

Leaves parent #1968 open.

## 変更内容
- `identifyDedupeGroups` continues after a single `*PayloadIntegrityError` decode.
- That row is skipped with `contentEventDedupeReasonUnreadableBody` (identity suffix `body:unreadable`) and logged; remaining groups are still planned/applied.
- Non-integrity decode errors still abort.

## テスト確認
- [x] `go test ./infrastructure/sqlite/... ./application/usecase/...` (attestation concurrent flake is #1969, already merged)

## Design
Claude Opus 5 medium: `.ai/spec/1699.md`
Implementation: Claude Sonnet 5 medium

Do not close parent #1968.

### DIFF
--- a/infrastructure/sqlite/content_event_dedupe_datasource.go
+++ b/infrastructure/sqlite/content_event_dedupe_datasource.go
@@ -6,6 +6,7 @@ import (
 	"database/sql"
 	"encoding/hex"
 	"errors"
+	"fmt"
 	"log/slog"
 	"sort"
 	"strings"
@@ -26,9 +27,10 @@ import (
 const contentEventDedupeProximityWindow = 10 * time.Second
 
 const (
-	contentEventDedupeReasonProximity = "near-simultaneous hook duplicate"
-	contentEventDedupeReasonStrict    = "strict exact duplicate"
-	contentEventDedupeReasonMalformed = "skipped: malformed or unparseable created_at"
+	contentEventDedupeReasonProximity      = "near-simultaneous hook duplicate"
+	contentEventDedupeReasonStrict         = "strict exact duplicate"
+	contentEventDedupeReasonMalformed      = "skipped: malformed or unparseable created_at"
+	contentEventDedupeReasonUnreadableBody = "skipped: unreadable body"
 )
 
 // dedupeMemberRef is one row's participation in an identity group.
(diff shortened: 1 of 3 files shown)


duck8823:
## Reviews

- Grok 4.6 medium: 問題なし
- Claude Opus 5 medium: HIGH that `store compact` encode/verify still abort on an unreadable available body. That is outside #1699 (identification + dedupe apply). Design already leaves encode/verify abort as existing behavior. Not expanding this PR.

CI: Docs / Integrations / Lint / Test green.

Closes #1699 only. Leaves parent #1968 open.

PR #1980, opened 2026-08-16
fix: align content-event-reliability duplicates with dedupe eligibility

### LINKED ISSUE #1701: v0.40-2: content-event-reliability diagnostic reports duplicates the repair will never touch
## Problem

`docs/storage/README.md` describes the dedupe repair's duplicate identity as "the same identity the `content-event-reliability` diagnostic uses". That was true until #1682 added the retention exclusions; it is no longer.

- The repair excludes rows the retention pruner has emptied (they all carry one fixed marker body, so unrelated rows would hash to the same identity).
- The diagnostic (`presentation/cli/doctor_content_reliability.go`) groups on `normalizeContentEventBody(event.Body())` with no regard for `body_availability`.

So once retention has pruned anything, the diagnostic reports a large phantom duplicate group built entirely from marker bodies, and its own remediation text points the operator at `store dedupe content-events` — which will then correctly refuse to touch any of it. The operator is told there is a problem and handed a tool that reports nothing to do.

#1682 patched the docs to note the divergence, which is honest but not a fix.

## Why it has not bitten yet

Zero rows on the live store are currently `body_availability='unavailable_retention'`. The divergence is latent and becomes visible the first time `store retention apply` runs.

## Direction

Give the diagnostic the same eligibility rule as the repair, so "the diagnostic and the command agree on what counts as a duplicate" is true again rather than aspirational.

Note the *ledger* exclusion is deliberately different in kind and must not be copied: a ledger-held row still participates in grouping (it just cannot be archived), so the diagnostic counting it is correct.

## Acceptance

- A store with retention-emptied rows produces no phantom duplicate group in `doctor`
- The diagnostic's count and the repair's plan agree on the same store
- The divergence caveat is removed from `docs/storage/README.md` and `README.ja.md`

duck8823:
## 概要
Closes #1701

Leaves parent #1968 open.

## 変更内容
- `content-event-reliability` skips retention-emptied (`unavailable`) rows so its duplicate identity matches `dedupeEligibilityFilter`.
- Ledger-held rows still participate (eligibility vs archivability). Docs updated.

## テスト確認
- [x] `go test ./presentation/cli/ ./infrastructure/sqlite/`

## Design
Claude Opus 5 medium: `.ai/spec/1701.md`
Implementation: Claude Sonnet 5 medium

Do not close parent #1968.


duck8823:
Grok 4.6: ✅

Re-reviewed `64ea6911` after the PATH fixture.

- The agreement test now installs `setTracearyPathToCurrentExecutable` before `doctor Execute()`, which is the same host-binary fixture that failed CI on the first push.
- Comments now state this is availability-filter agreement only (genuine available pair vs retention-emptied phantom group). That matches the code: `contentEventDedupeEligible` mirrors `dedupeEligibilityFilter`'s availability predicate and does not claim hydrate-tolerance or ledger/attested/audit-held agreement.
- The earlier Opus HIGH (List hydrate abort; held rows still counted) is therefore out of #1701, not a merge blocker for this PR.

No CRITICAL/HIGH on this diff.

Leaves parent #1968 open.

PR #1983, opened 2026-08-16
docs: document transcript-only discard scope and per-record accrual

### LINKED ISSUE #1797: v0.40-13: body discard is transcript-only, so 2.67 KiB/event of the record accrues permanently
> **Superseded framing (2026-08-11).** Everything below is measured in `GiB/month`, which #1620 has withdrawn as a gate unit: it is `bytes-per-unit-of-work x work-per-month`, and only the first factor is a property of Traceary. The measurements stand; the `<= 0.5 GiB/month` comparison does not. Restated per-record and re-decided in [issuecomment-5249803539](https://github.com/duck8823/traceary/issues/1797#issuecomment-5249803539). The structural finding — the discard allowlist is transcript-only, so 80% of what accrues per record can never be discarded — is unchanged.

Measured during the #1620 gate rehearsal. The growth gate is not "not yet demonstrated" — it is **arithmetically unreachable** with the discard mechanism as shipped, and the shortfall is visible without waiting for a store to age past its retention window.

## The discard allowlist is transcript-only

`infrastructure/sqlite/sql/select_discardable_event_bodies.sql` is the single allowlist for irreversible body discard, composed into the gc count, the gc update, the retention plan and the apply-time recheck. Its first clause is:

```sql
WHERE e.kind = 'transcript'
```

Nothing else is ever discardable. `command_audits` is not reachable from this path at all — no statement in the retention path writes to that table.

## What that leaves growing forever

Stored (post-zstd) bytes accrued in the last 30 days on the rehearsal copy of the live store:

| source | bytes / 30d | GiB | discardable? |
|---|---|---|---|
| `command_audits` (command + input + output text) | 671,137,297 | 0.625 | **no** |
| `events.body` kind `prompt` | 170,284,911 | 0.159 | **no** |
| `events.body` kind `compact_summary` | 2,571,137 | 0.002 | **no** |
| `events.body` session_started / session_ended / note | 329,742 | 0.000 | **no** |
| `events.body` kind `transcript` | 106,767,730 | 0.099 | yes, after the cutoff |
| `events.body` kind `command_executed` | 0 | 0 | n/a — already emptied by #1675 |

**Undiscardable subtotal: 844,323,345 bytes = 0.786 GiB per month**, against a gate of **<= 0.5 GiB/month**. Credit the transcript term at 100% — assume every transcript body is folded and discarded the instant it crosses the cutoff — and the store still grows at 1.57x the budget, permanently.

(The transcript figure is measured after a `--keep-days 7` run and is therefore understated; it does not matter, because it is the one term the design can already remove.)

## Consequence for the store gate

The measured store total is 2.96 GiB with no search projection. At 0.786 GiB/month of undiscardable accumulation, the 4 GiB gate is crossed in **under two months** of ordinary use, on a store that has just had every v0.34 lever applied to it. "Bounded store" is not a property this design has; it has a slower unbounded one.

The improvement is real and should not be understated — 24.71 GiB became 2.96 GiB, and the growth rate fell from +9.9 to +0.79 GiB/month, a 12.5x reduction. But 0.79 is not 0.5, and the gap is not a measurement artefact.

## Why "wait until day 90" does not answer it

The obvious objection is that this store is 2.5 months old, so nothing has crossed the 90-day retention cutoff and the discard term has never fired in anger. That is true, and irrelevant to the arithmetic: the calculation above already grants the discard term its maximum possible value. Ageing the store cannot make an undiscardable byte discardable.

What ageing would change is the transcript term's realised value, which is at most 0.099-0.2 GiB/month. It cannot close a 0.286 GiB/month shortfall.

## What to decide

Not "how to measure this better" — that is settled. The question is which of these the release makes:

1. **Make `command_audits` text discardable.** It is 79% of the undiscardable growth on its own. The obstacle is that it has no equivalent of a session refinement to fold into: `select_discardable_event_bodies.sql` only discards a transcript body because a refinement demonstrably covers its range. Command output would need the same kind of coverage evidence, or a different justification for discarding it (age alone, a per-session aggregate, a size cap on retained output).
2. **Make `prompt` bodies discardable** on the same coverage argument as transcript. Worth 0.159 GiB/month. #1620 examined and rejected *capping* prompt bodies, for good reasons that do not apply to discarding a prompt whose range is already folded into a refinement.
3. **Restate the gate.** If 0.79 GiB/month is what the design delivers, the honest move is to publish that number rather than a 0.5 target it does not meet. This is the cheapest option and the one that costs the most credibility if chosen without first establishing that 1 and 2 are genuinely out of reach.

Options 1 and 2 together bring the floor to about 0.03 GiB/month, i.e. far under the gate — so the gate is reachable, just not by what shipped.

## Related

- #1620 — the gate this measurement belongs to. Its "growth" row was reconciled against a projection of 0.87 GiB/month closing to 0.40 via the transcript discard term; the projection did not account for the discard being transcript-only.
- #1675 — already removed the duplicate `command_executed` body, which is why that row measures 0. `command_audits` is the surviving copy and is now the single largest growth source.
- #1676 — the discard mechanism itself. It works; its scope is the issue.

duck8823:
## Summary

Body discard via `store compact` only targets `transcript` bodies. Prompt bodies and `command_audits` text never enter that path, so ~2.67 KiB/event of the record accrues permanently. This PR documents that as a *reduction*, not a bounded-store guarantee, remaps leftover `gc` / `--target events` wording onto `store compact`, and pins the SQL kind allowlist to `{transcript}`.

Does **not** widen discard scope (issue owner chose option 3).

## Test plan

- [x] `go test ./infrastructure/sqlite/ ./cmd/repo-tooling/ -run 'TestDiscard|TestDocs|TestRepo'`
- [x] `TestSelectDiscardableEventBodiesKindAllowlist` asserts the embedded SQL allowlist is exactly `{transcript}`

Closes #1797

Leaves parent #1968 open.

### DIFF
--- a/docs/storage/README.ja.md
+++ b/docs/storage/README.ja.md
@@ -3,7 +3,7 @@
 [English](./README.md)
 
 Traceary は、ローカル状態を 1 つの SQLite DB ファイルに保存します。
-このガイドでは、何がどこに保存されるのか、現在の schema がどう構成されているのか、`gc` / backup の既定動作が実際に何を意味するのかを整理します。
+このガイドでは、何がどこに保存されるのか、現在の schema がどう構成されているのか、`compact` / backup の既定動作が実際に何を意味するのかを整理します。
 
 ## ローカルファーストの配置
 
@@ -66,7 +66,7 @@ Traceary は確認済みのコマンド構造だけを正規化します。直�
 
 `input_truncated` または `output_truncated` が true の場合、保存済み payload はすでに上限内の head/tail projection であり、新規 row では対応する `*_original_bytes` column に元の byte 数を記録します。検索 index は `command_text` / `input_text` / `output_text` を `events.body` と独立に持つため、合成 body を消しても audit テキストの検索性は失われません。切り詰められた byte は過去 row から復元できません。
 
-`command_audits.event_id` は `ON DELETE CASCADE` なので、`gc` で event を削除すると対応する audit payload も同時に消えます。
+`command_audits.event_id` は `ON DELETE CASCADE` なので、event row を削除すると対応する audit payload も同時に消えます。
 
 ### `sessions`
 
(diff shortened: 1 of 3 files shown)


duck8823:
Grok 4.6: ✅

Reviewed `1c572eb9` against #1797 option 3 (restate, do not widen discard).

- Docs state the allowlist is transcript-only, name prompt + `command_audits` text as permanently accruing, and call compact a reduction rather than a bounded store.
- Retired `gc` / `--target events` wording in the practical-implications paragraph is remapped onto `store compact`.
- `TestSelectDiscardableEventBodiesKindAllowlist` parses the shipped embedded SQL (`selectDiscardableEventBodiesQuery`) and asserts `cmp.Diff` against `{transcript}`. That drives the real allowlist, not a copy of the docs.
- SQL is not widened. `store gc` is not restored.

No CRITICAL/HIGH on this diff.

Leaves parent #1968 open.

duck8823:
Claude Opus 5: ✅

Reviewed PR #1983 against #1797 option 3.

SQL allowlist is not widened. `store gc` is not restored. The new test parses the shipped embedded SQL.

Leaves parent #1968 open.

Issue #1982, opened 2026-08-16
v0.40-24: store compact unconditionally DROPs event_content_dedupe_archive on the work copy


duck8823:
v0.40-24: store compact unconditionally DROPs event_content_dedupe_archive on the work copy

Parent: duck8823/traceary#1968

Found while designing #1702. After #1872 the only shipped apply path is compact copy-filter. That path writes quarantine rows onto the work copy under a run id, then `infrastructure/sqlite/compaction_copy_filter.go` DROPs `event_content_dedupe_archive` unconditionally before swapping the copy in.

Consequence: any live-store quarantine that existed from an older apply (or any rows written on the work copy that an operator might still need) disappear when compact succeeds. #1702 can put the run id on apply errors, but compact success still destroys the archive table, so the id is not a durable recovery handle across compact.

Out of #1702 on purpose (1 issue = 1 PR). Do not restore the deleted `store dedupe` restore/purge CLI in this issue unless that is the chosen surface after a design note.

## Done when

- compact either preserves a needed quarantine table, migrates it with a documented contract, or doctor/docs state that compact is the terminal discard of that archive
- A test pins the chosen contract (preserve vs explicit discard)
- Live store is not used as a gate

Leaves parent #1968 open.

PR #1984, opened 2026-08-16
fix: carry the dedupe apply run id on failure

### LINKED ISSUE #1702: v0.40-3: A failed dedupe apply drops the run id that would recover its committed batches
## Problem

Since #1682 an apply commits in batches, so a run that fails partway has already quarantined rows under its run id. But `application/usecase/store_management_usecase_impl.go` mints the run id, calls the datasource, and on error returns `ContentEventDedupeResult{}` — the id is dropped and never reaches the operator.

Those rows are gone from `events` and reachable only through `--list-runs`, which lists every run without distinguishing "this one completed" from "this one died halfway". The operator has to guess.

`--list-runs` was added in #1682 precisely so an interrupted run stays findable. It closes the *interruption* case (SIGKILL, OOM) but not the *error* case, which is the one the code can actually do something about.

## Direction

Carry the run id out with the failure. Wrapping the error with the id is enough to make it recoverable — the point is that the operator reading the failure knows what to hand to `--restore` or `--purge` without inspecting the archive.

Consider also marking runs that ended in error, so `--list-runs` can say which is which rather than making completeness something the operator infers from row counts.

## Related

Made more likely by the identification path aborting on a single undecodable body (see the sibling issue): that turns a corrupt row into a mid-apply failure rather than a skipped row.

## Acceptance

- A failing apply surfaces its run id in the error
- The operator can act on that id directly with `restore` / `purge`
- A test drives a mid-apply failure and asserts the id is present

duck8823:
## Summary

A failing content-event dedupe apply committed batches under a minted run id, then dropped that id when returning the error. This PR:

- Adds `apptypes.ContentEventDedupeApplyError{RunID, Err}` (`Error` / `Unwrap`)
- Wraps `Apply:true` usecase failures with that type (dry-run stays untyped)
- Mints a unique run id per `store compact` copy-filter apply instead of the fixed `compact-copy-filter` literal
- Pins mid-apply cancel: committed archive rows are addressable by the error's run id and restorable via `RestoreContentEventDedupeRun`

Does **not** restore the `#1872` `store dedupe content-events` restore/purge CLI. Compact recovery remains compact resume/rollback. Compact still DROPs the work-copy archive on success — that contract is #1982.

## Test plan

- [x] `go test ./application/types/ ./application/usecase/ ./infrastructure/sqlite/ -run 'Dedupe|ApplyError|ApplyFailure|compactCopyFilter|CompactCopy'`

Closes #1702

Leaves parent #1968 open.


duck8823:
Grok 4.6: ✅

Reviewed the three commits on `fix/1702-dedupe-apply-runid` against `.ai/spec/1702.md`.

- Apply failures are wrapped only in the minting usecase (`ContentEventDedupeApplyError`). Dry-run failures stay untyped. Mint failure is still "failed to mint".
- Datasource is not the wrap layer. The test hook `onAfterDedupeBatchCommit` only injects cancel between batches.
- Mid-apply test drives the shipped usecase + real SQLite store: `errors.As` recovers the run id, archive rows exist under that id, kept/unprocessed rows remain, `RestoreContentEventDedupeRun` brings them back.
- Compact copy-filter mints `compact-copy-filter-<hex>` per execution and wraps failures with the same type. It does not restore the deleted restore/purge CLI.

No CRITICAL/HIGH on this diff. Restore/purge CLI acceptance remains unmet by design (#1872); compact DROP of the archive is #1982.

Leaves parent #1968 open.

duck8823:
Claude Opus 5: ✅

Apply failures carry the minted run id via `errors.As`. Dry-run does not claim a run id. Deleted `store dedupe` / `store gc` CLI is not restored.

Leaves parent #1968 open.

PR #1985, opened 2026-08-16
fix: normalize payload_backfill_runs timestamp ordering

### LINKED ISSUE #1815: v0.40-14: normalize the remaining lexical timestamp ordering on payload_backfill_runs
Found while reviewing #1814 (#1729). Out of that PR's scope on purpose — #1729 enumerated four sites and `payload_backfill_runs` was not one of them.

## The hazard

`payload_backfill_runs.started_at` is written as RFC3339Nano:

```go
// infrastructure/sqlite/payload_backfill.go:467
now := time.Now().UTC().Format(time.RFC3339Nano)
```

RFC3339Nano is variable width, and `'.'` (0x2E) sorts below `'Z'` (0x5A). So a lexical comparison places `2026-01-01T00:00:00.5Z` **before** `2026-01-01T00:00:00Z` — the later instant sorts first ascending, last descending. This is the #1185 hazard.

Two reads order by that column lexically:

| site | function | consequence |
|---|---|---|
| `payload_backfill.go:930` | `loadActiveBackfillRun` | with more than one `running`/`paused` row, "the active run" can resolve to the older one |
| `payload_backfill.go:943` | `loadLatestBackfillRun` | "the latest run" can resolve to the older one |

Both are `LIMIT 1` over a `DESC` order, which is exactly the shape where one lexical inversion changes the answer rather than just the presentation order.

## How likely is it to bite

Lower than #1729's, because these tables normally hold one active run at a time — but not zero. Sub-second `started_at` values only collide with whole-second ones when two runs start in the same second, or when rows are seeded by tests and fixtures. The fix is cheap and removes the class rather than arguing about the odds.

## Fix

The same shape #1814 used:

```sql
ORDER BY ts_norm(started_at) DESC, run_id DESC
```

`ts_norm` is registered globally on the modernc driver in `infrastructure/sqlite/database.go`'s `init()`, so it is available on every connection with no extra wiring. The `run_id` tie-break makes the order total, matching `search_projection_session_query.go` and the sites #1814 fixed.

## Also worth checking in the same pass

`infrastructure/sqlite/content_event_dedupe_datasource.go:305` orders by `created_at DESC, id DESC`. Confirm whether that column is stored normalised (`created_at_norm` exists elsewhere, e.g. migration 031 for `events`) or raw before deciding whether it belongs here.

## Done when

- Both `payload_backfill.go` sites order through `ts_norm` with an id tie-break
- A red-before-green test covers a sub-second `started_at` sorting after a whole-second one, asserted with `cmp.Diff`
- The `content_event_dedupe_datasource.go` site is either fixed or explicitly recorded as not affected, with the reason

Related: #1185, #1729

duck8823:
## Summary

`loadActiveBackfillRun` and `loadLatestBackfillRun` ordered by raw RFC3339Nano `started_at`. A sub-second timestamp sorts lexically before a later whole-second instant. Both loaders now use `ORDER BY ts_norm(started_at) DESC, run_id DESC`.

`content_event_dedupe_datasource.go` `ORDER BY created_at` is left as-is: that path is a diagnostic sample; canonical resolution parses timestamps in Go.

## Test plan

- [x] `go test ./infrastructure/sqlite/ -run 'TestLoadActiveBackfillRun|TestLoadLatestBackfillRun'`
  - later instant wins across the lexical inversion
  - `run_id` tie-break

Closes #1815

Leaves parent #1968 open.

### DIFF
--- a/infrastructure/sqlite/payload_backfill.go
+++ b/infrastructure/sqlite/payload_backfill.go
@@ -927,7 +927,7 @@ func loadActiveBackfillRun(ctx context.Context, db *sql.DB) (backfillRunRow, err
 		       failure_event_id, failure_reason
 		  FROM payload_backfill_runs
 		 WHERE state IN ('running', 'paused')
-		 ORDER BY started_at DESC
+		 ORDER BY ts_norm(started_at) DESC, run_id DESC
 		 LIMIT 1`))
 }
 
@@ -940,7 +940,7 @@ func loadLatestBackfillRun(ctx context.Context, db *sql.DB) (backfillRunRow, err
 		       partial_metadata_rows, rewritten_rows, plaintext_bytes, stored_bytes,
 		       failure_event_id, failure_reason
 		  FROM payload_backfill_runs
-		 ORDER BY started_at DESC
+		 ORDER BY ts_norm(started_at) DESC, run_id DESC
 		 LIMIT 1`))
 }
 
--- a/infrastructure/sqlite/payload_backfill_test.go
+++ b/infrastructure/sqlite/payload_backfill_test.go
@@ -1819,6 +1819,101 @@ func insertPlaintextEventAtRowID(t *testing.T, db *sql.DB, rowID int64, seed eve
 	}
 }
 
+// insertBackfillRunRow seeds a payload_backfill_runs row directly for
+// loader-ordering tests that bypass the high-level API.
+func insertBackfillRunRow(t *testing.T, db *sql.DB, runID, state, startedAt string) {
+	t.Helper()
+	if _, err := db.Exec(`
+		INSERT INTO payload_backfill_runs(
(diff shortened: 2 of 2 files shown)


duck8823:
Grok 4.6: ✅

Reviewed `78a7894d` against #1815.

- Both `LIMIT 1` loaders now `ORDER BY ts_norm(started_at) DESC, run_id DESC`. `ts_norm` is already registered on every connection.
- Tests seed a lexical inversion (sub-second vs later whole-second) and assert the shipped loaders pick the later instant, plus a `run_id` tie-break.
- Dedupe `created_at` ORDER BY is recorded as a diagnostic sample only — out of this PR on purpose.

No CRITICAL/HIGH.

Leaves parent #1968 open.

duck8823:
Claude Opus 5: ✅

Both LIMIT 1 sites use `ts_norm`. Tests drive the shipped loaders.

Leaves parent #1968 open.

PR #1986, opened 2026-08-16
fix: batch command-audit payload hydration to O(1) queries

### LINKED ISSUE #1730: v0.40-7: batch command audit payload hydration

duck8823:
## Summary

`HydrateCommandAudits` issued one schema probe and one payload SELECT per event per field. This PR hoists the schema probe and fetches the requested page in one `json_each` SELECT, still going through codec + integrity + size limits.

## Test plan

- [x] `go test ./infrastructure/sqlite/ -run TestHydrateCommandAudits`
  - `TestHydrateCommandAuditsQueryCount` asserts exactly schema + payload queries for a 25-event page (codec and legacy shapes)
  - decoded values remain correct; fail-closed integrity is unchanged

Closes #1730

Leaves parent #1968 open.

### DIFF
--- a/infrastructure/sqlite/command_audit_payload.go
+++ b/infrastructure/sqlite/command_audit_payload.go
@@ -2,6 +2,8 @@ package sqlite
 
 import (
 	"context"
+	"database/sql"
+	"encoding/json"
 	"log/slog"
 
 	"golang.org/x/xerrors"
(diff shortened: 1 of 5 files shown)


duck8823:
Grok 4.6: ✅

Reviewed `1dd231f6` against #1730.

- Schema probe is hoisted. Payload fetch is one `json_each` SELECT per page.
- Codec + integrity + `maxDecodedPayloadBytes` stay on the decode path; no raw JOIN of plaintext.
- `TestHydrateCommandAuditsQueryCount` drives shipped `HydrateCommandAudits` and asserts O(1) query kinds (`schema`, `payload`) for a 25-event page on both codec and legacy shapes.

No CRITICAL/HIGH.

Leaves parent #1968 open.

duck8823:
Claude Opus 5: ✅

O(1) schema + batch SELECT. Codec/integrity/size limits remain. No `store gc` restore.

Leaves parent #1968 open.

PR #1987, opened 2026-08-16
fix: report unknown events.body_codec values in doctor

### LINKED ISSUE #1760: v0.40-9: body_codec accepts unknown values that only fail at read time

## Problem

`events.body_availability` has a closed CHECK constraint (migration 026) that names every legal value. `events.body_codec` (migration 036) has none. Any string writes successfully; the failure surfaces later, at read time, in `payload_codec.go:124-143`, when the decoder finds no handler for it.

The two columns describe the same row's body and are enforced to opposite standards.

## Failure scenario

A future writer — a migration, a repair path, an import — stores `body_codec = 'zstd:19'` or `'gzip'`. The write commits. Every subsequent read of that event fails, and it fails at the point of use (`read`, `replay`, `search`, an MCP call), not at the point of the mistake. The row is intact and unreadable, and nothing in the store distinguishes it from a corrupt body.

This is not currently reachable from any shipped code path; the concern is that nothing stops it from becoming reachable, and that the diagnosis cost is paid far from the cause.

## Constraint on the fix

Adding a CHECK to an existing column requires a 12-step table rebuild, which conflicts with the additive-migration / binary-rollback contract documented in `schema/sqlite/migrations/000055_add_search_projection_index_family_budget.sql:8-12`. `events` is the largest table in the store. A rebuild is not proportionate to a hazard that is currently unreachable.

## Suggested approaches, cheapest first

1. A `doctor` check that reports any `body_codec` value the running binary cannot decode. Detects the problem in the store, costs nothing at write time, and needs no migration.
2. An insert/update trigger that rejects unknown values. Additive, no rebuild — but the legal set becomes schema state that must be migrated whenever a codec is added.
3. The table rebuild. Only if the store is being rebuilt for another reason anyway.

Option 1 is probably right on its own.

## Scope

One diagnostic, or one trigger. Not a migration of `events`.

duck8823:
## Summary

`events.body_codec` has no CHECK constraint. An unsupported value commits and only fails later in `decodePayload`. This PR adds a read-only `body-codec` doctor check that lists codecs the running binary cannot decode, with per-codec counts and bounded sample IDs.

NULL `body_codec` is the pre-codec identity row and is not reported as unknown.

No schema change, no event rewrite, no trigger.

## Test plan

- [x] `go test ./infrastructure/sqlite/ ./presentation/cli/ -run 'BodyCodec|SupportedBodyCodecs|DoctorBodyCodec'`
  - real store: identity + zstd + NULL ignored; gzip / zstd:19 warned
  - shipped `doctor --json` on the same fixture
  - `SupportedBodyCodecs` matches `decodePayload`

Closes #1760

Leaves parent #1968 open.

### DIFF
--- a/application/body_codec_checker.go
+++ b/application/body_codec_checker.go
@@ -1,0 +1,24 @@
+package application
+
+import "context"
+
+// BodyCodecRow is a single unknown-codec observation from the events table.
+// Count is the number of rows carrying that codec value; SampleIDs holds up
+// to a bounded number of event IDs so an operator can locate the rows.
+type BodyCodecRow struct {
+	Codec     string
+	Count     int64
+	SampleIDs []string
+}
+
+// BodyCodecState holds the result of scanning events.body_codec for values
+// the current binary cannot decode.
+type BodyCodecState struct {
+	UnknownRows []BodyCodecRow
+}
+
+// BodyCodecChecker reports events rows whose body_codec value the running
+// binary's decoder does not support.
+type BodyCodecChecker interface {
+	CheckBodyCodec(context.Context) (BodyCodecState, error)
+}
--- a/infrastructure/sqlite/body_codec_checker.go
+++ b/infrastructure/sqlite/body_codec_checker.go
@@ -1,0 +1,94 @@
+package sqlite
+
+import (
+	"context"
+	"strings"
+
+	"golang.org/x/xerrors"
+
+	"github.com/duck8823/traceary/application"
+)
+
+const bodyCodecSampleLimit = 5
+
+// BodyCodecChecker reads events.body_codec and reports values the running
+// binary cannot decode.
+type BodyCodecChecker struct{ db *Database }
+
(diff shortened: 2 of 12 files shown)


duck8823:
Grok 4.6: ✅

Reviewed `3f9387f3` + `28ddf881` against #1760 option 1.

- The check is read-only. No events rebuild, CHECK, or trigger.
- `SupportedBodyCodecs()` is sourced from the same `identity` / `zstd` set as `decodePayload`.
- NULL `body_codec` is excluded (`IS NOT NULL`); that is the pre-codec identity row.
- Tests drive `CheckBodyCodec` on a real store and the shipped `doctor` command (`newTestRootCLI` + `WithBodyCodecChecker`) against the same unknown-codec fixture.

No CRITICAL/HIGH.

Leaves parent #1968 open.

duck8823:
Claude Opus 5: ✅

Read-only doctor. NULL `body_codec` is not unknown. Shipped `doctor` test drives a real store.

Leaves parent #1968 open.

PR #1989, opened 2026-08-16
fix: diagnose and once-retry rehearsal WAL autocheckpoint disable

### LINKED ISSUE #1780: v0.40-12: payload rehearsal aborts intermittently on "cannot disable rehearsal WAL autocheckpoint"
## Observation

`TestPayloadRehearsalResumesAfterArbitraryCommittedBatch` failed on CI and then passed on a re-run of **the same commit**, with no code change in between:

- `77ecf0e2` (branch `fix/1749-rehearsal-byte-sizing`), run [31382224983](https://github.com/duck8823/traceary/actions/runs/31382224983) — `Test` **failure**
- same run, `gh run rerun --failed` — **success**

Failure output:

```
--- FAIL: TestPayloadRehearsalResumesAfterArbitraryCommittedBatch (0.39s)
    payload_rehearsal_test.go:644: interrupted run error = advance payload rehearsal: cannot disable rehearsal WAL autocheckpoint
FAIL	github.com/duck8823/traceary/infrastructure/sqlite	123.774s
```

Not reproducible locally: 8 consecutive runs of `go test ./infrastructure/sqlite/ -run TestPayloadRehearsal` (including `-count=5`) all pass on an M-series Mac. CI is slower and more contended, which fits a timing-dependent failure.

## Where it comes from

`infrastructure/sqlite/payload_rehearsal_store.go:72-77`:

```go
if _, err = conn.ExecContext(ctx, `PRAGMA wal_autocheckpoint=0`); err != nil {
    return err
}
var autoCheckpoint int
if err = conn.QueryRowContext(ctx, `PRAGMA wal_autocheckpoint`).Scan(&autoCheckpoint); err != nil || autoCheckpoint != 0 {
    return errors.New("cannot disable rehearsal WAL autocheckpoint")
}
```

The guard is right to exist — the rehearsal's whole safety argument is that it never lets SQLite checkpoint behind its back, so a silent failure to disable autocheckpoint must abort. The problem is what it does on a transient failure and what it tells you afterwards.

Two defects, independent of the root cause:

1. **The error swallows both the `Scan` error and the observed value.** `err != nil || autoCheckpoint != 0` collapses two very different situations into one message. Whoever debugs this next cannot tell whether the pragma read failed or came back non-zero, nor what value it came back as. That is why this report cannot say which happened.
2. **No retry.** The pragma is set on a connection obtained from `s.db.Conn(ctx)` immediately above. If the failure is a transient driver or connection-state condition, one retry on a fresh connection would resolve it; today a whole rehearsal run aborts.

## Candidate root causes, none confirmed

- The connection is interrupted between the `Exec` and the read (`modernc.org/sqlite` installs an interrupt handler on a context with a `Done` channel), so the read returns a stale or default value.
- A DSN-level `_pragma` is re-applied on connection state reset, undoing the explicit `=0`.
- The read genuinely returns non-zero for a reason specific to the target's journal mode at that moment.

Determining which requires the failure to report the actual value, which is fix (1).

## Why it matters

`traceary store payload-rehearsal` is how an operator verifies a payload migration before running it, and it is on the v0.34 path. An intermittent abort in the release-critical preflight erodes trust in exactly the tool meant to build it. It also produces a red CI on a PR that changed nothing relevant, which is the precise failure mode that trains people to re-run instead of read.

## Acceptance

- The error distinguishes a failed pragma read from an unexpected value, and includes the value observed.
- A transient failure retries on a fresh connection before aborting the run; an actual non-zero value after the retry still aborts.
- Whatever the root cause turns out to be, it is recorded on this issue rather than left as "flaky".

duck8823:
## Summary

The rehearsal WAL autocheckpoint guard collapsed `err != nil || value != 0` into one opaque error. This PR:

- Extracts `disableRehearsalAutocheckpoint` with a typed `{stage: exec|read|value, observed, cause}` error
- Retries once on a fresh connection for exec/read failures
- Still aborts if the value is non-zero after the retry

The guard is not weakened. No `store gc` restore.

## Test plan

- [x] `go test ./infrastructure/sqlite/ -run 'Autocheckpoint|RehearsalWAL|disableRehearsal'`
  - transient read retries (fresh conn)
  - persistent non-zero still aborts with the observed value
  - exec failure is distinguishable from the value case

Closes #1780

Leaves parent #1968 open.

### DIFF
--- a/infrastructure/sqlite/payload_rehearsal_store.go
+++ b/infrastructure/sqlite/payload_rehearsal_store.go
@@ -18,6 +18,53 @@ import (
 
 type rehearsalMutationGuard func() error
 
+// rehearsalAutocheckpointStage identifies which step of the disable+readback
+// sequence produced a typed error so callers can distinguish exec from read
+// from a leftover non-zero value.
+type rehearsalAutocheckpointStage string
+
+const (
+	rehearsalAutocheckpointStageExec  rehearsalAutocheckpointStage = "exec"
+	rehearsalAutocheckpointStageRead  rehearsalAutocheckpointStage = "read"
+	rehearsalAutocheckpointStageValue rehearsalAutocheckpointStage = "value"
+)
+
+// rehearsalAutocheckpointError is returned by disableRehearsalAutocheckpoint.
+// Stage distinguishes exec/read failures (transient, retryable) from a
+// non-zero readback value (structural, not retried).
+type rehearsalAutocheckpointError struct {
+	Stage    rehearsalAutocheckpointStage
+	Observed int
+	Cause    error
+}
+
+func (e *rehearsalAutocheckpointError) Error() string {
+	if e.Stage == rehearsalAutocheckpointStageValue {
+		return fmt.Sprintf("rehearsal WAL autocheckpoint is %d after PRAGMA wal_autocheckpoint=0", e.Observed)
+	}
+	return fmt.Sprintf("cannot disable rehearsal WAL autocheckpoint (%s): %v", e.Stage, e.Cause)
+}
+
+func (e *rehearsalAutocheckpointError) Unwrap() error { return e.Cause }
+
+// disableRehearsalAutocheckpoint issues PRAGMA wal_autocheckpoint=0 then reads
+// it back. It returns a typed *rehearsalAutocheckpointError on any failure so
+// run can distinguish exec/read failures (retryable) from a non-zero value
+// (not retried).
+func disableRehearsalAutocheckpoint(ctx context.Context, conn *sql.Conn) (int, error) {
+	if _, err := conn.ExecContext(ctx, `PRAGMA wal_autocheckpoint=0`); err != nil {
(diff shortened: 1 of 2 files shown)


duck8823:
Grok 4.6: ✅

Reviewed `d104902e` against #1780.

- Guard stays fail-closed. Non-zero autocheckpoint after the single retry still aborts.
- Typed error distinguishes exec / read / value. One retry on a fresh connection for exec/read only.
- Tests inject `acquireConn` / `readAutocheckpoint` seams on the shipped `walBudgetedMutationSession` path.

No CRITICAL/HIGH.

Leaves parent #1968 open.

duck8823:
Claude Opus 5: ✅

Reviewed `facd78a6` against #1780.

- WAL autocheckpoint guard stays fail-closed: `Stage=value` bypasses retry and still aborts.
- Exactly one retry on exec/read: conn is closed, one fresh `acquireConn` + disable is attempted.
- Tests drive the real session on a WAL-mode fixture; conn count is 1 (happy) / 2 (retry).

Leaves parent #1968 open.

PR #1990, opened 2026-08-16
feat: classify retention-plan exclusions by first failing allowlist condition

### LINKED ISSUE #1762: v0.40-10: retention plan blames active_session for exclusions it does not cause
Found while implementing #1676 (PR #1757), raised by an independent review of that PR.

## Problem

`store retention plan` reports, per skipped row, why it was not selected. One of those reasons is `active_session`. That classification was written when "belongs to an active session" was the main thing keeping an old body out of the plan.

Since #1676, the plan is built from a positive allowlist: `transcript` kind, `available` body, an existing and **ended** session, older than the cutoff, and covered by a refinement of that same session. Ending the session satisfies exactly one of those five conditions.

`infrastructure/sqlite/raw_body_retention_datasource.go:94-101` still attributes the exclusion to the active session alone.

## Failure scenario

An active session holds an old `prompt`, an old `note`, or an uncovered `transcript`. The plan reports them as excluded because the session is active. The operator ends the session — deliberately, to free the space — re-runs `plan`, and the rows are still excluded, now for a reason nothing told them about. `prompt` and `note` bodies will never be candidates at all.

The output is not wrong about the session being active. It is wrong to present that as the reason the row was held back, because it implies an action that will not work.

## Suggested fix

Classify the exclusion by the first allowlist condition the row actually fails, so `not_transcript`, `uncovered`, `already_discarded`, `session_active` and `within_retention` are distinguishable. Reuse the shared predicate rather than re-deriving the conditions, so the reasons cannot drift from what the discard actually does.

## Scope

Reporting only. No change to which rows are selected.

duck8823:
## Summary

Retention plan no longer reports every skipped row as `active_session`. Exclusions are classified by the first allowlist condition the row actually fails (`not_transcript`, `already_discarded`, `session_active`, `within_retention`, `uncovered`), using the same discardable-event-bodies predicate as selection.

## Why

Ending a session does not free `prompt` / `note` / uncovered `transcript` bodies. Presenting `active_session` as the reason implied an action that does not work.

## Scope

Reporting only. Candidate selection is unchanged.

Closes #1762

Leaves parent #1968 open.

## Test plan

- [x] `go test ./infrastructure/sqlite/ ./application/usecase/ -run 'Retention|Exclusion|RawBody'`
- [x] `go tool golangci-lint run ./infrastructure/sqlite/... ./application/usecase/... ./application/types/... ./domain/types/...`


duck8823:
Grok 4.6: ✅

Reviewed `40b593bb` + `c623561a` against #1762.

- Exclusion reasons are the first failing allowlist condition (`not_transcript`, `already_discarded`, `session_active`, `within_retention`, `uncovered`), not a catch-all `active_session`.
- Classifier SQL composes the same discardable-event-bodies predicate as selection, so reasons cannot drift from what apply would do.
- `TestRawBodyRetention_classifiesExclusionReasons` drives `ListRawBodyCandidates` on a real fixture store for each reachable reason.
- Retention-plan schema and bilingual docs now include the new reason tokens so plan JSON stays in the published enum.

Leaves parent #1968 open.

duck8823:
Claude Opus 5: reviewed `c623561a`. Five HIGH findings. Addressed on `d2765614`:

- Exclusion listing is no longer every age-eligible non-candidate. Scope is available bodies in an active session plus available transcripts (the operator-relevant set). Historical prompts no longer explode plan JSON.
- Missing session rows classify as `session_missing`, not `session_active`.
- Exclusion / candidate reason sort uses schema `x-traceary-order`, not lexicographic order.

`already_discarded` stays in the enum and CASE but is not emitted per-row: discarded bodies were never in the pre-#1762 plan, and enumerating them is unbounded. `within_retention` remains the invalid-`created_at` branch inside the age-eligible window (the issue's token).

Leaves parent #1968 open.

PR #1991, opened 2026-08-16
fix: honour wall-time budget on payload-rehearsal rollback verification

### LINKED ISSUE #1858: v0.40-15: payload-rehearsal rollback verification has no wall-time budget

duck8823:
## Summary

`--wall-time-limit` now covers the read-only rollback verification phase (`fileDigest`, backup/copy verify, `PRAGMA integrity_check`). A budget miss fails before mutation; the prepared target and journal remain retryable. The atomic filesystem step after mutation intent stays un-interruptible.

## Why

On a large store, rollback verification ignored the operator budget and ran unbounded with no progress until JSON printed.

Closes #1858

Leaves parent #1968 open.

## Test plan

- [x] `go test ./infrastructure/sqlite/ ./application/usecase/ -run 'Budget|Digest|Rollback|Rehearsal|fileDigest|WallTime'`
- [x] `go tool golangci-lint run ./infrastructure/sqlite/... ./application/usecase/...`

### DIFF
--- a/application/usecase/payload_rehearsal.go
+++ b/application/usecase/payload_rehearsal.go
@@ -220,6 +220,11 @@ func (u *payloadRehearsalUsecase) Rollback(ctx context.Context, c types.PayloadR
 	if err := validateRehearsal(c); err != nil {
 		return types.PayloadRehearsalMetrics{}, err
 	}
+	if c.WallTimeLimit > 0 {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithDeadline(ctx, time.Now().Add(c.WallTimeLimit))
+		defer cancel()
+	}
 	result, err := u.runner.Rollback(ctx, c)
 	if err != nil {
 		return types.PayloadRehearsalMetrics{}, xerrors.Errorf("rollback payload rehearsal: %w", err)
--- a/infrastructure/sqlite/compaction_files.go
+++ b/infrastructure/sqlite/compaction_files.go
@@ -469,7 +469,7 @@ func (f PreparedStoreUpgradeFiles) Plan(ctx context.Context, run domain.Compacti
 		return run, insufficientCompactionSpaceError(required, available, id.Size)
 	}
 	run.SourceIdentity = id
-	digest, err := fileDigest(run.SourcePath)
+	digest, err := fileDigest(ctx, run.SourcePath)
 	if err != nil {
 		return run, errors.New("cannot digest prepared upgrade source")
 	}
@@ -524,7 +524,7 @@ func (f PreparedStoreUpgradeFiles) Recheck(ctx context.Context, run domain.Compa
 	if current != run.SourceIdentity {
 		return errors.New("source identity drift after plan")
 	}
-	digest, err := fileDigest(run.SourcePath)
+	digest, err := fileDigest(ctx, run.SourcePath)
 	if err != nil || digest != run.SourceDigest {
 		return errors.New("prepared upgrade source content changed")
 	}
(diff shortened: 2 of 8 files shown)


duck8823:
Grok 4.6: ✅

Reviewed `15bf5ce2` + `8478a5bd` against #1858.

- Usecase `Rollback` attaches `context.WithDeadline` from `--wall-time-limit`.
- `fileDigest` is context-aware and checks `ctx.Err()` per chunk.
- `TestRollbackBudgetExceededAbortsBeforeMutation` drives the real usecase Rollback on a fixture and asserts the target is unchanged.
- Post-mutation recovery stays detached from SIGINT via `WithoutCancel` / `detachedBudgeted` so the atomic step is not claimed interruptible.

Leaves parent #1968 open.

duck8823:
Claude Opus 5: two HIGH findings on `8478a5bd`. Addressed on `72981c34`:

- `fileDigest` context errors propagate as `context.Canceled` / `DeadlineExceeded` instead of tamper strings.
- Post-mutation rollback verification uses `context.WithoutCancel` so a deadline that expires during the atomic copy cannot fail a completed restore.

Leaves parent #1968 open.

PR #1995, opened 2026-08-16
fix: bind retention apply recheck to the persisted plan cutoff

### LINKED ISSUE #1763: v0.40-11: retention apply rechecks against the wall clock instead of the plan cutoff
Found while implementing #1676 (PR #1757), raised by an independent review of that PR.

## Problem

PR #1757 made `store retention` revalidate the whole discard predicate at apply time, inside the same transaction as the update, so a candidate that stopped being eligible between plan and apply is rejected. The cutoff bound into that recheck is the **apply timestamp**, not the cutoff the plan was built with (`infrastructure/sqlite/raw_body_retention_datasource.go:133`, `:306-307`).

That was a deliberate choice and it is safe in the direction that matters: the recheck only ever runs against candidates the plan already selected, and a later cutoff can only make more rows eligible, never fewer. It cannot admit a body the reviewed plan did not contain.

## Failure scenario

It is not sound for *resumability*, which is the reason to fix it rather than document it.

A plan is built with cutoff `2026-07-01` and selects an event from `2026-06-01`. The host clock is then corrected backwards to `2026-05-01` — NTP step, a VM restored from a snapshot, a container with a bad clock at start. The apply, or a retry of a partially-applied run, computes a cutoff of roughly `2026-02-01`, the unchanged candidate no longer passes, and the run stops as stale. Some earlier batches have already been applied. Resuming now depends on the wall clock rather than on the plan.

## Suggested fix

Persist the plan's cutoff with the plan and bind that at apply time. The recheck keeps its meaning — "is this still the row the plan reviewed" — and stops depending on when the apply happens to run.

Add a test that applies a plan under a clock set behind the plan's cutoff and still completes.

## Scope

One value threaded from the stored plan into the recheck, plus a test. No predicate change.

duck8823:
## Summary

The plan now persists `cutoff_at`. Apply-time eligibility recheck binds that value, not `time.Now()`. Older plans fall back to `created_at`. Selection SQL is unchanged.

## Why

A clock that moves backwards between plan and apply made remaining candidates look ineligible, so a partially-applied run could not resume.

Closes #1763

Leaves parent #1968 open.

## Test plan

- [x] `go test ./infrastructure/sqlite/ ./application/usecase/ -run 'RawBodyRetention|RetentionPlan|cutoff|Cutoff'`
- [x] `go tool golangci-lint run ./infrastructure/sqlite/... ./application/usecase/...`

### DIFF
--- a/application/raw_body_retention.go
+++ b/application/raw_body_retention.go
@@ -15,6 +15,8 @@ type RawBodyRetentionPlanner interface {
 
 // RawBodyRetentionExecutor applies and restores exact reviewed body identities.
 type RawBodyRetentionExecutor interface {
-	ApplyRawBodyPlan(ctx context.Context, databaseIdentity string, sqliteUserVersion int, migrationDigest, planID string, candidates []apptypes.RawBodyCandidate, appliedAt time.Time) (apptypes.RawBodyApplyResult, error)
+	// ApplyRawBodyPlan rechecks candidate eligibility against cutoffAt, the
+	// plan's persisted selection boundary, not the wall clock at apply time.
+	ApplyRawBodyPlan(ctx context.Context, databaseIdentity string, sqliteUserVersion int, migrationDigest, planID string, candidates []apptypes.RawBodyCandidate, cutoffAt, appliedAt time.Time) (apptypes.RawBodyApplyResult, error)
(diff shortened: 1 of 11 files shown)


duck8823:
Grok 4.6: ✅

Reviewed `873e8ded` (rebased onto #1762) against #1763.

- Plan persists `cutoff_at`; apply recheck binds that value, not wall-clock `appliedAt`.
- Older plans fall back to `created_at`. Selection SQL is unchanged.
- Tests drive apply under a clock set behind the plan cutoff.

Leaves parent #1968 open.

PR #1996, opened 2026-08-16
fix: preserve event_content_dedupe_archive across compact

### LINKED ISSUE #1982: v0.40-24: store compact unconditionally DROPs event_content_dedupe_archive on the work copy

duck8823:
## Summary

Copy-filter compact no longer `DROP`s `event_content_dedupe_archive`. The work copy already contains the table; compact trims rows older than a 90-day `archived_at` window. A store without the table is a no-op.

## Why

Compact was the only shipped apply path after #1872, and an unconditional DROP destroyed the quarantine that a run id is supposed to recover.

Closes #1982

Leaves parent #1968 open.

## Test plan

- [x] `go test ./infrastructure/sqlite/ -run 'DedupeArchive|CopyFilter|DeleteNonCanonical|Preserve'`

### DIFF
--- a/docs/storage/safe-compaction.ja.md
+++ b/docs/storage/safe-compaction.ja.md
@@ -44,6 +44,9 @@ compact は拒否しません。詳細は
 [`search-retirement.ja.md`](../operations/search-retirement.ja.md) を参照して
 ください。
 
+compact は `event_content_dedupe_archive`（content-event dedupe の隔離監査証跡）
+を 90 日の retention window 内で保持し、それより古い行は compact 時に破棄します。
+
 DarwinとLinuxでは通常のphysical
 SQLite connectionが、隣接するstableな `<database>.traceary.lock` にshared
 advisory lockを保持します。compact はjournalや向きの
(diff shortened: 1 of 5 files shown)


duck8823:
Grok 4.6: ✅

Reviewed `643ec6c5` against #1982.

- Both `DROP TABLE IF EXISTS event_content_dedupe_archive` sites are gone.
- Compact trims archive rows by `archived_at` (90-day window) instead of destroying the table.
- Tests on a TempDir fixture pin retain / over-retention / missing-table.

Leaves parent #1968 open.

PR #1998, opened 2026-08-16
fix: reuse one read-only handle for an orphan consolidation pass

### LINKED ISSUE #1722: v0.40-5: orphan consolidation opens one read-only connection per candidate

duck8823:
## Summary

`Database.WithReadScope` opens one read-only handle, runs `checkStoreCompatibility` once, and reuses the handle for `openReadOnly` while the scope is active. The compact orphan-consolidation pass enters that scope so N candidates cost O(1) connections, not O(N). Callers outside a scope keep the per-call path. No `*sql.DB` appears in `domain/` or `application/` signatures.

Closes #1722

Leaves parent #1968 open.

## Test plan

- [x] `go test ./infrastructure/sqlite/ -run 'WithReadScope|ReadScope|OrphanRangeDatasource_WithReadScope'`
- [x] `go test -run Leak .`

### DIFF
--- a/infrastructure/sqlite/database.go
+++ b/infrastructure/sqlite/database.go
@@ -185,6 +185,69 @@ type Database struct {
 	// resolved generation ids inside its snapshot transaction and before it
 	// reads generation-scoped counters. Tests commit a cutover here.
 	afterStatusGenerationScopeRead func()
+	// afterReadOnlyConnectionOpened runs each time openReadOnly or
+	// WithReadScope opens a genuinely fresh read-only connection (i.e. not a
+	// scope/shared-handle reuse). Tests use it to assert O(1) opens across a
+	// read-scoped pass; production leaves it nil.
+	afterReadOnlyConnectionOpened func()
+	// afterCompatibilityCheck runs each time checkStoreCompatibility succeeds
+	// inside openReadOnly or WithReadScope. Tests use it to assert the guard
+	// runs once per scope; production leaves it nil.
+	afterCompatibilityCheck func()
+}
+
+// readScopeKey is the unexported context key WithReadScope uses to carry its
+// shared read-only handle. Unexported so no package outside sqlite can read
+// or forge a scope value.
+type readScopeKey struct{}
+
+// readScope holds the read-only handle opened for the lifetime of a
+// WithReadScope call.
+type readScope struct {
+	db *sql.DB
+}
+
+// WithReadScope opens one read-only handle for the duration of fn, running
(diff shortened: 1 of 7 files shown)


duck8823:
Grok 4.6: ✅

Reviewed the #1722 commits against the spec.

- `WithReadScope` opens one read-only handle and runs compatibility once.
- Nested scopes reuse the outer handle. Unscoped `openReadOnly` stays per-call.
- Compact's orphan pass enters the scope. Tests count opens on a real fixture.
- Root leak test keeps `*sql.DB` out of domain/application signatures.

Leaves parent #1968 open.

PR #1999, opened 2026-08-16
fix: store encoded payload bytes in the dedupe archive

### LINKED ISSUE #1744: v0.40-8: dedupe quarantine archive stores decoded plaintext, inflating the store and mis-reporting released bytes after the payload codec backfill
Found while sweeping body-interpreting SQL for #1685 (decision D6). Out of scope there; filing rather than widening that PR.

## What

`event_content_dedupe_archive.body` stores the **decoded plaintext**, not the stored representation.

`decodeDedupeCandidateBody` (`infrastructure/sqlite/content_event_dedupe_datasource.go:379-401`) mirrors `loadEventPlaintext`: it decodes the payload through the codec and returns plaintext. That plaintext is what `applyDedupe` writes into the archive (`content_event_dedupe_datasource.go:790-800`), and the source row is then deleted from `events`.

Before #1685 this was invisible, because every `events.body` was already plaintext. After the payload codec backfill it has two consequences.

## Consequence 1 — quarantine inflates the store

Quarantining a compressed row writes it back out uncompressed. A dedupe run over a compressed corpus therefore *grows* the file by roughly the compression ratio for every duplicate it quarantines, until the archive is purged. That works directly against the #1620 store-size gate, and it is the opposite of what a cleanup command should do.

The archive table has no codec columns at all (`schema/sqlite/migrations/000019_create_event_content_dedupe_archive.sql:22` is a bare `body TEXT NOT NULL`), so there is currently nowhere to record that a stored value is encoded.

## Consequence 2 — `ReleasedBody` becomes a logical figure

```
SELECT COUNT(*), SUM(length(CAST(body AS BLOB)))
  FROM event_content_dedupe_archive WHERE dedupe_run_id = ?
```

(`content_event_dedupe_datasource.go:885-890`, reported as `ReleasedBody` at :908, and the same sum per run at :931-934.)

Since the archived body is plaintext, that sum is the *logical* size of the quarantined content, while the bytes actually freed from `events` were the *compressed* ones. Operators reading "released bytes" after a purge would get a number several times larger than the space the purge can return.

## Options

1. Give the archive the same five codec columns and store the payload as-is, decoding only on restore. Keeps the archive compressed and lets the byte sum say which figure it means.
2. Keep plaintext in the archive but record the source's stored size alongside it, and report that as released bytes.

(1) is the one that also fixes consequence 1.

## Not in scope

`#1685` leaves this untouched. The bodies it rewrites are in `events`; nothing in the backfill reads or writes `event_content_dedupe_archive`.

Depends on: #1685 (the backfill is what makes this observable).

duck8823:
## Summary

`event_content_dedupe_archive` gets the same codec columns as `events`. Apply copies the encoded payload as-is. `ReleasedBody` sums stored size. Restore decodes only to validate integrity and writes the original encoded bytes back. Legacy plaintext archive rows stay identity-codec.

Closes #1744

Leaves parent #1968 open.

## Test plan

- [x] `go test ./infrastructure/sqlite/ -run 'Dedupe|Archive|Codec|ReleasedBody|content_event_dedupe'`

### DIFF
--- a/infrastructure/sqlite/content_event_dedupe_codec_internal_test.go
+++ b/infrastructure/sqlite/content_event_dedupe_codec_internal_test.go
@@ -1,0 +1,234 @@
+package sqlite
+
+import (
+	"context"
+	"database/sql"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"testing"
+	"testing/fstest"
+	"time"
+
+	apptypes "github.com/duck8823/traceary/application/types"
+)
+
+// dedupeCodecMigrations returns the on-disk embedded migration set, matching
+// the pattern content_event_dedupe_unreadable_internal_test.go already uses
+// for internal (package sqlite) tests that need an on-disk store.
+func dedupeCodecMigrations() fs.FS {
+	return os.DirFS(filepath.Join("..", "..", "schema", "sqlite", "migrations"))
+}
+
+// migrationsBeforeDedupeArchiveCodec returns every embedded migration older
+// than #1744's, so a test can build a genuine pre-migration archive row
+// before reopening with the full set -- mirrors
+// migrationsBeforeCodecFoundation in payload_codec_compatibility_internal_test.go.
+func migrationsBeforeDedupeArchiveCodec(t *testing.T) fstest.MapFS {
+	t.Helper()
+	dir := filepath.Join("..", "..", "schema", "sqlite", "migrations")
+	entries, err := os.ReadDir(dir)
+	if err != nil {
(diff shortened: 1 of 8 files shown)


duck8823:
Grok 4.6: ✅

Reviewed `29616131` against #1744.

- Migration 67 adds codec columns; apply copies encoded bytes.
- ReleasedBody uses stored size, not decoded plaintext length.
- Tests drive apply on a compressed fixture and assert archive body equals stored bytes.

Leaves parent #1968 open.

PR #2000, opened 2026-08-16
fix: fold oldest orphan ranges first so compact can delete during a long drain

### LINKED ISSUE #1721: v0.40-4: let gc delete during a long consolidation drain by folding oldest ranges first

duck8823:
## Summary

Bounded orphan consolidation used to gate compact `--force` deletion on a 100% drained backlog (`!HasMore && Skipped == 0`). On a store with a large but recent unfolded backlog that blocked `CollectGarbage` even though deletion only ever touches events older than the retention cutoff.

This remaps the #1721 gate onto compact's work-copy cover (no `store gc` restoration):

- Discovery paginates ended/stale sessions oldest-first via a composite `(earliest_event_norm, session_id)` keyset cursor.
- `OrphanConsolidationResult` reports `EarliestUnprocessedEventTime` / `EarliestSkippedEventTime` (`nil` = unknown = refuse).
- `ForceCoverSafeToDelete` replaces `ForceCoverMustComplete`: deletion proceeds when leftover ranges are entirely newer than the cutoff, even with `HasMore`.
- `--force` cover is a bounded pass, not `Unlimited`.
- #1722's `WithReadScope` wrapping is kept.

Leaves parent #1968 open.

## Test plan

- [x] `go test ./domain/model/ ./application/ ./application/types/ ./application/usecase/ ./infrastructure/sqlite/`
- [x] `go build .`
- [x] `go tool golangci-lint run` on touched packages
- [x] Fixture: leftover newer than cutoff + `HasMore` → compact proceeds (`TestCompactForceCoverProceedsWhenLeftoverIsNewerThanCutoff`)
- [x] Fixture: leftover older than cutoff → compact refuses (`TestCompactForceCoverRefusesWhenLeftoverIsOlderThanCutoff`)

Closes #1721

### DIFF
--- a/application/compact.go
+++ b/application/compact.go
@@ -38,8 +38,12 @@ type CommandBodyReclaim struct {
 // CompactFilter configures the copy-filter inside Build.
 // A zero value is vacuum-only: no body discard, no AfterClone.
 type CompactFilter struct {
-	Cutoff     time.Time
-	AfterClone func(context.Context, string) error
+	Cutoff time.Time
+	// AfterClone runs the --force mechanical cover on the work copy before the
+	// discard/vacuum steps. It receives Cutoff so the cover can decide it has
+	// folded enough of the oldest backlog to let CollectGarbage proceed even
+	// when unrelated, newer-than-cutoff material is still unfolded (#1721).
+	AfterClone func(ctx context.Context, work string, cutoff time.Time) error
 }
 
 // BodyGate classifies discardable-age transcript bodies on the source.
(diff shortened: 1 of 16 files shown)


duck8823:
## Grok 4.6 review

Reviewed the two implementer commits plus the follow-up on `7e3c6d76`.

The original discovery loop filled the bounded page from recorded compact markers first (`session_id` order, not age) and skipped the ended/stale scan once that page was full. `ForceCoverSafeToDelete` would then see only newer leftovers and allow `CollectGarbage` while older ended ranges were still unfolded. That is the #1719/#1721 hazard.

`7e3c6d76` collects every still-active recorded shortcut and the oldest ended/stale page independently, then merge-sorts. Regression: `TestSessionOrphanRangeDatasource_RecordedShortcutsDoNotHideOlderEndedSessions`.

`lastAttempted` as the unprocessed bound is conservative (oldest-first remaining is newer). Fixture tests drive compact, not only the helper.

Grok 4.6: ✅ after `7e3c6d76`.

duck8823:
## Opus review (independent)

Opus REQUEST_CHANGES on the pre-follow-up discovery loop:

1. CRITICAL — recorded shortcuts filled `target` and aborted the ended/stale oldest-first scan, so `earliestUnprocessed` could be newer than unseen pre-cutoff ranges.
2. HIGH — `lastAttempted` as the unprocessed lower bound is fail-open unless discovery is a global oldest-first prefix.

Addressed:
- `7e3c6d76` collects every still-active recorded shortcut and the oldest ended/stale page independently, then merge-sorts. That prefix is what makes `lastAttempted` a safe lower bound (remaining candidates are newer).
- `d9d89dbf` fixes the regression fixture: recorded sessions now have last activity inside the 24h window so they stay on the recorded path (Opus correctly noted the first fixture was stale and never hit the hole).

Opus: ✅ after `d9d89dbf` (the production hole was already closed in `7e3c6d76`; the test now actually covers it).

PR #2001, opened 2026-08-16
fix: fold pre-cutoff tails of never-ending active sessions

### LINKED ISSUE #1724: v0.40-6: long-lived active sessions are never consolidated but their old events are deleted

duck8823:
## Summary

Orphan discovery only saw recorded compact markers and ended-or-stale sessions. A session that stays continuously active (activity at least every 24h, never `ended_at`) never became a candidate, so its pre-cutoff tail was never folded.

This adds a third discovery source, opt-in via `OrphanConsolidationInput.RetentionCutoff` (zero value keeps the previous two-source behaviour):

- Sessions with events past `covers_to` older than the compact retention cutoff, regardless of lifecycle.
- Fold terminus is the last-observed event strictly before the cutoff, not "now" or the session's latest event.
- Marker source defers to this live re-derivation when both match.
- `compactWorkCover` threads the compact cutoff through so `--force` cover closes the gap in production.

Coverage-aware DELETE is still out of scope. Transcript body discard already refuses `ended_at IS NULL` sessions; this PR is the incremental fold (so the backlog does not surface all at once when the session finally ends or goes stale).

Leaves parent #1968 open.

## Test plan

- [x] `go test ./domain/model/ ./application/ ./application/usecase/ ./infrastructure/sqlite/`
- [x] `go build .`
- [x] `go tool golangci-lint run` on touched packages
- [x] Discovery fixtures: all-newer / straddling / all-older / tail past marker
- [x] Compact `--force` pin: never-ending session folds at last pre-cutoff event (`TestCompactForceCoverFoldsNeverEndingSessionsPreCutoffTail`)

Closes #1724

### DIFF
--- a/application/types/orphan_consolidation_result.go
+++ b/application/types/orphan_consolidation_result.go
@@ -120,11 +120,15 @@ func (r OrphanConsolidationResult) Failures() OrphanConsolidationFailures { retu
 // can actually act on.
 //
 // This is a statement about this pass, not a proof that nothing unfolded
-// remains anywhere. Discovery sees ended or stale sessions plus recorded
-// markers, so a session that has stayed continuously active for longer than
-// the retention window still holds old unfolded events. Those keep their
-// bodies rather than losing them, but they also never become discardable; that
-// gap predates bounding and is tracked in #1724.
+// remains anywhere. Discovery sees ended or stale sessions, recorded markers,
+// and (when the caller supplies a retention cutoff) sessions that never go
+// ended or stale but still hold material past coverage older than that
+// cutoff. The last of these closes what used to be a standing gap: a session
+// that stayed continuously active for longer than the retention window used
+// to be invisible to every discovery source, so its old material was never
+// folded no matter how old it got. It is now discovered and folded up to,
+// but never past, the supplied cutoff — the still-recent tail is left
+// uncovered and surfaces again on a later pass once it too ages past cutoff.
 //
 // There is deliberately no Complete() beside this. It used to mean "no more
 // candidates and nothing was skipped", and conflating those two facts is the
--- a/application/usecase/orphan_consolidation_usecase.go
+++ b/application/usecase/orphan_consolidation_usecase.go
@@ -24,6 +24,15 @@ type OrphanConsolidationInput struct {
 	// Unlimited ignores Limit and Budget so compact --force can cover every
 	// discardable-age unrefined session in one rewrite.
 	Unlimited bool
+	// RetentionCutoff enables the third orphan-discovery source (#1724):
+	// sessions that never go ended/stale but still hold material past
+	// refinement coverage older than this cutoff. The zero value disables
+	// that source, keeping discovery to the recorded-marker and
+	// ended/stale sources only. Callers that know the compact retention
+	// cutoff (e.g. the --force cover pass) should pass it here so an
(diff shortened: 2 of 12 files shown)


duck8823:
## Grok 4.6 review

Third source is opt-in via `RetentionCutoff` and `compactWorkCover` actually passes the compact cutoff. Fold terminus is the last pre-cutoff event (`gapPastCoverageBeforeCutoff`). Recorded markers for the same session are skipped when source 3 also matches (and the fixture asserts one candidate whose `ToEventID` is the live tail, not the frozen marker).

`select_discardable_event_bodies.sql` already refuses `ended_at IS NULL`, so compact does not physically discard an active session's transcript bodies today. The real #1724 value is incremental folding so the backlog does not arrive all at once when the session later ends or goes stale. That matches the issue's "done when" for discovery/terminus.

Grok 4.6: ✅

duck8823:
## Independent reviewer

Reviewed the five hunt items on the current branch. Terminus is last pre-cutoff event, `compactWorkCover` passes `RetentionCutoff`, marker/source-3 dedupe is explicit, discovery merge-sort does not starve source 3, and `DiscoverCandidates` arity matches at every call site.

Independent reviewer: ✅ 問題なし

Issue #2025, opened 2026-08-16
v0.41.0: large-store read reliability and capture hygiene


duck8823:
Dogfood pass on 2026-08-16 after upgrading to v0.40.0 (brew CLI + all six host plugins; doctor reports every plugin-version check pass at 0.40.0). Findings were measured against the live 36.2 GiB store with official brew 0.40.0, or against main at 8c291152. Repo-built binaries were never pointed at the live store.

Constraint: no external interface changes as the default. Prefer existing surface (doctor --fix, existing flags). A new flag/command is an explicit in-sprint design decision on that child, not a reason to move the child out of v0.41.

Do not shrink this set. Every child listed below ships in v0.41.0. Do not shelf, split to v0.42, or leave a child standalone because it is hard.

Closing policy:
- Implementation PRs close children only (Closes #<child>).
- This parent stays open through the children and the release-preparation PR.
- Close this parent only after the tagged v0.41.0 release publishes.
- Never put Closes #2025 on an implementation PR or the release-prep PR.

Wave 1 — hot-path correctness:
- [ ] #2010 (v0.41-1) search projection cleanup never completes on large stores
- [ ] #2011 (v0.41-2) session lineage SQL materializes every event body
- [ ] #2012 (v0.41-3) parent SessionEnd leaves child subagent sessions open
- [ ] #2013 (v0.41-4) stop_hook-family usage observations record epoch-zero timestamps

Wave 2 — large-store reads:
- [ ] #2014 (v0.41-5) bound timeline and replay walks
- [ ] #2015 (v0.41-6) bound/cache dbstat inspections
- [ ] #2016 (v0.41-7) rate-limit the per-command store-compact reminder
- [ ] #2006 (v0.41-18) metadata_only_large_store doctor hides host hook/config diagnostics
- [ ] #2008 (v0.41-20) 36 GiB store cannot be compacted — full compact-path work stays in v0.41 (external-volume runbook, reclaimable-bytes estimate, and incremental/in-place compact if a replica still cannot fit). Not a deferral ticket.

Wave 3 — hygiene / observability:
- [ ] #2017 (v0.41-8) GC per-PID hook state files, SessionEnd diagnostics, ended markers
- [ ] #2018 (v0.41-9) report should disclose excluded usage observations
- [ ] #2019 (v0.41-10) sparse additive usage coverage (claude and grok both stay in this child)
- [ ] #2020 (v0.41-11) memory extraction turns code/diff/doc fragments into candidates
[...]

PR #2026, opened 2026-08-16
fix: rank session latest events without scanning every body

### LINKED ISSUE #2011: v0.41-2: session lineage query materializes every event body and bypasses the created_at_norm index

## Symptom

`traceary sessions --snapshot` ran **31+ minutes** (132 s user / 258 s system CPU, state mostly uninterruptible disk wait) before being killed. Error chain on cancel: `セッション lineage の取得に失敗しました: failed to get session lineage: failed to query session lineage: context canceled`. The root help advertises `sessions --snapshot` as a primary read entry; on the live store it is effectively unusable.

## Mechanism

`infrastructure/sqlite/sql/session_lineage.sql`, `latest_events` CTE:

```sql
ROW_NUMBER() OVER (
  PARTITION BY e.session_id
  ORDER BY ts_norm(e.created_at) DESC, e.id DESC
) AS rn
```

- The window orders by the **expression** `ts_norm(e.created_at)`, so the existing index `idx_events_session_created_at_norm_id_desc ON events(session_id, created_at_norm DESC, id DESC)` cannot serve it — the `created_at_norm` column exists precisely for this and is bypassed.
- The inner select projects `e.body` (via `COALESCE(NULLIF(e.body,''), ...)`) for **every** event in the lineage before discarding all but `rn = 1`, materializing gigabytes of bodies on a large store.
- `sessions --snapshot` runs the lineage query across the active-session set (~60 never-ended sessions at measure time), multiplying the scan.

`session handoff` shares the query but happened to be fast (4 s) because its lineage held 31 events; cost is proportional to lineage event volume.

- Order the window (or a correlated latest-1 lookup) by the `created_at_norm` column so the per-session index applies.
- Fetch `body` / `command_audits` join only for the selected `rn=1` row, not for every candidate.
- Add a regression test with a large-event-count session asserting the query plan uses the index (or a bounded-time assertion in the harness used for #1746-style pins).

Constraint: no external interface changes — output shape of `sessions --snapshot` / `session handoff` stays identical.

duck8823:
## Summary
- Rank lineage/tree latest events by `created_at_norm` and join only that row's body.
- Query-plan tests require the session/created_at_norm index and forbid `ts_norm(created_at)` ranking.

Closes #2011
Leaves parent #2025 open.

## Test plan
- [x] `go test ./infrastructure/sqlite/ -run 'TestDatasource_Lineage|TestDatasource_List|TestDatasource_Tree|TestSessionLineageLatestEvent|TestSessionTreeLatestEvent'`

### DIFF
--- a/infrastructure/sqlite/list_sessions_test.go
+++ b/infrastructure/sqlite/list_sessions_test.go
@@ -998,6 +998,34 @@ func TestDatasource_LineageAndTreeLatestEventsArePerSession(t *testing.T) {
 	assertSummaryLatest(t, tree, "latest-child", types.EventKindReviewed, "child latest")
 }
 
+func TestDatasource_LineageOfLatestEventAmongManyEventsInOneSession(t *testing.T) {
+	t.Parallel()
+
+	fixture := newListSessionsFixture(t, filepath.Join(t.TempDir(), "traceary.db"), listSessionsTestMigrations())
+	ctx := context.Background()
+	if err := fixture.storeManager.Initialize(ctx); err != nil {
+		t.Fatalf("Initialize() error = %v", err)
+	}
+	started := time.Date(2026, 4, 10, 12, 0, 0, 0, time.UTC)
+	saveTestSession(ctx, t, fixture.sessionDS, "busy-session", started, types.None[time.Time](), "codex", "duck8823/traceary")
+
+	const eventCount = 500
+	for i := 0; i < eventCount; i++ {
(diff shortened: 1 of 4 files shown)


duck8823:
Grok 4.6 medium: review of 42710261

- latest-event ranking uses created_at_norm + idx_events_session_created_at_norm_id_desc and defers body/audit join until rn=1. Matches #2011.
- Query-plan tests fail if ts_norm(created_at) remains in the ranking path. They correctly do not treat the final session ORDER BY temp b-tree as a regression.
- event_agg still scans lineage events for counts/agents; that is out of #2011 scope.

No CRITICAL/HIGH findings. Approve for CI + merge after green.

Leaves parent #2025 open.

PR #2031, opened 2026-08-16
fix: stamp stop_hook usage observations with wall time

### LINKED ISSUE #2013: v0.41-4: stop_hook-family usage observations record epoch-zero timestamps
Dogfood 2026-08-16, live store queried read-only.

## Symptom

`usage_observations` holds **860 rows with `observed_at = '1970-01-01T00:00:00Z'`** (and matching epoch-zero `finalized_at`): grok 412, antigravity 343, kimi 105. All are `stop_hook` / `session_end_hook` sources. Example row:

```
observation_id: grok:stop_hook:e532b0ff...   host: grok   source_name: stop_hook
observed_at: 1970-01-01T00:00:00Z   finalized_at: 1970-01-01T00:00:00Z
accounting: excluded   input_state: unavailable
```

Epoch-zero timestamps make these rows invisible to every period-windowed consumer (`report`, retention windows, decay-style scans) and misrepresent when the observation actually happened. The rows themselves are intended (stop_hook availability observations legitimately carry `unavailable` token states), but their timestamps are not.

- Fix the stop_hook-family write path to stamp `observed_at` / `finalized_at` with the hook wall clock (the surrounding events on the same session carry correct times, so the bug is in the observation writer, not the host payload).
- One-shot repair for the existing 860 rows: backfill from the owning session's nearest event/boundary time, or leave a documented decision that historical rows stay epoch-zero (either way, decide explicitly in the PR).
- Regression test per affected host source.

Constraint: no external interface changes — table schema and JSON contracts stay as-is; this is a value bug.

duck8823:
## Summary
- Stop-hook-family unavailable writers stamp wall clock (or payload event time), never Unix epoch 0.
- Store Initialize backfills existing epoch-zero hook rows from the session's nearest event or boundary.

Closes #2013
Leaves parent #2025 open.

## Test plan
- [x] `go test ./application/usecase/ ./infrastructure/sqlite/ -run 'TestGrokUsageCapture_|TestClaudeUsageCapture_|TestRepairEpochZero|TestInitialize_RepairsEpochZero|TestUsageCaptureUsecases_WriteNonEpoch'`

### DIFF
--- a/application/usecase/antigravity_usage_capture.go
+++ b/application/usecase/antigravity_usage_capture.go
@@ -180,7 +180,10 @@ func (u *antigravityUsageCaptureUsecase) CaptureStopUnavailable(
 	if err != nil {
 		return AntigravityUsageCaptureResult{}, xerrors.Errorf("invalid Antigravity Stop usage source: %w", err)
 	}
-	observedAt := time.Unix(0, 0).UTC()
+	observedAt, err := resolveUnavailableObservationTime(ctx, u.repository, id, time.Time{})
+	if err != nil {
+		return AntigravityUsageCaptureResult{}, err
+	}
 	descriptor, err := model.NewUsageObservationDescriptor(
 		id, input.SessionID, source, types.UsageScopeCall, types.UsageAccountingExcluded, observedAt,
 	)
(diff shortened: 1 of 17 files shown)


duck8823:
Grok 4.6 medium: review of 8c568d3c

- All six stop-hook-family unavailable writers use resolveUnavailableObservationTime (preferred event time, else now; replay reuses the first stamp).
- Initialize runs a bounded repair: drop the descriptor trigger in-tx, rewrite epoch-zero hook rows from session event/boundary time, restore the trigger.
- Tests seed real fixture rows and drive RepairEpochZeroHookUsageObservations plus Initialize.

No CRITICAL/HIGH findings. Approve after CI.

Leaves parent #2025 open.

PR #2032, opened 2026-08-16
fix: rate-limit the per-command store-compact reminder

### LINKED ISSUE #2016: v0.41-7: rate-limit the per-command store-compact reminder trailer

## Symptom

Every single command — `list`, `search`, `session gc --dry-run`, `memory decay`, `store search-projection status`, `replay`, … — appends:

```
TRACEARY: ストアに回収できる領域があります。`traceary store compact` を実行してください
```

On this machine the advice is currently **infeasible**: compact is a copy-filter needing ~36 GiB free while the volume has ~30 GiB (#2008). The operator sees the same nag dozens of times per hour with no way to act on it, which trains them to ignore Traceary trailers entirely — including ones that matter.

- Rate-limit the reminder per store via a small state file next to the store (e.g. once per 24 h per store path), unconditionally — this keeps the hint while removing the spam.
- Optionally (cheap, metadata-only): compare `statvfs` free bytes against the DB size before printing, and when a replica clearly cannot fit, say so once instead of recommending the impossible command (feeds the #2008 discussion without waiting for it).
- Keep `doctor` output unchanged — doctor is the place where the full remediation story lives.

Constraint: no external interface changes — stderr/stdout trailer text frequency only; no flags, no JSON changes.

duck8823:
Closes #2016

Rate-limit the compact reclaim trailer to once per 24 hours per store. When the volume cannot hold a replica, the trailer says so instead of recommending an impossible command. Doctor output is unchanged.

Leaves parent #2025 open.

### DIFF
--- a/presentation/cli/compact_reclaim_space_unix.go
+++ b/presentation/cli/compact_reclaim_space_unix.go
@@ -1,0 +1,17 @@
+//go:build unix
+
+package cli
+
+import (
+	"fmt"
+
+	"golang.org/x/sys/unix"
+)
+
+func volumeAvailableBytes(path string) (uint64, error) {
+	var stat unix.Statfs_t
+	if err := unix.Statfs(path, &stat); err != nil {
+		return 0, fmt.Errorf("statfs %s: %w", path, err)
+	}
+	return uint64(stat.Bavail) * uint64(stat.Bsize), nil
+}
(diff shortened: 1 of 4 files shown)


duck8823:
Grok 4.6 medium: ✅ rate-limit + infeasible-replica wording. No CRITICAL/HIGH on the diff. Sidecar next to the store is the right isolation; doctor still skipped.

duck8823:
Claude review: summary=問題なし. Local go test CompactReclaim passed. Merged on Grok + Claude receipts and green CI.

PR #2033, opened 2026-08-16
fix: bound timeline and replay walks on created_at_norm

### LINKED ISSUE #2014: v0.41-5: bound timeline and replay walks on large stores

## Measurements

- `traceary timeline` — **590 s** for the default window (output itself is small: ~14 workspace rows over 64 h).
- `traceary replay --out …` — **980 s** to produce a 122 KB HTML (sessions=10, events=127, memories=1, timeline=20, hotspots=10). Replay embeds the timeline-block and failure-hotspot walks, so it inherits (and exceeds) the timeline cost.

Both are read commands an operator reaches for casually ("what happened this week", "export a retrospective page"); 10–16 minutes on a real store makes them practically unusable, and neither prints a partial/coverage hint the way `report` does (`coverage=partial`, `--result-cap`).

- Route the timeline gap-walk and replay's timeline/hotspot collection through the indexed `created_at_norm` access paths and bound the scanned row volume the way `report` bounds aggregation (internal caps — no new flags needed; existing defaults like `--timeline-blocks` / `--hotspots` already define how much output is wanted, so the walk should read backward only until those quotas are filled).
- If a bound is hit, keep the existing honest-partial pattern in the rendered output (replay HTML/markdown already has freeform areas; timeline can reuse the report-style coverage note in its existing text output).

Constraint: no external interface changes — no new flags, same output shapes; only how much the walk reads changes, disclosed with the established partial-coverage wording.

duck8823:
Closes #2014

Timeline gap-walk no longer selects event bodies. It pages newest-first on created_at_norm with an internal scan cap and prints coverage=partial when the existing block limit is hit. Replay keeps using the same ListTimelineBlocks path.

Leaves parent #2025 open.

### DIFF
--- a/application/types/timeline_block.go
+++ b/application/types/timeline_block.go
@@ -86,6 +86,7 @@ type TimelineBlock struct {
 	eventCount         int
 	agents             []string
 	workspaceBreakdown []TimelineWorkspaceBreakdown
+	scanTruncated      bool
 }
 
 // TimelineBlockOf creates a TimelineBlock.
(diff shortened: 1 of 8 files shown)


duck8823:
Grok 4.6 medium: ✅ newest-first created_at_norm walk, no e.body in the gap CTE, query-plan test names the index. coverage=partial is text-only so JSON shape stays. No CRITICAL/HIGH.

duck8823:
Claude opus-class review: HIGH — scan cap can truncate newest_events while len(blocks) < limit, so coverage=partial never prints and the oldest returned block's event_count/start can be wrong. Follow-up commit will surface scan-cap truncation in text output without changing JSON.

PR #2035, opened 2026-08-16
fix: emit host hook and config checks on large-store doctor

### LINKED ISSUE #2006: v0.41-18: metadata_only_large_store doctor hides host hook and config diagnostics
On the live 36.2 GiB dogfood store, official 0.40.0 doctor stays in metadata_only_large_store and only warns store-size + large-store-diagnostics. The same binary against a tiny scratch DB immediately warns gemini-config (5s vs 10s timeouts), codex-config (plugin hooks plus leftover ~/.codex/hooks.json duplicates), hook-grok-transcript, hook-memory-extract, claude-hook-cancellations (5119 markers), and claude-plugin-cache (0.39.0 snapshots). #1970 kept plugin-version visible; that family now passes after the plugin refresh. Host-file checks that do not open SQLite are still gated by store size, so an operator who only doctors the live store never sees FixCommand. gemini-config and codex-config were remediable with doctor --fix against the scratch DB.

duck8823:
Closes #2006

metadata_only_large_store still does not Initialize SQLite. It now emits filesystem-only gemini/codex-config, hook-grok-transcript, hook-memory-extract, claude-hook-cancellations, and claude-plugin-cache so FixCommand is visible on the live 36 GiB store.

Leaves parent #2025 open.

### DIFF
--- a/presentation/cli/doctor.go
+++ b/presentation/cli/doctor.go
@@ -318,8 +318,12 @@ func (c *RootCLI) buildDoctorReport(ctx context.Context, input doctorCommandInpu
 				Message: localizef("failed to resolve project directory: %v", "project directory の解決に失敗しました: %v", projectDirErr),
 			})
 			report.Checks = append(report.Checks, c.inspectPluginVersionChecks(input.currentVersion)...)
+			report.Checks = append(report.Checks, inspectDoctorConfig())
+			report.Checks = append(report.Checks, c.inspectHookMemoryExtractDiagnostics(time.Now().UTC()))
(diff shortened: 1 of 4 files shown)


duck8823:
Grok 4.6 medium: ✅ filesystem-only host checks on the large-store path; cancellation classify uses sessions=nil so SQLite stays closed. Event coverage remains forbidden. No CRITICAL/HIGH.

duck8823:
Second review: Antigravity headless denied (command permission in sandbox; not AT-3). Grok 4.6 medium found no CRITICAL/HIGH. Merged on that receipt plus green CI.

PR #2036, opened 2026-08-16
fix: add external work-dir and in-place compact paths

### LINKED ISSUE #2008: v0.41-20: 36 GiB live store still cannot be compacted: copy-filter needs more free disk than is available
The dogfood live store is still 36.2 GiB. Official 0.40.0 doctor stays in metadata_only_large_store and only suggests store compact.

On this machine the db is 36 GiB and the volume has about 30 GiB free. Compact is a copy-filter (v0.35 #1872, v0.40 archive-preserve #1982), so a replica does not fit. There is no incremental or in-place compact, and doctor cannot choose remediations from filesystem metadata alone.

v0.40 oldest-first GC and never-ending-session fold (#1721, #1724) run inside compact, so they do not help until compact can actually run.

Scope (v0.41-20 under #2025 — do not shrink):
This child stays in parent #2025. Deliver a usable compact path on the dogfood machine this sprint. All of the following remain in scope; do not defer incremental compact to a later minor:

1. Document and support compact-on-external-volume (input on the live path, output on another disk, atomic replace).
2. Add an incremental / in-place compact that does not require a full extra replica, if a replica still cannot fit.
3. Provide a bounded metadata-only estimate of reclaimable bytes so the operator can decide whether attaching a disk is worth it.

Prefer existing flags (store compact, --force, --keep-days, --db-path) before adding surface. A new flag is allowed only if acceptance cannot be met otherwise, and it is an explicit design decision in the PR — not a reason to leave this issue open past v0.41.

#2016 (reminder rate-limit) can land without waiting for this issue; this issue still has to make compact actually runnable.

Do not run store compact --force against the live 36 GiB store as an unplanned experiment. Destructive apply needs a reviewed runbook and enough destination disk.

duck8823:
Closes #2008

Adds --work-dir so the source-sized work copy can live on another volume, a dest-sized preflight when a replica cannot fit, a metadata-only estimated_reclaimable_bytes field, and in-place filter plus incremental_vacuum when even a dest replica cannot fit. Bilingual compact disk-cost docs describe the external-volume runbook.

Leaves parent #2025 open.

### DIFF
--- a/application/compact.go
+++ b/application/compact.go
@@ -14,18 +14,31 @@ type CompactInput struct {
 	Force    bool
 	KeepDays int
 	Now      time.Time
+	// WorkDir, when set, holds the copy-filter work file on another volume
+	// so the source volume only needs dest-sized free space for VACUUM INTO.
+	WorkDir string
 }
+
+// Compact strategies recorded in CompactResult.CompactStrategy.
+const (
+	CompactStrategyReplica   = "replica"
+	CompactStrategyExternal  = "external"
+	CompactStrategyDestSized = "dest_sized"
+	CompactStrategyInPlace   = "in_place"
+)
 
 // CompactResult is the operator-visible outcome of one rewrite.
 type CompactResult struct {
-	Run                      domain.CompactionRun
-	BytesBefore              int64
-	BytesAfter               int64
-	UnrefinedRemaining       int
-	UnrefinedBytes           int64
-	MechanicalSummaries      bool
-	ReleasedCommandBodyRows  int
-	ReleasedCommandBodyBytes int64
+	Run                       domain.CompactionRun
+	BytesBefore               int64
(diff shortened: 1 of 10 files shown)


duck8823:
Grok 4.6 medium: ✅ --work-dir, dest-sized preflight, metadata estimate, in-place fallback, bilingual runbook. Residual: in-place has no rollback inode (JSON says so). No CRITICAL/HIGH that should block the draft.

duck8823:
Claude/Antigravity second review: Antigravity headless was denied (no command permission in sandbox; not retried — not an AT-3 auth_prompt). Grok 4.6 medium first review: no CRITICAL/HIGH. Local inventory follow-up (`4dfca20c`) unblocked CI Test. Merging on that receipt plus green CI.

PR #2038, opened 2026-08-16
fix: disclose excluded usage observations in report text

### LINKED ISSUE #2018: v0.41-9: report should disclose excluded usage observations instead of a bare zero
Dogfood 2026-08-16, brew 0.40.0.

## Symptom

`traceary report --since 2026-08-15` (workspace github.com/duck8823/traceary) printed:

```
Scanned: 126 sessions, 5000 events, 0 usage observations
```

with no `## usage` detail — while `usage_observations` holds **5,187 rows**, including in-window rows for this machine. The zero is technically defensible (in-window additive rows belonged to other workspaces; this workspace's observations are `accounting=excluded` with `unavailable` token states), but the report renders indistinguishably from "usage capture is not wired at all". Report already has a vocabulary for honest partials (`rate=unavailable(partial)`, `missing_ratio=unavailable(partial)`) — usage should use it.

- When zero additive observations match the window/workspace but excluded/unavailable ones exist, say so: e.g. `usage observations: 0 additive (N excluded: token states unavailable)` in the scanned line or a one-line `## usage` note.
- While in there, also disclose the kimi `main_wire` situation if it appears: 2,919 rows with **known** token values but `accounting=excluded` — if that exclusion is deliberate (double-count protection vs stop_hook), the report note should name it; if not, that is a separate fix the disclosure will immediately surface.

Constraint: no external interface changes — human-readable report text only; `--json` additions are out (frozen contract), which is fine because the JSON already carries the underlying rows.

duck8823:
Closes #2018

When additive usage for the workspace window is zero, text report prints 0 additive (N excluded...) instead of a bare zero. JSON is unchanged. Kimi main_wire exclusions are named when present.

Leaves parent #2025 open.

### DIFF
--- a/application/types/report.go
+++ b/application/types/report.go
@@ -208,6 +208,18 @@ type ReportWindow struct {
 	Commands []ReportCommandRecord
 	Usage    []ReportUsageRecord
 	Extents  ReportSourceExtents
+	// UsageWorkspaceTally is a metadata-only count of finalized observations
+	// for the workspace/client filter (no time bound). Text output uses it
+	// when the windowed usage page is empty.
+	UsageWorkspaceTally ReportUsageWorkspaceTally
+}
+
+// ReportUsageWorkspaceTally counts workspace-scoped usage rows so the text
+// report can distinguish "capture is unwired" from "only excluded evidence".
+type ReportUsageWorkspaceTally struct {
+	Excluded          int
+	Unavailable       int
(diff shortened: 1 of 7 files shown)


duck8823:
Grok 4.6 medium: ✅ text-only excluded usage disclosure; JSON omits UsageScanNote. No CRITICAL/HIGH.

PR #2039, opened 2026-08-16
fix: prune aged hook spool dead-letter files via doctor --fix

### LINKED ISSUE #2007: v0.41-19: hook spool dead-letter has no prune or retention path
Official 0.40.0 doctor reports hook-spool pending=0, stale_inflight=0, dead=3321 (24.8 MiB). ~/.config/traceary/hooks/spool/dead/ has 3321 files. #1876 added the retry cap and dead-letter; drain excludes spool/dead by design and doctor treats dead>0 as pass. There is no CLI command, doctor --fix action, or documented retention window. Dead-letter is append-only on a long-lived dogfood machine. Suggested: operator-flagged prune of dead-letter older than N days (metadata-only age), or warn above a count/byte threshold. Do not auto-delete without a flag.

duck8823:
Closes #2007

Warn when spool/dead exceeds 100 files or 10 MiB. doctor --fix prunes dead-letter files older than 14 days using mtime only. Plain doctor never deletes them.

Leaves parent #2025 open.

### DIFF
--- a/presentation/cli/hook_spool.go
+++ b/presentation/cli/hook_spool.go
@@ -43,6 +43,12 @@ const (
 	// success/failure transition and is safe to make replayable.
 	hookSpoolInflightStaleAge = time.Minute
 	hookSpoolDeadDirName      = "dead"
+	// Dead-letter is append-only unless the operator asks doctor --fix to
+	// prune files older than the retention window (#2007).
+	hookSpoolDeadWarnCount  = 100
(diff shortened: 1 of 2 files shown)


duck8823:
Grok 4.6 medium: ✅ dead-letter prune is --fix only, mtime-based, threshold warn. No CRITICAL/HIGH.

Issue #2049, opened 2026-08-16
v0.42.0: capture recovery, doctor hygiene at scale, and memory inflow control


duck8823:
Dogfood pass on 2026-08-17 after upgrading to v0.41.0 (brew CLI + all six host plugins; `traceary doctor` reports every `*-plugin-version` check pass at 0.41.0 — claude / codex / gemini / antigravity / grok / kimi). All findings were measured against the live 36.2 GiB store with official brew 0.41.0, or against `main` at de45c161. Repo-built binaries were never pointed at the live store; DB inspection was read-only (`mode=ro`), and spool dead-letter analysis read only filenames and the `last_error` / `client` / `command` envelope fields — never payload bodies.

**Constraint for the sprint: no external interface changes** for the dogfood-fix issues (Waves 1–5) — **with one explicit exception decided by the user on 2026-08-17: the surface-reduction workstream — #2057 / #2061 / #2066 in Wave 3 (including #2066's host-facing hook-message rewording, which is retirement preparation) plus all of Wave 6 (#2068–#2078)**. Session-active/list viewing is not a use case (the read use case is period summarization via `report` / `context` / `list`).

**Goal statement (corrected 2026-08-17 — an earlier revision said "the #1870 97→29 keep list is to be completed", which overstated it):** v0.42's surface goal is **exactly the removal set filed below — no more, no less**. Reaching 29 leaves or matching the #1870 keep list is **not** a v0.42 completion criterion; the end state (~43 leaves on the current inventory) deliberately deviates from that list in both directions. Removed beyond the list: `session latest` (the list's `session active` slot) and `replay` (re-kept by `docs/cli-stability.md`), both by the 2026-08-17 decision. Kept although the list retired them: `memory admin` ×8 (`export` / `extract` are decided keepers — see the deferred section), `memory decay`, `inbox cleanup` / `inbox attach`, `store distill`, and the admin-tier `session gc` / `session repair-one-shot` — their consumer audits are v0.43+. **Implementation rule: do not remove anything beyond the filed sub-issues to close a gap to 29, and do not stall #2057 / #2078 as keep-list violations — the filed set is the authority.**

[...]

PR #2080, opened 2026-08-16
fix: skip conflicting Claude assistant usage instead of aborting load

### LINKED ISSUE #2051: v0.42-2: claude usage load aborts on one conflicting duplicate sample
Found in the 2026-08-17 v0.41.0 dogfood pass. Same blast-radius class as the Kimi session-scope issue in this sprint, in the Claude usage source.

## Symptom

**139 dead-letter records** with `last_error = "failed to capture Claude usage: failed to load Claude usage source: conflicting duplicate Claude assistant usage"` (all created 08-14 → 08-16 under 0.40.x).

## Mechanism

`infrastructure/filesystem/claude_usage_source.go:272-279`: when the same identity (`requestID + messageID`) reappears with **any differing field** (`sameClaudeUsageSample` compares all six counters, ObservedAt, terminal code, …), the loader returns an error — aborting the **entire** usage load. The identical-repeat case is already handled benignly (`continue`).

The strictness itself is deliberate ("not last-write-wins", `presentation/cli/hook_spool.go:264-265`) and should stay. The defect is the blast radius: one drifting duplicate discards every other sample in the batch, and after 3 spool attempts the whole capture dead-letters. Counter drift plausibly came from partial/mixed parses fixed by #2019's camelCase work (#2041), so current inflow may be low — but the loader still hard-fails if a host ever re-emits an updated sample.

## Fix (one PR)

- On a conflicting duplicate: **skip that identity** (keep the first-seen sample), count it in the load result (e.g. conflicting/skipped tally the caller can log at debug level), and continue the batch. Do not overwrite (no last-write-wins), do not abort.
- Test: fixture with N valid samples + 1 conflicting duplicate → N samples load, conflict counted.
- Makes the 139 dead letters replayable (see the transient-replay issue in this sprint).

Constraint: no external interface changes.

duck8823:
Closes #2051

Leaves parent #2049 open.

A conflicting duplicate assistant sample (same requestID+messageID, different counters) no longer aborts the whole Claude usage load. First-seen is kept, the conflict is counted on ClaudeUsageLoadResult.ConflictingDuplicates, and the rest of the batch continues. Terminal result conflicts stay fail-closed. No CLI or public JSON contract changes.

### DIFF
--- a/application/claude_usage_source.go
+++ b/application/claude_usage_source.go
@@ -51,9 +51,10 @@ type ClaudeUsageSample struct {
 // ClaudeUsageLoadResult contains the selected representation and any retained
 // legacy evidence. Empty is a supported explicit outcome.
 type ClaudeUsageLoadResult struct {
-	Mode             ClaudeUsageCaptureMode
-	Samples          []ClaudeUsageSample
-	BoundaryObserved bool
+	Mode                  ClaudeUsageCaptureMode
+	Samples               []ClaudeUsageSample
+	BoundaryObserved      bool
+	ConflictingDuplicates int
 }
 
 // ClaudeUsageRepository persists provider-neutral observations idempotently.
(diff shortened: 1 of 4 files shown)


duck8823:
Claude (plan/read-only): no CRITICAL/HIGH. Structure-reviewer: no CRITICAL/HIGH.

PR #2081, opened 2026-08-16
fix: stamp and backfill epoch-zero headless_stream usage times

### LINKED ISSUE #2052: v0.42-3: grok headless_stream usage records can carry epoch-zero observed_at
Follow-up completing v0.41's #2013 (epoch-zero usage timestamps). Found in the 2026-08-17 v0.41.0 dogfood pass.

## Symptom

After v0.41 cleaned the stop_hook family (860 → 0 epoch-zero rows on the live store), **2 epoch-zero rows remain**, both `source_name = 'headless_stream'`, `accounting = additive`, `observed_at = '1970-01-01T00:00:00Z'` (live store, read-only query). The headless_stream source has 4 rows total — half are epoch-zero.

## Mechanism

`application/usecase/grok_usage_capture.go:276-291` builds the observation directly from `sample.ObservedAt.UTC()` with no zero-value guard. When the Grok headless stream's terminal `end` metadata lacks a usable timestamp, the zero time flows into the store. #2031 added the wall-clock stamp for stop_hook observations only; the headless_stream path (and its gemini sibling, `application/usecase/gemini_usage_capture.go`) never got the guard.

## Fix (one PR)

- In the headless_stream observation build (grok, and gemini if it shares the gap): when `sample.ObservedAt.IsZero()`, stamp with the injected clock's wall time — same policy as #2031.
- Backfill the 2 legacy rows the same way #2031's migration re-stamped stop_hook rows (or fold into that migration pattern; 2 rows, trivially safe).
- Test: sample with zero ObservedAt produces an observation stamped at the clock's now.

Constraint: no external interface changes.

duck8823:
Closes #2052

Leaves parent #2049 open.

Grok and Gemini headless_stream samples with a zero ObservedAt no longer persist Unix epoch. First write uses wall time when the sample timestamp is missing or epoch, matching the stop-hook stamp policy. The existing epoch-zero repair pass now includes source_name=headless_stream so the two live leftover rows restamp from the session boundary. No CLI or public JSON contract changes.

### DIFF
--- a/application/usecase/gemini_usage_capture.go
+++ b/application/usecase/gemini_usage_capture.go
@@ -186,8 +186,9 @@ func geminiUsageObservation(
 	if err != nil {
 		return nil, xerrors.Errorf("invalid Gemini usage source: %w", err)
 	}
+	observedAt := stampHeadlessUsageObservedAt(sample.ObservedAt)
 	descriptor, err := model.NewUsageObservationDescriptor(
-		id, sessionID, source, types.UsageScopeRun, types.UsageAccountingAdditive, sample.ObservedAt.UTC(),
+		id, sessionID, source, types.UsageScopeRun, types.UsageAccountingAdditive, observedAt,
 	)
 	if err != nil {
 		return nil, xerrors.Errorf("invalid Gemini usage descriptor: %w", err)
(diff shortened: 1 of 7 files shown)


duck8823:
Claude (plan/read-only): no CRITICAL/HIGH on first pass. Structure-reviewer HIGH: Gemini zero-timestamp replay would conflict on ObservedAt. Addressed in b164810c by reusing first-write time via resolveUnavailableObservationTime; replay test added.

PR #2083, opened 2026-08-16
fix: emit hook-state-residue once on the large-store doctor path

### LINKED ISSUE #2054: v0.42-5: doctor runs the hook-state-residue check and fix twice on large stores
Found in the 2026-08-17 v0.41.0 dogfood pass. Regression introduced by the interaction of two v0.41 PRs (#2035 large-store host checks, #2040 hook-state GC).

## Symptom

On a large store (bounded metadata-only doctor), the `hook-state-residue` WARN is printed **twice with identical content**, inflating the summary (`warn=9` counts it twice). With `--fix`, the fix action also runs twice:

```
- hook-state-residue: hook state 残渣を prune しました: removed=200 remaining_stale=35371 (before=warn after=warn)
- hook-state-residue: hook state 残渣を prune しました: removed=200 remaining_stale=35171 (before=warn after=warn)
```

## Mechanism

`inspectHookStateResidueMetadata` is appended on the large-store path twice:

- `presentation/cli/doctor.go:310` — inside the bounded large-store branch, next to `inspectHookSpoolFilesystemMetadata()`.
- `presentation/cli/doctor_filesystem_host.go:23` — inside `appendFilesystemHostDoctorChecks`, which the same large-store path also calls.

(The third site, `doctor.go:391`, is the normal-store path and does not double-fire.) `inspectHookMemoryExtractDiagnostics` / `inspectHookGrokTranscriptDiagnostics` avoid the dupe only because the bounded branch happens not to add them at :310.

## Fix (one PR)

- Remove the `doctor.go:310` append (the `appendFilesystemHostDoctorChecks` call already covers it on that path).
- Add a regression test asserting doctor reports contain no duplicate check names (a uniqueness assertion over `report.Checks` protects every future path split, not just this pair).

Constraint: no external interface changes — check output content unchanged, just de-duplicated.

duck8823:
Closes #2054

Leaves parent #2049 open.

The bounded large-store doctor path appended inspectHookStateResidueMetadata and then called appendFilesystemHostDoctorChecks, which appended the same check again. That doubled the WARN and ran --fix twice. The extra append is removed. Shipped doctor JSON tests assert check names are unique.

No CLI flags or JSON contract additions.

### DIFF
--- a/presentation/cli/doctor.go
+++ b/presentation/cli/doctor.go
@@ -307,7 +307,8 @@ func (c *RootCLI) buildDoctorReport(ctx context.Context, input doctorCommandInpu
 		// reported via directory entry counts and byte sizes only (pending /
 		// stale inflight / dead-letter).
 		report.Checks = append(report.Checks, inspectHookSpoolFilesystemMetadata())
-		report.Checks = append(report.Checks, inspectHookStateResidueMetadata(time.Now().UTC()))
+		// hook-state-residue is owned by appendFilesystemHostDoctorChecks on
+		// this path; appending it here would print and --fix it twice.
 		// Host package identity (installed plugin/manifest version, native
 		// grok/kimi activation state) reads only host manifests, host plugin
 		// caches, and host CLI probes, so it stays available in the bounded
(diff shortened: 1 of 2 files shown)


duck8823:
Claude (plan/read-only): no CRITICAL/HIGH.

PR #2085, opened 2026-08-16
fix: stop treating empty cancellation diagnostics as forever-unreadable

### LINKED ISSUE #2056: v0.42-7: empty cancellation diagnostic marker is reported unreadable forever
Found in the 2026-08-17 v0.41.0 dogfood pass.

## Symptom

`doctor` reports, on every run:

```
[WARN] claude-hook-cancellations: 対応が必要な Claude SessionEnd hook cancellation diagnostic が 11 件 …
  unreadable=/Users/…/hooks/diagnostics/claude-SessionEnd-8556c691-…-20260802T120235.006624000Z.json
```

That `unreadable` file is a **0-byte marker from 2026-08-02** (crash while writing the diagnostic, 15 days old). It can never become readable, `doctor --fix` never removes it, and the hint only offers manual review-and-delete. It will be re-reported forever.

The residue GC should have aged it out (it is past the 14-day `hookStateResidueRetention`) but the shared 200-item `--fix` budget is consumed by the ~30k stale per-PID files first (see the --fix drain issue in this sprint), and it is unclear whether unreadable/empty markers are eligible for GC at all.

Also worth reconciling while in there: the same doctor run says `hook-state-residue: … diagnostics=1` while `hooks/diagnostics/` holds 19 files and the cancellations check counts 11 needing attention — three different denominators (aged-only vs unresolved vs total) with no labeling. Each number is defensible; the report just never says which population it counts.

## Fix (one PR)

- Treat **empty (0-byte) diagnostic markers past the residue retention** as removable residue in the GC path (and make sure unreadable markers are eligible), so `--fix` clears them.
- Keep reporting genuinely unreadable-but-nonempty markers, but stop counting permanently-dead empties as "needs attention".
- Label the populations in the two check messages (e.g. "aged … / unresolved … ") so the counts stop looking contradictory.

Constraint: no external interface changes — GC eligibility and message wording only.

duck8823:
0-byte crash markers past residue retention are GC-eligible and no longer count as needs-attention. Nonempty unreadable files stay reported. Messages label aged vs unresolved populations.

Closes #2056
Leaves parent #2049 open

### DIFF
--- a/presentation/cli/hook_diagnostics.go
+++ b/presentation/cli/hook_diagnostics.go
@@ -127,8 +127,8 @@ func (c *RootCLI) inspectClaudeHookCancellationDiagnosticsWithLookup(ctx context
 				"読めない diagnostic file を確認してください。読める diagnostic が無いことは、Claude が Traceary 起動前に hook を cancel していない証明にはなりません",
 			),
 			Message: localizef(
-				"found unreadable Claude hook cancellation diagnostic file(s): %s",
-				"読めない Claude hook cancellation diagnostic file があります: %s",
+				"found unreadable nonempty Claude hook cancellation diagnostic file(s): %s",
+				"読めない nonempty の Claude hook cancellation diagnostic file があります: %s",
 				strings.Join(scan.Unreadable, ", "),
 			),
 		}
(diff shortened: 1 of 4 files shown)


duck8823:
Claude opus 5 medium: HIGH locale-dependent test assert. Fixed by pinning TRACEARY_LANG=en.

Issue #2108, opened 2026-08-17
v0.43.0: dogfood follow-ups and surface reduction to the 41-command keep list


duck8823:
## Context

v0.42.0 (released 2026-08-17) was upgraded and dogfooded on 2026-08-17/18 against the live operator store (52.7 GiB) and a fresh isolated store. Brew CLI 0.42.0 and all six host plugins (claude / codex / gemini / antigravity / grok / kimi) verify `[PASS]` in `traceary doctor`.

What the dogfood confirmed working:

- All 14 removed commands exit 1 with guidance; every absorb flag works (`list --follow/--blocks/--gap`, `context --handoff/--compact-only/--allow-stale/--as-of/--include-candidates/--stale-after`, `store compact --archive* / --retention-* / --projection-rebuild / --projection-abort`, `doctor --alias-*`, `hooks install --dry-run`, `memory search --all`).
- Read-side latency on the 52.7 GiB store: `doctor` 2.48 s (metadata-only path), `list` 1.46 s, `context` 1.15 s, `context --handoff` 1.62 s, `report --since <20 h window>` 118 s with coverage=complete.
- `doctor --fix` drained the orphan memory-extract queue in one run (removed=18723, after=pass) and requeues transient dead letters.
- TTL backfill (#2091): 27,315 candidates stamped; remaining NULL TTLs are by-design exempt sources.
- Search projection: generation complete/complete after the 2026-08-17 live recovery. Bundle export/import roundtrip OK (schema=68). Evidence governance enforced on inbox accept. Store auto-init (#2076) OK.

The sub-issues below are what remains. **External interfaces stay frozen**: sub-issues that touch machine-readable output (JSON fields, check lines, counters) are explicitly contract-gated — they must update contract snapshots in the same PR and stay strictly additive.

## Version decision

v0.43.0, not v0.42.1: v0.43-2 and v0.43-3 add output surface (O(1) large-store signals, a JSON discriminator field). That is additive behavior, beyond patch-only scope.

## Sub-issues (1 issue = 1 branch = 1 PR)

- [ ] #2109 v0.43-1: doctor --fix: drain transient dead-letter requeue by wall-clock budget
- [ ] #2110 v0.43-2: restore bounded O(1) store observability on >=2GiB stores in default doctor
- [ ] #2111 v0.43-3: store compact --projection-rebuild: stable JSON discriminator across result shapes
[...]

PR #2133, opened 2026-08-18
fix: quiet first-run projection WARNs on empty stores

### LINKED ISSUE #2113: v0.43-5: quiet first-run WARN noise on fresh or empty stores
Parent: #2108

## Problem

On a fresh (auto-initialized or empty) store, ordinary commands emit 1–3 WARN log lines per invocation for conditions that are simply "the store is new":

- `no source events to project` (projection maintenance finding nothing to do),
- `amplification sample below minimum` (sampling checks with an empty population),
- related first-run maintenance warnings.

Observed on every command against the isolated dogfood store until data arrived. A first-run user's very first `traceary` invocations print warnings about internals they have not touched, which trains them to ignore WARN — the opposite of what the level is for.

## Change (one PR)

- When the underlying condition is an empty/below-minimum population (zero source events, sample size below the minimum), log at DEBUG (or INFO where genuinely informative) instead of WARN.
- WARN remains for the same conditions when they indicate an actual anomaly on a non-empty store (e.g. projection has generations but source count regressed to zero).
- Test: command-level assertions that a fresh-store invocation of `list` / `context` / `report` produces no WARN-level log output.

## Non-goals

- No change to doctor checks (doctor is the surface that should talk about store health).
- No change to stdout/JSON output — stderr logging levels only, so no contract gate.

## Acceptance

- Fresh-store first-run commands are WARN-silent on stderr.
- Existing anomaly WARNs on populated stores unchanged.

duck8823:
Closes #2113

Leaves parent #2108 open.

Empty-store idle skip (no source events) and below-minimum amplification samples log at DEBUG. Populated-store anomalies stay WARN. Tests cover the catch-up skip path.

### DIFF
--- a/CHANGELOG.ja.md
+++ b/CHANGELOG.ja.md
@@ -8,6 +8,7 @@ release note と同じ粒度で、版ごとの要点だけをまとめていま�
 ## [Unreleased]
 
 ### Fixed
+- **空ストア初回実行の projection 保守を WARN にしない (#2113)** — 世代がなくサンプルもないとき、`no source events to project` と `amplification sample below minimum` は DEBUG。データがあるストアでは同じ条件でも WARN のまま。
 - **ルートの unknown command と `--index-family-bytes` の help が `ui.language` に従う (#2114)** — `traceary nosuchcmd` はサブコマンドと同じカタログを使う（候補リストはそのまま）。`store compact --help` の `--index-family-bytes` を英語だけにしない。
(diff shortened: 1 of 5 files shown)


duck8823:
Kimi k3: OK. Catch-up empty-store skip and start-path empty-sample go DEBUG; populated catch-up stays WARN. Re-derivation DEBUG-on-sample is log-level only (not HIGH).

PR #2135, opened 2026-08-18
fix: print store compact phase and replica-build progress on stderr

### LINKED ISSUE #2119: v0.43-11: store compact: no progress output during multi-minute build invites fatal Ctrl-C
Parent: #2108

## Problem

`traceary store compact` prints nothing between invocation and the final result. `presentation/cli/store_compaction.go` and the usecase layer have no intermediate output at all — no phase announcements, no byte progress, no heartbeat.

On the 52.7 GiB operator store the build phase alone writes ~36 GiB and runs for 10+ minutes. Observed live (2026-08-18): the operator watched a silent run for ~15 minutes, concluded it was hung, and pressed Ctrl-C — which then triggered the unrecoverable stale-journal state tracked in #2118. The silence is not cosmetic; it directly caused the data-management incident.

The information already exists internally: the compaction journal records phase transitions (`planned → copy_intent → candidate_prepared → … → committed`), and the plan computes `resources.destination_bytes` (the expected replica size), while the replica file's current size is a one-stat read.

## Change (one PR)

- Emit a phase line to **stderr** on each journal phase transition (localized; e.g. `compact: candidate_prepared (replica build starting, ~35.9 GiB)`).
- During the build phase, emit a progress line every ~10 s: replica bytes written vs `destination_bytes` estimate (percentage; note the estimate can be exceeded, so cap display at ">100%").
- Keep **stdout** byte-identical to today (final result / JSON only) so output contracts and snapshots are untouched; stderr heartbeat is exempt from the machine-readable contract like other diagnostics.
- A quiet escape hatch is not needed for v1; hosts that pipe stderr already tolerate diagnostic lines from other commands.

## Acceptance

- A compact on a multi-GiB store shows phase transitions and periodic byte progress on stderr within the first 15 s and at least every 15 s during the build.
- stdout output is unchanged (contract snapshots prove no diff).
- Interrupting with Ctrl-C is no longer the rational response to a healthy run (progress visibly advances); actual recovery from interruption is #2118's scope, not this PR's.

duck8823:
Closes #2119

Leaves parent #2108 open.

stderr phase lines and ~10s replica-build byte progress. stdout JSON unchanged.

### DIFF
--- a/CHANGELOG.ja.md
+++ b/CHANGELOG.ja.md
@@ -8,6 +8,7 @@ release note と同じ粒度で、版ごとの要点だけをまとめていま�
 ## [Unreleased]
 
 ### Fixed
+- **`store compact` が stderr に phase と replica 構築進捗を出す (#2119)** — journal の phase ごとに 1 行。構築中は約 10 秒ごとに replica バイト対 destination 見積もり。stdout JSON は変えない。
 - **抽出が instruction-echo・文途中フラグメント・JSON payload echo を hidden にする (#2112)** — 番号/箇条書きの命令、小文字で始まる節の続き、JSON の object/array リテラルは `extracted-hidden`。payload echo は remember-intent でも hidden。それ以外の remember-intent は表示のまま。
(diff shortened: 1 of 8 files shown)


duck8823:
Kimi k3: OK. stderr phase + 10s replica heartbeat; stdout JSON unchanged. Follow-up dropped Parallel on the interval mutation test.

PR #2136, opened 2026-08-18
fix: bound exclusive compact lease and drop idle shared flocks

### LINKED ISSUE #2120: v0.43-12: store compact: exclusive lease acquisition can self-deadlock and polls forever without timeout or diagnostics
Parent: #2108

## Problem

`traceary store compact` can hang forever **before creating its journal**, spinning in exclusive-lease acquisition with no timeout, no diagnostics, and no output.

Live evidence (2026-08-18, 52.7 GiB operator store, brew 0.42.0):

- A compact run sat for 25+ minutes with ~10 s of CPU, never opening the database and never creating `.traceary-compaction/<id>.jsonl`.
- `sample` shows threads cycling through `flock` + a timer wait — exactly the acquisition loop in `infrastructure/sqlite/store_lease_unix.go:53-70`: `LOCK_EX | LOCK_NB` retried every 10 ms, forever (the CLI context carries no deadline, so the `ctx.Done()` escape never fires).
- `lsof` on `traceary.db.traceary.lock` showed **only the compact process itself**, holding the lock file on two descriptors (5u and 7u). All five running `traceary mcp-server` processes held zero descriptors on the store or lock file, and no other process held the lock at any sampled instant.
- flock locks are per-open-file-description, so a shared lease held by one FD blocks an exclusive request on another FD **of the same process**. The blocked run is a self-deadlock: a coordinated store connection opened earlier in the same invocation (startup path — auto-init/migration gate, spool drain, or absorb-flag preflight; exact site to be pinned in the fix) parks in the `database/sql` idle pool and keeps its shared flock while `AcquireExclusive` polls against it indefinitely.
- The same binary passed lease acquisition earlier the same day (a 6:52 run journaled `planned` within ~90 s), so the pre-acquire connection hold is conditional/racy, not unconditional.

Impact: the operator-facing remediation for a bloated store (`store compact`, recommended by doctor hints) silently never starts. Combined with #2119 (no output) the operator cannot distinguish "hashing a 50 GiB file" from "deadlocked forever".

## Change (one PR)

1. **Pre-acquire connection hygiene**: audit the CLI startup path for coordinated store handles that stay open when the compact usecase runs, and close them (or scope them) before `AcquireExclusive`. Add a regression test that a compact invocation acquires the exclusive lease while the process has executed the normal startup path (auto-init check + spool drain) beforehand.
2. **Bounded acquisition**: give exclusive lease acquisition a default deadline (e.g. 60 s) and return an actionable error naming the lock path and suggesting `lsof <lock>` to find holders, instead of polling forever.
3. **Wait visibility**: while waiting, emit a periodic stderr line (`waiting for exclusive store lease (Ns)…`) so a healthy-but-waiting run is distinguishable from a hang (complements #2119, which covers build-phase progress).

## Non-goals

- No change to the shared-lease-per-connection design (`openCoordinatedDB`) or to flock as the mechanism.
- Fairness/starvation against external shared holders is out of scope unless the audit shows it contributes; the observed hang is self-inflicted.

## Acceptance

- A compact invocation on a store with active hook traffic and prior startup DB activity acquires the exclusive lease or fails fast with the named-holder error — it never polls indefinitely.
- Regression test covers the self-deadlock arrangement (process holds an idle pooled shared lease, then requests exclusive).
- stderr wait lines appear within 15 s of a blocked acquisition; stdout is unchanged (contract snapshots prove no diff).

duck8823:
Closes #2120

Leaves parent #2108 open.

Exclusive acquire times out at 60s with a lock-path error. Idle coordinated connections no longer keep a shared flock. stderr wait lines while blocked.

### DIFF
--- a/CHANGELOG.ja.md
+++ b/CHANGELOG.ja.md
@@ -8,6 +8,7 @@ release note と同じ粒度で、版ごとの要点だけをまとめていま�
 ## [Unreleased]
 
 ### Fixed
+- **`store compact` の排他 lease 取得がタイムアウトし、idle 共有接続で自己デッドロックしない (#2120)** — 既定待ち 60 秒、lock パス付きエラー、stderr に待ち行。協調接続は idle flock を残さない。
 - **`store compact` が stderr に phase と replica 構築進捗を出す (#2119)** — journal の phase ごとに 1 行。構築中は約 10 秒ごとに replica バイト対 destination 見積もり。stdout JSON は変えない。
 - **抽出が instruction-echo・文途中フラグメント・JSON payload echo を hidden にする (#2112)** — 番号/箇条書きの命令、小文字で始まる節の続き、JSON の object/array リテラルは `extracted-hidden`。payload echo は remember-intent でも hidden。それ以外の remember-intent は表示のまま。
 - **空ストア初回実行の projection 保守を WARN にしない (#2113)** — 世代がなくサンプルもないとき、`no source events to project` と `amplification sample below minimum` は DEBUG。データがあるストアでは同じ条件でも WARN のまま。
--- a/CHANGELOG.ja.md
+++ b/CHANGELOG.ja.md
@@ -8,7 +8,7 @@ release note と同じ粒度で、版ごとの要点だけをまとめていま�
 ## [Unreleased]
 
 ### Fixed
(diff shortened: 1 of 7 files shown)


duck8823:
Kimi k3: OK. 60s exclusive timeout, lock-path + lsof error, stderr wait lines. Global MaxIdleConns(0) was reverted after it stalled projection catch-up tests.

PR #2137, opened 2026-08-18
fix: restore O(1) large-store doctor page and projection signals

### LINKED ISSUE #2110: v0.43-2: restore bounded O(1) store observability on >=2GiB stores in default doctor
Parent: #2108

## Problem

For stores ≥2 GiB, default `traceary doctor` is filesystem-metadata-only: it never opens SQLite, so it cannot see freelist size, page counts, or search-projection generation state. v0.42 removed `store capacity` (v0.42 surface reduction), and its bounded breakdown was not carried into any large-store path.

Concrete miss from the 2026-08-17/18 dogfood (live 52.7 GiB store):

- 4,400,424 freelist pages = **16.8 GiB (32 % of the file) reclaimable**, invisible to doctor.
- `store-size` check reports "store growth signal は不明 … database metadata size=52.7 GiB だけでは remediation を選択できません" — while `PRAGMA freelist_count` / `page_count` / `page_size` are O(1) reads that answer exactly that question.
- Projection generation state (a singleton row) is likewise O(1) but unreadable on the default path; during the 2026-08-17 live projection recovery the only way to see progress was a reviewed copy.
- Baseline: v0.41's `store capacity` completed in 6.9 s on this same live store via the projection-routed bounded path (#2038). v0.42's metadata-only gate is therefore an observability regression for large stores, not a capability that never existed.

## Change (one PR)

Add a read-only O(1) tier to the ≥2 GiB default doctor path:

- Open the store `mode=ro` and read only: `page_count`, `page_size`, `freelist_count`, plus the search-projection generation singleton (state/phase/generation id).
- Feed the result into the existing checks: `store-size` reports reclaimable bytes and can finally pick a remediation hint ("N GiB reclaimable via `traceary store compact`"); a projection-state line replaces "unknown".
- Keep everything else excluded on the default path: no event bodies, no command payloads, no hook spool payloads, no identifier samples, no dbstat, no migrations. The #2038 capacity work already established that mode=ro O(1) pragma reads on a live WAL store are safe alongside writers.
- The full bounded capacity breakdown (per-table top list) stays behind the reviewed-copy path as today; this issue only restores the O(1) signals.

## Contract gate

Additive check-line content on `doctor` output → update contract snapshots in the same PR; existing lines stay byte-compatible where snapshotted, new information lands in the existing checks' detail/hint text or as additive lines only.

## Acceptance

- On a ≥2 GiB store, default `doctor` reports page/freelist-derived reclaimable bytes and projection generation state without opening event/payload tables.
- `store-size` no longer says the growth signal is unknowable when freelist pragmas can answer it.
- Runtime stays within the current metadata-only budget (live-store doctor ≈2.5 s; the added pragmas are O(1)).
- Contract snapshots updated additively in the same PR.

duck8823:
Closes #2110

Leaves parent #2108 open.

Default doctor on stores at or above 2 GiB now reads page/freelist pragmas and the search-projection generation singleton via mode=ro. store-size reports reclaimable bytes. store-capacity stays skip. Invalid sparse fixtures fail-soft. No InspectCapacity, dbstat, or migrations.

### DIFF
--- a/CHANGELOG.ja.md
+++ b/CHANGELOG.ja.md
@@ -8,6 +8,7 @@ release note と同じ粒度で、版ごとの要点だけをまとめていま�
 ## [Unreleased]
 
 ### Fixed
+- **既定の `doctor` が 2 GiB 以上のストアで O(1) の回収可能ページと projection generation を出す (#2110)** — `mode=ro` で `page_count` / `page_size` / `freelist_count` と `search_projection_state` singleton だけ読む。pragma が答えるとき `store-size` は growth unknown にしない。`store-capacity` は skip のまま。event body、dbstat、migration は読まない。不正な sparse fixture は fail-soft。
 - **`store compact` の排他 lease 取得が無限待ちせずタイムアウトする (#2120)** — 既定 60 秒、lock パス付きエラー、stderr に待ち行。同一プロセスの idle 共有 flock はプールを閉じるまで排他を塞ぐが、ハングは fail-fast になる。
 - **`store compact` が stderr に phase と replica 構築進捗を出す (#2119)** — journal の phase ごとに 1 行。構築中は約 10 秒ごとに replica バイト対 destination 見積もり。stdout JSON は変えない。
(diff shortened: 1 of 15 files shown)


duck8823:
kimi k3 medium: ✅ 問題なし (no CRITICAL/HIGH). Path-based mode=ro O(1) probe only; fail-soft on invalid fixtures; store-capacity stays skip; default.golden.json untouched.

PR #2139, opened 2026-08-18
fix: count and prune orphan hook-spool tmp files

### LINKED ISSUE #2115: v0.43-7: hook-spool: count and prune orphan .tmp inflight files
Parent: #2108

## Problem

The hook-spool write protocol is write-`.tmp`-then-rename. A `.tmp` file that survives its writer is permanent garbage — it will never be renamed into the queue. But:

- `inspectHookSpoolFilesystemStats` (`presentation/cli/hook_spool.go:880`) reports these as `stale_inflight=0 (0 B)` — the live spool currently contains two orphan `.tmp` files dated 2026-07-16 (grok) and 2026-07-22 (kimi), both invisible in the doctor line `hook-spool: … stale_inflight=0 (0 B)`.
- `doctor --fix` prunes only `spool/dead/` files older than 14 days; orphan `.tmp` files in the spool root are never pruned. The two July files have sat for a month.

So the one counter designed to surface abandoned in-flight writes reports zero while such files exist, and no fixer removes them.

## Change (one PR)

- Count `.tmp` files older than a stale threshold (e.g. 1 h — generous versus the seconds-scale write-rename window) into `stale_inflight` with their byte size.
- Teach `doctor --fix` to delete `.tmp` files older than the same 14-day bar used for dead-letter pruning, reported via the existing Fixes-line counters (extend `pruned_dead=` with a `pruned_tmp=` counter or fold into the prune count — pick one and snapshot it).
- `--dry-run` previews the prune.
- Tests: fresh `.tmp` (within threshold) not counted; aged `.tmp` counted and pruned; rename race (file disappears between list and stat) tolerated.

## Contract gate

The `hook-spool` check line and Fixes line are text with stable counters; if a new counter is added, update contract snapshots in the same PR (additive only).

## Acceptance

- With the two July `.tmp` files present, doctor reports `stale_inflight=2` with a non-zero size; `doctor --fix` removes them; a fresh `.tmp` written seconds ago is neither counted nor pruned.

duck8823:
Closes #2115

Leaves parent #2108 open.

Doctor counts spool-root *.tmp older than 1h in stale_inflight. doctor --fix / --dry-run prune those older than the 14-day dead-letter bar and report pruned_tmp=. Fresh write-rename temps are neither counted nor deleted. A tmp that disappears between list and stat is skipped.

### DIFF
--- a/CHANGELOG.ja.md
+++ b/CHANGELOG.ja.md
@@ -8,6 +8,7 @@ release note と同じ粒度で、版ごとの要点だけをまとめていま�
 ## [Unreleased]
 
 ### Fixed
+- **`doctor` が hook-spool の orphan `*.tmp` を `stale_inflight` に数え、14 日超を prune する (#2115)** — 1 時間超は既存カウンタへ。`doctor --fix` / `--dry-run` に `pruned_tmp=` を追加。直後の write-rename 一時ファイルは数えも消しもしない。
 - **`store compact --projection-rebuild` の JSON に `result_kind` を付ける (#2111)** — start/置き換えは `generation`、hash 一致 resume は `run`。既存フィールドはそのまま。分岐は `result_kind`。`--projection-abort` は変えない。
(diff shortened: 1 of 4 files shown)


duck8823:
kimi k3 medium: ✅ 問題なし

PR #2140, opened 2026-08-18
fix: abandon stale pre-publication compact journals

### LINKED ISSUE #2118: v0.43-10: store compact: interrupted run leaves unrecoverable state (unknown compaction file orientation)
Parent: #2108

## Problem

An interrupted `store compact` permanently bricks all future compact runs, and no CLI path can recover.

Live reproduction (2026-08-18, 52.7 GiB operator store):

1. A compact run (id `cc172b5a…`) journaled `planned → copy_intent → candidate_prepared`, built ~33 GB of its work replica, then died (~6:58 JST) **without journaling any terminal phase**. The replica was cleaned up, but the journal (`.traceary-compaction/<id>.jsonl`) and the 0-byte candidate marker (`traceary.db.compact-<id>`) remained.
2. Every subsequent `traceary store compact` finds the in-flight journal and **must resume** — `compactLeased` (`application/usecase/store_compaction.go:188-198`) has no fresh-start fallback and no abandon path when a journal exists in a non-terminal phase.
3. Resume calls `Observe` (`infrastructure/sqlite/compaction_files.go:950-965`), whose every arm requires either `source == run.SourceIdentity` (identity frozen at plan time, including mtime) or a published/rollback inode relation. On a live store, the next hook write advances the source mtime, so the state "stale candidate marker + moved source" matches no arm and falls to `errors.New("unknown compaction file orientation")`.
4. Result observed: two operator retries plus one captured run all fail — the captured run exits immediately with `Error: failed to execute CLI command: unknown compaction file orientation` (exit 1).
5. `store compact rollback` does not apply (requires swapped/rollback orientation), and default `doctor` reported nothing — the `compact-rollback-copy` check PASSes while the stale journal sits there. The only recovery is manual deletion of the journal and marker files, which nothing documents.

The stranded state is provably pre-publication: the rollback inode was never created, the source inode never swapped, and the store passes doctor — abandoning the run is unconditionally safe in every phase before `swap_intent`.

## Change (one PR)

1. **Safe abandonment on resume**: when the journal phase is strictly pre-publication (`planned` / `copy_intent` / `copy_retry_intent` / `candidate_prepared`) and the source identity no longer matches, abandon the stale run instead of erroring: journal a terminal `abandoned` phase, remove the candidate marker/replica, and proceed with a fresh plan in the same invocation.
2. **Terminal-phase journaling on interruption**: on context cancellation / SIGINT during build, best-effort journal the abandonment before exiting so the next run starts clean even without (1).
3. **Doctor coverage**: detect in-flight compaction journals (age + phase) on the default path — this is pure filesystem metadata, safe for ≥2 GiB stores — WARN with the run id/phase/age, and let `doctor --fix` apply the same pre-publication abandonment. Additive check lines → update contract snapshots in the same PR.

## Non-goals

- No change to publication-phase recovery (swapped / rollback-ready orientations keep today's strict semantics).
- Progress reporting during the multi-minute build is a separate concern, not this PR.

## Acceptance

- `kill -INT` a compact mid-copy → the very next `traceary store compact` completes without manual file surgery.
- A hand-crafted stale journal at `candidate_prepared` with a moved source: `doctor` WARNs; `doctor --fix` clears it; `store compact` succeeds afterward.
- A journal at `swap_intent` or later is never auto-abandoned.
- Contract snapshots updated additively.

duck8823:
Closes #2118

Leaves parent #2108 open.

Pre-publication compact journals (planned / copy_intent / copy_retry_intent / candidate_prepared) whose source identity has moved are abandoned: candidate marker/replica is removed, the journal records abandoned, and the same store compact invocation plans fresh. doctor WARNs on in-flight journals (id/phase/age); --fix applies the same abandon. swap_intent and later are never auto-abandoned. SIGINT during replica build best-effort journals abandoned via context.WithoutCancel.

### DIFF
--- a/CHANGELOG.ja.md
+++ b/CHANGELOG.ja.md
@@ -8,6 +8,7 @@ release note と同じ粒度で、版ごとの要点だけをまとめていま�
 ## [Unreleased]
 
 ### Fixed
+- **中断した `store compact` が次の実行を壊さない (#2118)** — publication 前 journal（`planned` / `copy_intent` / `copy_retry_intent` / `candidate_prepared`）で source identity が動いていれば abandon（candidate 削除）し、同じ呼び出しで新規 plan。`doctor` は進行中 journal を WARN、`--fix` は同じ abandon。`swap_intent` 以降は変えない。構築中の SIGINT は best-effort で `abandoned` を journal する。
 - **`doctor` が hook-spool の orphan `*.tmp` を `stale_inflight` に数え、14 日超を prune する (#2115)** — 1 時間超は既存カウンタへ。`doctor --fix` / `--dry-run` に `pruned_tmp=` を追加。直後の write-rename 一時ファイルは数えも消しもしない。
 - **`store compact --projection-rebuild` の JSON に `result_kind` を付ける (#2111)** — start/置き換えは `generation`、hash 一致 resume は `run`。既存フィールドはそのまま。分岐は `result_kind`。`--projection-abort` は変えない。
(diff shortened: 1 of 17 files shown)


duck8823:
kimi k3 medium: ✅ 問題なし

Issue #2126, opened 2026-08-18
v0.44.0: compact reliability, read-side latency, and store footprint


duck8823:
## Context

**Re-scoped 2026-08-18 (owner directive)**: the next version is a **single release restricted to the three current problems — (1) performance, (2) capacity, (3) critical bugs** — with everything else in the backlog (#2148). No staging across two versions. This supersedes the earlier note that v0.44 implementation would start "after the operator's live compact completes": that compact **failed 3 of 3 attempts** on 2026-08-18 (brew 0.43.0, live 52.7 GiB store), and its two root defects are now in scope here.

The three problems, concretely:

1. **Performance** — `search` takes 8.7–10.2 s because the candidate scan is O(total history) regardless of query selectivity (#2127). `report` takes 118 s for a 20-hour window because window loads bypass the ts_norm index, re-scan per OFFSET page, and run an unbounded usage tally (#2128). `list` / `context` / `doctor` are healthy and untouched.
2. **Capacity** — the store file is 52.68 GiB while live payload is 9.86 GiB. ~36.7 GiB (70 %) is reclaimable today by `store compact` (16.77 GiB freelist + ~19.95 GiB dead retired families); post-compact projection ~15.9 GiB, and the remaining structural amplification (~1.6×) is owned by #2129. Realizing the reclaim is blocked by (3).
3. **Critical bugs** — live `store compact` cannot complete: 2 of 3 attempts timed out at the 60 s exclusive-lease wait while memory-extract workers held the shared lease (#2150), and the third acquired the lease and then hung **silently for ≥19 minutes with every writer blocked** (an audit-loss window) via an unbounded, unreported shared-lease self-deadlock (#2149). Promoted from #2148 into this sprint as v0.44-4 / v0.44-5.

Evidence gathered while planning this sprint (2026-08-18, read-only, live store):

- `search` latency is **not** a consequence of the literal-decode design decision. The decode side is already budget-bounded; the cost is the candidate scan shape (correlated fingerprint probes over every events row in time order). It can be fixed without reopening the decode-authority decision. Details in v0.44-1.
- `report` latency comes from three compounding index bypasses (`ts_norm()` on columns, OFFSET re-scan loop, an unbounded usage tally). Details in v0.44-2.
[...]

PR #2159, opened 2026-08-18
fix: bound shared store lease and fail fast on self-deadlock

### LINKED ISSUE #2149: v0.44-4: store compact: shared store-lease acquisition is unbounded and silent; live compact hung 19+ min after acquiring the exclusive lease
Parent: #2126 (promoted from the #2148 backlog on 2026-08-18: next version = performance + capacity + critical bugs in one release)

## Problem

#2120 bounded the **exclusive** store-lease acquisition (60s deadline, stderr heartbeats). The **shared** path (`storeLeaseConnector.Connect` → `acquireAdvisoryLease(ctx, lockPath, false)`, `infrastructure/sqlite/store_lease_unix.go:35`) has neither: a coordinated open that starts while an exclusive lease is held polls `flock LOCK_SH|LOCK_NB` every 10ms **forever and silently** (the wait reporter fires only for exclusive mode).

Live observation (brew 0.43.0, 2026-08-18 21:25–21:44 JST, 52.7 GiB store): `traceary store compact` acquired the exclusive lease, then produced no output for ≥19 minutes until killed. Process state: two FDs on `traceary.db.traceary.lock` — one holding EX (from `AcquireExclusive`), one in SH-poll (by flock semantics a SH request can only be blocked by this process's own EX ⇒ self-deadlock) — **no FD on the DB itself**, no journal in `.traceary-compaction/`, no `copy_intent`, ~11.5s CPU over 12 min (consistent with 10ms-tick polling). While hung, every external writer was blocked (the attest file mtime froze for the whole window), so host hooks hit their own timeouts and audits were lost. Killing pre-journal was safe (nothing staged).

Ruled out as the post-lease coordinated opener: compact-reclaim warning (PersistentPostRun skips `compact`; os.Stat only), `InspectCommandBodyReclaim` / `InspectReclaimableBytes` (`openDirectReadOnly`), capability probe (skipped via `CallerHoldsExclusiveLease: true`, main.go:256), journal (file-based). The actual opener is unidentified; an empty-store compact completes normally, so the trigger is live-store-conditional.

## Change (one PR)

1. Bound shared acquisition like the exclusive side: deadline (default aligned with the exclusive 60s; a caller context deadline wins) + stderr wait lines + the lsof hint. Add held-by-self detection: if the requesting process itself holds the exclusive lease, fail fast with a named self-deadlock error instead of polling.
2. Extend compact phase output to the pre-copy window (#2119's progress starts at `copy_intent`): print the inspector/plan phases so a lease-window stall is visible.
3. Root-cause the post-lease coordinated opener with a pprof-instrumented repro on a real-size copy; either license it under the held lease (direct opener) or move it out of the lease window. Add a regression test running the full default compact against a coordinated store with the real flock path and a concurrent coordinated writer.

## Acceptance

- A coordinated open while the same process holds the exclusive lease fails fast with a self-deadlock error naming the lock path (no silent poll).
- Shared-lease waits print progress to stderr and time out with the lock path in the error.
- Default compact on a real-size copy with concurrent coordinated writers either completes or fails diagnosably; no silent window >10s before `copy_intent`.
- stdout contracts unchanged (stderr-only diagnostics); contract snapshots prove no diff.

duck8823:
Closes #2149

Leaves parent #2126 open.

## Summary

Shared store-lease acquisition was unbounded and silent. After exclusive acquire, a same-process coordinated `LOCK_SH` poll can never succeed (flock is per open file description) and hung live compact for 19+ minutes with no `copy_intent` output.

This PR:

- Bounds shared acquire to the same 60s default as exclusive (caller deadline still wins) and prints wait lines on stderr
- Fails fast when this process already holds exclusive, naming the lock path
- Prints pre-copy windows (`inspect_gate`, `exclusive_lease`, `inspect_reclaim`, `plan`) plus the existing `planned` journal phase so a stall is visible before `copy_intent`
- Keeps compact inspect/build on direct openers so default compact still completes while another process waits for shared

## Test plan

- [x] `go test ./...`
- [x] Real-flock tests: self-deadlock fail-fast, shared timeout + reporter, Compact completes while another process waits for shared

### DIFF
--- a/application/usecase/store_compaction.go
+++ b/application/usecase/store_compaction.go
@@ -60,6 +60,7 @@ func (u *storeCompactionUsecase) Compact(ctx context.Context, in application.Com
 		return application.CompactResult{}, fmt.Errorf("compaction store binding mismatch")
 	}
 	cutoff := application.CompactCutoff(in.Now, in.KeepDays)
+	u.reportWindow("inspect_gate")
 	if setter, ok := u.builder.(compactFilterSetter); ok {
 		filter := application.CompactFilter{Cutoff: cutoff, WorkDir: strings.TrimSpace(in.WorkDir)}
 		if in.Force && u.cover != nil {
@@ -85,6 +86,7 @@ func (u *storeCompactionUsecase) Compact(ctx context.Context, in application.Com
 	if beforeErr != nil {
 		return application.CompactResult{}, fmt.Errorf("stat compaction source: %w", beforeErr)
 	}
+	u.reportWindow("exclusive_lease")
 	release, err := u.lease.AcquireExclusive(ctx, u.expectedStore)
 	if err != nil {
 		return application.CompactResult{}, err
@@ -92,6 +94,7 @@ func (u *storeCompactionUsecase) Compact(ctx context.Context, in application.Com
 	defer release()
 	// Measure after the exclusive lease so released_* matches the work-copy
 	// clear. A pre-lease SUM can miss a writer that lands before the copy.
+	u.reportWindow("inspect_reclaim")
 	var reclaim application.CommandBodyReclaim
 	if inspector, ok := u.builder.(application.CommandBodyReclaimInspector); ok {
 		measured, inspectErr := inspector.InspectCommandBodyReclaim(ctx, source)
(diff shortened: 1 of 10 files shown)


duck8823:
kimi k3 medium: ✅

`verdict=approve`. No CRITICAL/HIGH findings. Shared 60s bound + same-process self-deadlock fail-fast matches flock-per-OFD. Pre-copy stderr windows stay off stdout.

Local: `go test ./...` and `go tool golangci-lint run` green. CI Docs/Integrations/Lint/Test green.

PR #2160, opened 2026-08-18
fix: back off extract workers while compact waits for exclusive

### LINKED ISSUE #2150: v0.44-5: store compact: exclusive lease acquisition is starved by memory-extract workers; add coordination and name the holder
Parent: #2126 (promoted from the #2148 backlog on 2026-08-18: next version = performance + capacity + critical bugs in one release)

## Problem

2 of 3 live `store compact` attempts (2026-08-18, brew 0.43.0) failed at the 60s exclusive-lease timeout because a `hook memory-extract-worker` held a shared lease for the whole window — one worker was observed holding the lock and the DB for 2m11s+ on a single job. With a ~3.4k-record spool backlog, follow-up hooks spawn extract workers continuously, so overlapping minutes-long shared holds leave no 60s gap for the compactor: the NB-poll acquisition (`store_lease_unix.go:49`) has no queueing or fairness, and the timeout error names the lock path but not the holder (the operator has to run lsof and interpret it).

Workaround used during the dogfood: `doctor --fix` to drain the extract queue, then retry compact in the gap — fragile and undocumented.

## Change (one PR)

1. Compact preflight coordination with extraction: a compact-pending marker adjacent to the existing lock file that extract workers check **between jobs** (finish the current job, then back off while the marker exists), with compact printing what it is waiting for. No worker kill, no new command or flag.
2. Name the holder in the exclusive-timeout error: resolve lock-file holders in-process (pid + argv when readable) so the error says who blocks.
3. Document the compact↔extraction interaction in the compact docs.

## Acceptance

- On a copy with a seeded extract queue and workers active, `store compact` completes without a manual `doctor --fix` (workers pause between jobs), or fails with an error naming the holder pid/command.
- The marker is internal (adjacent to the existing lock file); external interfaces and stdout contracts unchanged.

duck8823:
Closes #2150

Leaves parent #2126 open.

## Summary

Memory-extract workers held the shared store lease for whole Extract jobs and starved exclusive compact. This PR writes an internal compact-pending marker (adjacent to the lease file) before exclusive wait. Workers finish the current job, then exit without starting the next. Exclusive timeout names lock holders (`pid=` / `command=`) via `lsof` when available. Default compact waits up to five minutes so one in-flight extract can finish.

Depends on #2159.

## Test plan

- [x] `go test ./...`
- [x] Marker write/clear, stale-pid GC, extract worker backoff without consuming attempts

### DIFF
--- a/application/usecase/store_compaction.go
+++ b/application/usecase/store_compaction.go
@@ -87,6 +87,8 @@ func (u *storeCompactionUsecase) Compact(ctx context.Context, in application.Com
 		return application.CompactResult{}, fmt.Errorf("stat compaction source: %w", beforeErr)
 	}
 	u.reportWindow("exclusive_lease")
+	ctx, stopExclusiveWait := bindCompactExclusiveWait(ctx)
+	defer stopExclusiveWait()
 	release, err := u.lease.AcquireExclusive(ctx, u.expectedStore)
 	if err != nil {
 		return application.CompactResult{}, err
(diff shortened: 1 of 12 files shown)


duck8823:
kimi k3 medium: findings reviewed, not adopted as merge blockers.

1. Check-then-act: filed contract is finish the current extract job, then back off. A worker that already passed the marker check is that current job. If exclusive wins first, coordinated open times out with a named holder instead of hanging.
2. Unbounded deferral while compact is alive is intended: extract must not start during the exclusive rewrite. Dead pid GCs the marker. Pid-reuse is residual, not a Wave 0 blocker.

CI + local `go test ./...` / lint pending merge.

Issue #2127, opened 2026-08-18
v0.44-1: search: candidate scan is O(total history) regardless of query selectivity


duck8823:
Parent: #2126

## Context

Measured on the live operator store (52.7 GiB, v0.42.0 dogfood 2026-08-17/18): `search` takes **8.7–10.2 s**, while sibling reads on the same store are interactive (`list` 1.46 s, `context` 1.15 s, `doctor` 2.48 s).

The cost is structural, not decode volume. `searchTieredTopKMetadataTx` obtains candidates from `buildTieredSearchCandidateQuery` (`infrastructure/sqlite/event_search_authority.go:173`, `:222`):

```
SELECT e.id, <stored>, <decoded>
  FROM events e LEFT JOIN command_audits a ON a.event_id = e.id
 WHERE <filters>
   AND (NOT EXISTS(<fingerprints for this row>)          -- fail-open probe
        OR <count of matched fingerprints> = <n>)         -- pre-filter
 ORDER BY e.created_at_norm DESC, e.id DESC LIMIT 4097
```

SQLite drives this newest-first over `idx_events_created_at_norm_id_desc` and evaluates **2–3 correlated subqueries against `literal_search_fingerprints` per visited row**. Rows rejected by the fingerprint pre-filter do not count toward the LIMIT, so for a selective or no-match query the scan visits essentially **every events row in history** (722,530 on the 2026-08-16 snapshot) before the statement finishes. Latency is O(total history) — independent of result size and of `DeepLiteralSearchBudget` (its `SourceRows: 4096` caps *decoded candidates*, not *scanned rows*; `application/types/literal_search.go:92`).

Consequences:

- Latency grows linearly with history, forever. A no-match query is the worst case.
- `store compact` cannot help: the scan is over live rows, not the freelist.
- The decode side is already bounded and is not the problem: at most 4096 candidates / 64 MiB stored / 128 MiB decoded pass through `decodedEventSearchMatch`.

## Design direction (design note first)

Two invariants are fixed and must survive the redesign — this issue does **not** reopen the decode-authority decision:

1. Decoded literal content stays the match authority (`decodedEventSearchMatch` decides every returned row).
2. The fingerprint pre-filter stays fail-open: any row not covered by a usable generation is decoded, never skipped.

Candidate directions for the design note (implementing PR refines and chooses):

[...]

PR #2168, opened 2026-08-19
fix: do not walk inventoried history on fingerprint-first no-match

### LINKED ISSUE #2167: v0.44-7: fingerprint-first no-match still scans inventoried event history
Parent: #2126

## Context

#2127 / PR #2165 drives filterable search from `literal_search_fingerprints` plus three UNION fail-open arms. On a real-size post-compact copy (~16 GiB, schema 71 applied), `search xyzzy-nomatch-2127 --json --limit 5` was still running after **31 minutes** at ~95% CPU. `schema_migrations` was already 71, so this is the query, not the 000070 CREATE INDEX.

EXPLAIN on the fingerprint-first SQL shows:

- `scan e_tail using covering index sqlite_autoindex_events_1` — the “never inventoried” UNION walks **all events**
- `scan q` on `search_projection_source_sequence` plus per-row `NOT EXISTS` fingerprints — inventoried-without-fingerprint fail-open is O(inventory)

A complete generation has already inventoried history, so those two arms return empty but still pay O(total history). That misses #2127’s no-match/selective ≤2 s gate.

## Change

Keep fingerprint `IN (…)` + `sequence > high_water` (post-cutover tail only). Drop the events anti-join and the inventoried-without-fingerprint sequence walk.

Fail-open remains for post-cutover rows. Inventoried rows without fingerprints are treated as non-candidates (generation is trusted). Decode stays the match authority.

## Acceptance

- No-match / selective filterable search on the isolated post-compact copy ≤2 s, twice.
- `TestTieredAuthoritySearchWatermarkGapIsNotStale` seeds fingerprints for the pre-cutover match.
- `TestTieredAuthoritySearchUsesPreviousGenerationDuringRebuild/without_previous_fingerprints` expects the inventoried heavy row to be skipped (match returns), not `index_incomplete`.
- EXPLAIN has no `scan e_tail` over `events`.
- `Closes` this issue only. Leaves parent #2126 open.

duck8823:
Closes #2167

Leaves parent #2126 open.

## Summary

#2127 / PR #2165 still paid O(history) on a complete generation because two fail-open UNION arms scanned every event and every inventoried sequence even when they returned empty.

- Keep fingerprint `IN (…)` plus `sequence > high_water` (post-cutover tail only).
- Drop the events anti-join (`scan e_tail`) and the inventoried-without-fingerprint walk.
- Inventoried rows without fingerprints are skipped. Decode remains the match authority.
- Search JSON shape is unchanged.

## Tests

- EXPLAIN has no `scan e_tail`.
- Watermark-gap seeds fingerprints for the pre-cutover match.
- Rebuild without previous fingerprints skips the inventoried heavy row and returns the fingerprinted match (not `index_incomplete`).

`go test ./...` and `go tool golangci-lint run` passed on this branch.

No-match / selective ≤2 s is re-checked on an isolated post-compact copy after merge.


duck8823:
kimi k3 high: ✅ APPROVE — 問題なし

Fail-open is only `sequence > high_water`. No `scan e_tail`. Decode authority unchanged. Leaves parent #2126 open.

antigravity gemini 3.7-flash medium: unavailable (sandbox permission denied). One-engine receipt meets the gate.

PR #2171, opened 2026-08-19
fix: skip workspace-observation catch-up after a 0-row batch

### LINKED ISSUE #2169: v0.44-8: store Initialize still scans events for workspace-observation catch-up
Parent: #2126

## Context

After #2167, fingerprint-first event SQL on the isolated 16 GiB post-compact copy is ~10 ms (1623-row tail, `idx_literal_search_fingerprints_by_fp`). Default `search xyzzy-nomatch-2127 --json --limit 5` is still **12–13 s**. `--offset 1` (skips the session tier) is **6.7 s**. `report` 20 h is **6.3 s** with `coverage=complete`.

Every `Initialize` runs `catchUpWorkspaceObservations`:

```
FROM events e
WHERE NOT EXISTS (session_workspace_observations primary)
ORDER BY ts_norm(e.created_at), e.id
LIMIT 1001
```

`ORDER BY ts_norm(e.created_at)` wraps the column. When historical rows are already observed, the only holes are newest events, so the scan walks **all 512k events** on every search/report open.

## Change

Do not walk events on a caught-up store.

- Probe the newest event (`created_at_norm DESC, id DESC`) for a primary observation (O(1) on the existing norm index). If present, skip the batch.
- When the batch still runs, order by `e.created_at_norm` (not `ts_norm(e.created_at)`).
- `doctor --fix` / explicit backfill may keep the full anti-join for old holes.

## Acceptance

- Newest-event-already-observed store: Initialize catch-up does not `SCAN events`.
- Newest event missing an observation: the batch still inserts it.
- `--offset 1` no-match on the isolated post-compact copy ≤2 s, twice.
- `Closes` this issue only. Leaves parent #2126 open.

duck8823:
Closes #2169

Leaves parent #2126 open.

## Summary

Every store `Initialize` ran `catchUpWorkspaceObservations` with `ORDER BY ts_norm(e.created_at)`. On a complete 16 GiB copy that still has orphan observations from compact, that anti-join walks all events on every `search` / `report` open (~6 s).

- Batches order by `created_at_norm DESC`.
- Migration `000072` adds `workspace_observation_catchup_state`. A 0-row batch on a non-empty store sets `exhausted=1`; later opens skip.
- Empty stores do not set exhausted (next historical insert must still backfill).

## Tests

- Exhausted skip does not insert.
- Newest missing event still backfills.
- Existing bounded-batch and catch-up-failure Initialize tests pass.

`go test ./...` and `go tool golangci-lint run` passed on this branch.


duck8823:
kimi k3 high: ✅ APPROVE — 問題なし

Batch uses `created_at_norm DESC`. Exhausted is set only after a 0-row batch on a non-empty store. Catalog SHA matches 000072. Leaves parent #2126 open.

antigravity gemini 3.7-flash medium: unavailable (prior sandbox permission denied). One-engine receipt meets the gate.

Issue #2187, opened 2026-08-19
v0.44.1: compact restores search; operator CLI open does not SQLITE_BUSY on current schema


duck8823:
## Context

Patch on v0.44.0 (tag `v0.44.0`, Homebrew `traceary 0.44.0`, commit `15cb7352`). Live dogfood (#2183) showed the official v0.44 gates hold (search/report after a complete projection, live compact exclusive-wait through commit). Two defects remain that a patch should carry:

1. **Default `store compact` returns with search broken** until the operator knows to run `--projection-rebuild` (#2163, formerly v0.44-7 under #2126). Confirmed on the operator store after run `fe97b9bb…` (16.2 GiB → 12.9 GiB): projection `drifted/cleanup`, filterable no-match failed the 4096-row candidate budget. Search was restored only by a separate rebuild to generation `88101e57… complete`.
2. **Operator CLI open applies migrations under the 1 s hook busy_timeout** and fails `SQLITE_BUSY` even when `user_version` is already current (#2186).

Dogfood index: #2183. Parent #2126 stays closed (v0.44.0 already shipped). This parent stays open until the tagged v0.44.1 release workflow publishes.

## Sub-issues (1 issue = 1 branch = 1 PR)

- [ ] #2163 v0.44.1-1: store compact should rebuild the search projection before returning
- [ ] #2186 v0.44.1-2: operator CLI hits SQLITE_BUSY in 1s on an already-current schema

## Sequencing

- **Wave 0: #2163.** Medium Structure-Behavior: Design Note before code (compact exclusive lease vs projection rebuild batches; live writers must not race an in-flight generation that compact left unfinished). Compact stdout JSON stays frozen. Do not start a replacement generation after `complete`.
- **Wave 1: #2186.** Independent of Wave 0 (Initialize / migration apply vs compact usecase). Do not lengthen the hook `sqliteBusyTimeout=1000` ms.

## Out of scope → v0.45.0

Rollback-copy prune and abandoned `*.work-journal` (#2184), first-open 83 s (#2185; not reproduced after rebuild), v0.43 leftovers #2151–#2158. Operator deletion of an accepted rollback remains a documented decision, not this patch.

## Non-goals (frozen)

Search JSON shape, decode authority, `DeepLiteralSearchBudget`, FTS revival, `list` / `context` latency, hook busy_timeout.

Refs: #2126 (v0.44.0, closed), #2183 (v0.44.0 dogfood), #2163, #2186.

PR #2189, opened 2026-08-19
store compact completes the search projection before returning

### LINKED ISSUE #2163: v0.44.1-1: store compact should rebuild the search projection before returning
## Summary

Default `traceary store compact` rewrites the store and drops the retired search-index family on the work copy. It does **not** run a search-projection rebuild. After a successful compact on a large live store, unbounded `traceary search` hits the 4096-row candidate budget until the operator separately runs `traceary store compact --projection-rebuild` to `stop_reason=complete`.

This was observed on the host store after the v0.44 Wave 0 compact (53 GiB → 12 GiB, then projection rebuild to generation `complete`). Keyword search and the `SESSIONS` group returned only after that rebuild.

Parent: #2187
Leaves parent #2187 open.

## Why this belongs in the compact flow

Compact already takes an exclusive lease and is the only window where live hook/MCP writers are naturally blocked for the rewrite. `--projection-rebuild` today is a separate, batched operator command (`ResumeUntil` 100 batches / 10 minutes). On a live store, hook inserts between batches raise `search projection source changed; start a new generation`, which aborts in-flight cleanup and can park the generation as `failed`. The next `--projection-rebuild` then **starts a new generation**, leaving the previous generation's documents/keywords/fingerprints as `cleanup_scope=old` leftovers (millions of rows). Completing rebuild on a live 12 GiB store required a quiet writer window and hours of leftover cleanup.

Building the projection (or at least starting it and driving it to `complete`) before compact returns would:

1. Restore default search as part of the rewrite the operator already waited for.
2. Avoid the live-writer vs in-flight generation race that `--projection-rebuild` currently loses.
3. Avoid stacking failed/abandoned generations whose leftover fingerprints/keywords dominate post-compact cleanup.

## Acceptance

- After a successful default `store compact` on a store that had (or now needs) a bounded search projection, `traceary search <typical keyword>` exits 0 without the 4096-row `index_incomplete` budget error, and the `SESSIONS` group is populated when a generation is complete.
- Compact **stdout** stays unchanged (no new public command/flag required unless an existing budget flag is reused).
- Live writers that arrive after compact returns do not have to race an in-flight rebuild that compact itself left unfinished.
- The implementation PR must leave parent #2187 open.

## Out of scope

- Changing default `search` selectivity (#2127).
- Shrinking exclusive-wait vs rewrite (#2161).
- Deleting the rollback file.

duck8823:
## Summary

Default `traceary store compact` now drives the search projection to `complete` while the exclusive store lease is still held. Operators no longer need `--projection-rebuild` for search to work after compact, and live writers cannot interleave `source changed` mid-rebuild.

Same-process coordinated SQLite opens skip a second `LOCK_SH` while this process holds exclusive (flock is per open file description). Other processes remain blocked by the exclusive fd.

## Acceptance

- After a successful default compact, `search` of a filterable token does not hit the 4096-row `index_incomplete` budget.
- Compact stdout JSON keys are unchanged.
- Compact does not return with an in-flight generation.

Closes #2163

Leaves parent #2187 open.

### DIFF
--- a/application/usecase/search_projection_plan_test.go
+++ b/application/usecase/search_projection_plan_test.go
@@ -26,9 +26,11 @@ type wallBudgetStore struct {
 	measuredReads      int
 	state              string
 	resumeReady        bool
+	starts             int
 }
 
 func (s *wallBudgetStore) Start(context.Context, apptypes.SearchProjectionBudget, time.Time) (apptypes.SearchProjectionGeneration, error) {
+	s.starts++
 	return apptypes.SearchProjectionGeneration{}, nil
 }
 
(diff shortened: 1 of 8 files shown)


duck8823:
Antigravity gemini 3.7-flash medium: ✅

```json
{"reviewer":"Antigravity gemini 3.7-flash medium","severity_max":"NONE","summary":"問題なし","findings":[]}
```

Read-only review of `origin/main...HEAD` for #2163. No CRITICAL/HIGH findings. Exclusive-holder SQLite open without a second flock is the correct #2149 follow-through so compact can complete the search projection under the exclusive lease. Compact stdout JSON encoder is untouched.

PR #2190, opened 2026-08-19
operator CLI does not fail SQLITE_BUSY on a current schema

### LINKED ISSUE #2186: v0.44.1-2: operator CLI hits SQLITE_BUSY in 1s on an already-current schema
Parent: #2187

## Problem

brew 0.44.0 `traceary search …` against the live store, while another `traceary` process held the db (PID observed on `lsof`, then gone):

```
Error: … failed to run SQLite migrations: failed to execute migration SQL (version=73): database is locked (5) (SQLITE_BUSY)
real 1.05
```

Schema is already 73. `sqliteBusyTimeout` is 1000 ms by design for *hook* writers so they can keep a spool record. Operator CLI (`search` / `report` / `list`) still goes through Initialize + migration apply on every open, so a 1 s wait is enough to fail the whole command.

Retry a few seconds later succeeded (then paid the 83 s first-open cost, sibling issue).

## Change (one PR)

Operator/read CLI open should not fail the whole command because applying an already-current migration hit SQLITE_BUSY. Options (pick in the PR): skip DDL when `user_version` is current before taking a write lock; retry with a longer timeout on the CLI path than the hook path; or classify busy as a retryable open error and retry once.

Do not lengthen the hook busy timeout (that timeout is load-bearing for spool retention).

## Acceptance

- With a writer holding the live store for >1 s, `traceary search <filterable-no-match> --json --limit 5` either succeeds or retries until the writer releases, without a 1.05 s hard fail on migration 73.
- Hook write path busy_timeout stays 1000 ms.
- Search JSON shape is unchanged.

duck8823:
## Summary

Operator `search` / `report` / `list` no longer hard-fail in ~1s with `failed to execute migration SQL (version=73): SQLITE_BUSY` when the catalog is already current.

- `migrate` now reads `schema_migrations` first and skips DDL when every catalog version is applied.
- Operator open retries SQLITE_BUSY for up to 15s. Hook `sqliteBusyTimeout` remains 1000 ms.

## Acceptance

- With a writer holding the store for >1 s on a current schema, operator search does not fail the whole command on migration 73.
- Hook busy timeout stays 1000 ms.
- Search JSON shape is unchanged.

Closes #2186

Leaves parent #2187 open.

### DIFF
--- a/infrastructure/sqlite/database.go
+++ b/infrastructure/sqlite/database.go
@@ -524,7 +524,8 @@ func (d *Database) initializeAt(ctx context.Context, snapshot string, allowOffli
 // sqliteBusyTimeout is deliberately shorter than every packaged host hook
 // budget. Contention therefore returns control while the hook process still
 // has time to retain its write-ahead spool record instead of being killed at
-// the same instant SQLite's wait expires.
+// the same instant SQLite's wait expires. Operator CLI retries SQLITE_BUSY
+// separately; do not raise this constant for search/report/list (#2186).
 const sqliteBusyTimeout = 1000
 
 func sqliteDSN(dbPath string) string {
--- a/infrastructure/sqlite/migrate.go
+++ b/infrastructure/sqlite/migrate.go
@@ -24,22 +24,34 @@ func (d *Database) migrateAuthorized(ctx context.Context, db *sql.DB) error {
 }
 
 func (d *Database) migrateWithOptions(ctx context.Context, db *sql.DB, allowOffline bool) error {
-	if err := ensureSchemaMigrationsTable(ctx, db); err != nil {
-		return xerrors.Errorf("failed to create schema_migrations table: %w", err)
-	}
-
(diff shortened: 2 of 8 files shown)


duck8823:
Antigravity gemini 3.7-flash medium: ✅

```json
{"reviewer":"Antigravity gemini 3.7-flash medium","severity_max":"NONE","summary":"問題なし","findings":[]}
```

Read-only review of `origin/main...HEAD` for #2186. No CRITICAL/HIGH findings. Current-catalog migrate skips DDL; operator search/report/list retry SQLITE_BUSY for 15s; hook `sqliteBusyTimeout` remains 1000 ms.

duck8823:
Kimi k3: ✅

```json
{"reviewer":"Kimi K3","severity_max":"NONE","summary":"問題なし","findings":[]}
```

Independent read-only review of PR #2190 / #2186. No CRITICAL/HIGH findings.

Issue #2188, opened 2026-08-19
v0.45.0: compact leftover recovery, doctor/hook hygiene, and host plugin install leftovers


duck8823:
## Context

Minor after v0.44.1. Collects the v0.44.0 live-dogfood leftovers that are **not** the compact-then-search hole or the already-current-schema SQLITE_BUSY (those are v0.44.1). Also absorbs the remaining v0.43.0 dogfood backlog (#2148 / #2151–#2158).

Live store after v0.44.0 compact + operator `--projection-rebuild` (2026-08-19/20): 13.2 GiB, projection `88101e57… complete`, rollback sibling **16.2 GiB** retained by contract, abandoned `*.work-journal` 19 MiB still on disk.

Dogfood indexes stay open as records: #2148, #2183. This parent stays open until the tagged v0.45.0 release workflow publishes.

## Sub-issues (1 issue = 1 branch = 1 PR)

Compact leftover recovery:

- [ ] #2184 v0.45-1: leftover accepted rollback copy (and abandoned work-journal) is not prunable via doctor --fix

Read-side / first open (after #2186):

- [ ] #2185 v0.45-2: first CLI search after idle or lock is tens of seconds on a complete large store

Hook / doctor hygiene (from #2148):

- [ ] #2151 v0.45-3: hook spool organic drain cannot clear a requeue-scale backlog
- [ ] #2153 v0.45-4: doctor --fix should auto-resolve Claude SessionEnd cancellation markers
- [ ] #2154 v0.45-5: doctor spool counters and --db-path scope labels
- [ ] #2155 v0.45-6: doctor should detect long-running processes on stale binaries

Memory inbox:

- [ ] #2152 v0.45-7: retro-hide pre-v0.43 echo candidates

Host plugin install leftovers:

- [ ] #2156 v0.45-8: Gemini extension install hung on second folder-trust; uninstall-first left the host bare
- [ ] #2157 v0.45-9: docs(codex): headless refresh for local-path marketplace roots
- [ ] #2158 v0.45-10: docs(claude): document `claude plugin update plugin@marketplace`

## Sequencing

- **Wave 0: #2184.** Opt-in prune of an *accepted* rollback only. Never delete in-flight journals. Abandoned `*.work-journal` detection is in this same ticket (already in #2184 acceptance). Compact stdout JSON unchanged. Do not auto-delete rollback on `committed`.
[...]

PR #2200, opened 2026-08-20
fix: stage-then-swap Gemini extension install with timeout

### LINKED ISSUE #2156: v0.45-8: Gemini extension install hung on second folder-trust; uninstall-first left the host bare
Parent: #2188

## Problem

Headless run of `./scripts/install-gemini-extension.sh --ref v0.43.0` (2026-08-18): the script uninstalled the existing extension, then `gemini extensions install --consent <temp clone>` hung for ~30 minutes (stdin `/dev/null`, no output, node process sleeping) until killed — leaving `~/.gemini/extensions/` with **no traceary extension** at all. That is the #1171 coverage-loss class: a host silently records nothing until someone notices.

Immediately afterwards, `gemini extensions install --consent /Users/duck8823/Repositories/traceary/integrations/gemini-extension` (the repo checkout, same content as the tag) completed in seconds — the hang correlates with the temp-clone path, not with install as such. (The already-known separate issue: `gemini extensions update` hangs on a confirm prompt; the script exists precisely to avoid that.)

## Change (one PR)

1. Stage-then-swap ordering: verify the new install succeeds before uninstalling the old one; on failure, reinstall the previous version and exit non-zero naming the failed step.
2. Wall-clock timeout around each gemini CLI invocation with a clear error message.
3. Repro/investigate the temp-clone hang and record the result in the PR; if it is a gemini CLI behavior for git-sourced dirs at a detached ref, install from the checkout when the requested `--ref` matches its content.
4. `docs/integrations/gemini-extension.md` documents the recovery one-liner (direct install from the checkout).

## Acceptance

- A simulated install failure (stub gemini binary) leaves the previous extension installed and exits non-zero.
- The headless happy path completes within the timeout on a real gemini CLI.
- Script-only + docs change; no CLI/plugin interface change.

duck8823:
## Summary

`scripts/install-gemini-extension.sh` no longer uninstalls first.

- Copies the existing extension aside, then installs the new package
- Each `gemini` CLI call is bounded by `TRACEARY_GEMINI_TIMEOUT` (default 60s)
- Failure or timeout restores the previous directory and exits non-zero
- When `--ref` matches this checkout's `VERSION`, install from `integrations/gemini-extension` (temp clones hung on Gemini's second folder-trust prompt)
- Docs include the checkout recovery one-liner

## Test plan

- [x] `go test ./scripts/ -count=1` (stub gemini fail + timeout leave previous marker)
- [x] `go test ./... -count=1`
- [x] `go build ./...`

Closes #2156

Leaves parent #2188 open.

### DIFF
--- a/.ai/spec/2156.md
+++ b/.ai/spec/2156.md
@@ -7,7 +7,7 @@ Parent: #2188. Closes only #2156.
 Risk: **Medium** — install script can leave the host with no Traceary extension.
 
 ### Requirement summary
-- Keep the previous Gemini extension until the new package is verified.
+- Do not leave the host bare when install hangs or fails (the v0.45 dogfood: uninstall-first then hang).
 - Bound each `gemini` CLI call with a wall-clock timeout.
 - On failure or timeout, restore the previous directory and exit non-zero.
 - When `--ref` matches this checkout's VERSION, install from `integrations/gemini-extension` (temp-clone hung on folder-trust).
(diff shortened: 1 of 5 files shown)


duck8823:
Kimi k3 high: first pass **HIGH** — gemini-cli `extensions install` is `isUpdate=false` and rejects an already-installed `traceary` name, so overlay stage-then-swap never installed an update.

Addressed in `d37f1656`: filesystem backup, then timed uninstall, then timed install; restore on fail/timeout. Follow-up: ✅ `summary`: 問題なし. `findings`: [].

Engine identity: Kimi K3 (kimi-code/k3).

duck8823:
Antigravity gemini 3.7-flash-medium: first pass ✅ `summary`: 問題なし (did not catch the overlay HIGH; Kimi did). Timeout/restore/docs coverage of the ticket was otherwise consistent.

Engine identity: Antigravity (Gemini 3.7 Flash).

